PR #22, opened 2015-11-24
Feature/dynamic concat

bradphelan:


DynamicData/List/ObservableListEx.cs
```diff
@@ -1060,6 +1061,90 @@ public static IObservable<IChangeSet<T>> Or<T>([NotNull] this IObservable<IChang
...
+                .Switch()
+                .StartWith(new List<T>())
+                .Buffer(2, 1)
+                .Where(b=>b.Count==2)
+                .Select
+                (pair =>
+                 new ChangeSet<T>
+                     (new[]
+                      {
+                          new Change<T>(ListChangeReason.Clear, pair[0]), new Change<T>(ListChangeReason.AddRange, pair[1])
```
RolandPheasant: FYI: Clear is  more efficient but only noticeable for large collections

DynamicData/List/ObservableListEx.cs
```diff
@@ -1060,6 +1061,90 @@ public static IObservable<IChangeSet<T>> Or<T>([NotNull] this IObservable<IChang
...
+        } 
+
+        /// <summary>
+        /// Apply a concatenation of all the lists without preserving order. Child
+        /// lists can be added and removed
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="sources"></param>
+        /// <returns></returns>
+        public static IObservableList<T> DynamicConcatNonOrderPreserving<T>
```
RolandPheasant: I presume you mean concat because of the way it is implemented.  I would suggest the final version will only need one method as order should always be preserved

bradphelan: The order preserving one is not a good implementation. The non order preserving one I'm not sure how to make it order preserving.

oysteinkrog:
:+1: This would be useful

RolandPheasant:
I will produce an optimised version soon. I have been very busy with Tail Blazer

RolandPheasant:
Closed as this is now implemented. See https://github.com/RolandPheasant/DynamicData/issues/20

Issue #31, opened 2016-03-13
Group with IEqualityComparer?


flagbug:
Hey there, I'm looking for an overload of the `Group` function, that takes an `IEqualityComparer`, just like the `GroupBy` function in LINQ and Rx.

Is this something this library should/would/could implement, or is there any other method to achieve this functionality?

RolandPheasant:
Personally with dynamic data, I have never needed an overload for an equality comparer when grouping which is why I have never implement it,  

However I agree that it would be very useful. The first use case I can think of is grouping on a string and passing in a case insensitive comparer.   I will enable this soon. 

To achieve similar results but for different reasons, I have previously created custom structs which have their own equality implemented. For example

```
 myCache.Connect()
      .Group(x=> (CaseInsensitiveString)x.myString)
      ...     .
```

An example of the case insensitive string is here [CaseInsensitiveString.cs](https://github.com/RolandPheasant/TailBlazer/blob/master/Source/TailBlazer.Domain/CaseInsensitiveString.cs).  The reason I tend to do it that way is I define my equality once only and that equality is by default used everywhere.

flagbug:
>  The first use case I can think of is grouping on a string and passing in a case insensitive comparer.

This is exactly my usecase :smile: 

> To achieve similar results but for different reasons, I have previously created custom structs which have their own equality implemented.

Cool, I'll try that, thanks!

RolandPheasant:
Let me know how you get on. Also any further questions probably the best place to catch me is [Dynamic data on gitter](https://gitter.im/RolandPheasant/DynamicData)

projecjavascript:
Hello Roland how are you?

RolandPheasant:
Ancient so closing

PR #32, opened 2016-03-13
Create a failing test based on #31 and some experiments I made with Dynamic Data

flagbug:
So I've experimented with Dynamic Data, and hit a roadblock with the suggestions @RolandPheasant made in #31 

For some reason, the sequence in the test case I created, fails with a SortException "Current position cannot be found. The item is not in the collection". 

Interestingly, this test passes, if we apply "SortOptimisations.ComparesImmutableValuesOnly" to the Sort operator.

Sorry about the code that I just quickly pressed into the SortFixture class, I didn't know where to put it, but I'm open for suggestions!

### DIFF
--- a/DynamicData.Tests/CacheFixtures/SortFixture.cs
+++ b/DynamicData.Tests/CacheFixtures/SortFixture.cs
@@ -2,7 +2,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


RolandPheasant:
BTW no problems with putting the code into the SortFixture class.  I am just happy that you took the trouble to provide a failing unit test. Few people ever do that.

PR #35, opened 2016-04-20
Fix bug when subscribing to MutableFilter multiple times

MendelMonteiro:
If you subscribe twice to a stream which has a filter the change sets sent to the second subscriber will be incorrect.  This is because the MutableFilter class holds global state which will be mutated twice per input change set received.

### DIFF
--- a/DynamicData.Tests/ListFixtures/FilterFixture.cs
+++ b/DynamicData.Tests/ListFixtures/FilterFixture.cs
@@ -1,4 +1,7 @@
-﻿using System.Linq;
+﻿using System;
+using System.Diagnostics;
+using System.Linq;
+using DynamicData.Controllers;
 using DynamicData.Tests.Domain;
 using NUnit.Framework;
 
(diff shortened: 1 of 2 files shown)


RolandPheasant:
Thanks. I suspect that this same problem will apply to multiple operators. Are you  tempted to take a look?

ablanchet:
I'm afraid you might be right.
Every operator that have a global shared state outside of its process method that is actually mutated when a change is processed for a subscriber might have the same problem. But still, it's a bit more subtle than this, the state has to be actually mutated from one call to another.

For example I think the [Sort](https://github.com/RolandPheasant/DynamicData/blob/master/DynamicData/List/Internal/Sort.cs) operator is a good candidate for this issue with its `_innerList`

We'll have to add some more unit tests around this "multiple subscription" use case.

RolandPheasant:
I genuinely cannot believe I missed such an obvious detail. I suppose that comes from the fact that I use .Publish() for multiple subscriptions. But in the spirit of rx, the issue needs to be sorted

ablanchet:
Yes, and I guess you're using `.Publish()` via your magical `.RefCount()` implementation that supports "replays" when new subscribers shows up :) We had an issue with this for several days in our usage of DynamicData and we finally realised today it was because we were using the native Reactivex's RefCount (that does not replay). Tricky bug !

RolandPheasant:
Ah - good point. I suspect this is why we always hear the maxim not to keep state in rx. But I say 'bull' to that maxim as complex scenarios call for complex solutions.  Also I applaud you to recognise where rx and dynamic data cross over. No one has ever spotted that before so great observation.

RolandPheasant:
I have taken a look and can confirm there are several shared state issues.

The fix is easy so expect and update soon.

PR #37, opened 2016-04-20
Make sure that RemoveAt and RemoveRange have the same behavior on ChangeAwareList

ablanchet:
Hi Roland,

I think that ChangeAwareList's `RemoveAt` and `RemoveRange` methods should have the same behavior when trying to remove items outside of the `_innerList` range. 
Currently when you call `RemoveAt` with an index that's outside of the list range you get a nice `ArgumentOutOfRangeException`. But when you call `RemoveRange` with a range that begins or finishes outside of the `_innerList` range you get nothing (since the implementation is based on `Skip` / `Take` Linq methods).

This PR has two commits : first with only the failing unit tests, second commit with the fix in `ChangeAwareList.cs`

### DIFF
--- a/DynamicData.Tests/ListFixtures/ChangeAwareListFixture.cs
+++ b/DynamicData.Tests/ListFixtures/ChangeAwareListFixture.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using DynamicData.Internal;
 using DynamicData.Kernel;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


RolandPheasant:
Good call

ablanchet:
Thank you, we were reviewing with @MendelMonteiro and it's actually him who pointed it :)

RolandPheasant:
It's a good call. The ChangeAwareList is frightfully difficult so I really appreciate any input.

I can honestly say getting it right has been the most difficult bit of getting the observable list right.

RolandPheasant:
@ablanchet  and if you find any other issues please contribute!

RolandPheasant:
@ablanchet @MendelMonteiro having done the change aware list I see why Microsoft did a simple ObservableCollection as predicting and optimising collection changes is very difficult

ablanchet:
Yep ... The ChangeAwareList could have been way simpler if you didn't have to handle "Range" operations. It's a good performance improvement but it adds a lot of complexity to the whole library since all operators now have to handle four types of operations (instead of two before).

RolandPheasant:
Yep. But until I did the range stuff I could not get the list stuff to perform well, so it was a necessary complexity. Such complexities separates the wheat from the chaff.

PR #62, opened 2016-12-13
Make RefCount operators thread-safe

Hinidu:
Hi, @RolandPheasant!
I was very curious about DynamicData internals and started to read some code. I found that `RefCount` operator is not thread-safe and I decided to give it a spin :smile:
I've also added a probablistic test to check thread-safety (you can run it and see that the previous implementation throws `NullReferenceException` almost everytime).
It's unclear to me how to test it deterministically.
Though I was trying hard to prove it as strictly as I could.
I'm still unsure if I should write this informal "proof" alongside the actual code or in the method doc-string.
Regardless of the actual placement I will try to do it tomorrow. Until when I will be happy to hear any questions or another feedback in this PR.

### DIFF
--- a/DynamicData.Tests/CacheFixtures/RefCountFixture.cs
+++ b/DynamicData.Tests/CacheFixtures/RefCountFixture.cs
@@ -2,6 +2,8 @@
 using DynamicData.Tests.Domain;
 using NUnit.Framework;
 using System;
+using System.Threading.Tasks;
+using System.Linq;
 
 namespace DynamicData.Tests.CacheFixtures
 {
@@ -72,6 +74,27 @@ namespace DynamicData.Tests.CacheFixtures
 
             Assert.AreEqual(2, created);
             Assert.AreEqual(2, disposals);
+        }
+
+        // This test is probabilistic, it could be cool to be able to prove RefCount's thread-safety
+        // more accurately but I don't think that there is an easy way to do this.
+        // At least this test can catch some bugs in the old implementation.
+        [Test]
+        public async Task IsHopefullyThreadSafe()
+        {
+            var refCount = _source.Connect().RefCount();
+
+            await Task.WhenAll(Enumerable.Range(0, 100).Select(_ =>
+            {
+                return Task.Run(() =>
+                {
+                    for (int i = 0; i < 1000; ++i)
+                    {
+                        var subscription = refCount.Subscribe();
+                        subscription.Dispose();
+                    }
+                });
+            }));
         }
     }
 }
--- a/DynamicData.Tests/CacheFixtures/RefCountFixture.cs
+++ b/DynamicData.Tests/CacheFixtures/RefCountFixture.cs
@@ -85,16 +85,14 @@ namespace DynamicData.Tests.CacheFixtures
             var refCount = _source.Connect().RefCount();
 
             await Task.WhenAll(Enumerable.Range(0, 100).Select(_ =>
-            {
-                return Task.Run(() =>
+                Task.Run(() =>
                 {
                     for (int i = 0; i < 1000; ++i)
                     {
                         var subscription = refCount.Subscribe();
                         subscription.Dispose();
                     }
-                });
-            }));
+                })));
         }
     }
 }
--- a/DynamicData.Tests/ListFixtures/RefCountFixture.cs
+++ b/DynamicData.Tests/ListFixtures/RefCountFixture.cs
@@ -2,6 +2,8 @@ using System;
 using System.Reactive.Linq;
 using DynamicData.Tests.Domain;
 using NUnit.Framework;
+using System.Threading.Tasks;
+using System.Linq;
 
 namespace DynamicData.Tests.ListFixtures
 {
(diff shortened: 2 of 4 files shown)


RolandPheasant:
I am pleased you have found and plugged this hole. As for testing it deterministic I am not sure whether that is possible.  I am happy to commit this PR but will hold off for now to see whether you can find  the elusive scientific way of testing for thread safety.

Hinidu:
Yesterday I was curious if this `Interlocked` soup is any faster than more straightforward approach with locks. I've used [BenchmarkDotNet](http://benchmarkdotnet.org/) to compare two implementations and the version with `lock`s were actually faster by about 7-8% than my current implementation. But what is more important is that when I've changed the number of workers to 10 and the number of iteration in subscribe-dispose cycle I've got `NullReferenceException` with my current implementation on each benchmark run. While the version with `lock`s didn't throws in any run. I was trying to understand so long why the current implementation can cause this exception but I couldn't until I fall asleep :disappointed: But it seems that my mind was working hard while I was sleeping because right when I woke up I had this scenario in my head:

This sceanario involves 3 concurrent threads, let's call them A, B and C.
A subscribes and start disposing its subscription. Though A only managed to decrement `_refCount` for now.
B subscribes and increment `_refCount` so it is now equal to 1 again. B didn't create the new `_cache` though.
C subscribes and sees that `_refCount` is 2 so it should wait while `_cache` is `null`. But it is not `null` yet! Because A didn't dispose it yet and `_cache` is equal to the cache created by A which is going to be disposed. So C skips `SpinWait.SpinUntil` just  fine.
From this point there are two erroneous scenarios:
1) C subscribes to the `_cache` which will be disposed soon.
2) A finishes disposing this `_cache` and setting `_cache` to `null` (and I think that's how I got `NullReferenceException` in the first place).

I will commit the new version with `lock`s soon. I've made two conclusions:
1) Concurrent programming is very hard with mutable state so it's better to write simpler code and even sacrifice performance a bit (though the simpler code is actually faster in that particular case).
2) `lock`s and other synchronization primitives which allow to serialize access to the mutable state can act like transactions in databases and greatly simplify the reasoning because nobody will gonna see partial updates from another thread (it was the cause of the failure: thread A turned `_refCount` to 0, but didn't set `_cache` to `null` and thread C saw the system in that partially updated state). Though they are not silver bullets because with them we should be very careful to avoid deadlocks.

Hinidu:
I've commited the new version. But it still needs more thoughts or verification if we want to assure that now it's really ok. I will try to do something in that direction today.

MendelMonteiro:
@Hinidu, the lock version looks pretty sound to me, all your mutable state is modified inside the lock blocks.

I don't think that it needs to be lock-free as I don't think that there will be much contention on calling RefCount() - it's normally not driven by a data event.  If someone is creating DD pipelines often enough where contention is a problem I would say that they should probably change their design.

Hinidu:
> @Hinidu, the lock version looks pretty sound to me, all your mutable state is modified inside the lock blocks.

@MendelMonteiro Thank you for looking into it! Modifying mutable state under locks is good but it doesn't guarantee the algorithm correctness - for example we still can have deadlocks and/or race conditions during the access to the mutable state outside of the lock.

> I don't think that it needs to be lock-free as I don't think that there will be much contention on calling RefCount() - it's normally not driven by a data event. If someone is creating DD pipelines often enough where contention is a problem I would say that they should probably change their design.

I also thought whether we need this level of concurrency in `RefCount` or not. Your opinion makes sense. And by the way I don't think that we can do much better than what we have now:
```c#
lock (_locker)
    if (++_refCount == 1)
        _cache = _source.AsObservableCache();
```
The only possible long operation here is `_source.AsObservableCache()` and waiting for its completion looks unavoidable because if many threads are in this section and `_cache` was not created yet all threads should wait while one of them will create it. Other operations under this lock - increment and comparison can make an issue only under very heavy load I think.
```c#
IDisposable cacheToDispose = null;
lock (_locker)
    if (--_refCount == 0)
    {
        cacheToDispose = _cache;
        _cache = null;
    }
if (cacheToDispose != null)
    cacheToDispose.Dispose();
```
Here we moved the only possible heavy operation `Dispose` out of the `lock` and we do just a few simple operations inside it.

But all the reasoning above is only about performance. I still want to become more certain about the correctness of this algorithm. The thing that I found and that increase my confidence is that Reactive Extensions's `RefCount` implementation is very [similar](https://github.com/Reactive-Extensions/Rx.NET/blob/371c83c621562a2259580a03f0cb5bf8680ea720/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/RefCount.cs#L45-L69).

RolandPheasant:
This is a great discussion about the best way to resolve thread safety in RefCount. I am inclined to agree with @Hinidu that doing the locking in a similar way to Rx's RefCount is wise

RolandPheasant:
I will merge the PR as soon as you tell me you are happy

Hinidu:
```c#
                lock (_locker)
                    if (++_refCount == 1)
                        _cache = _source.AsObservableCache();

                var subscriber = _cache.Connect().SubscribeSafe(observer);

                return Disposable.Create(() =>
                {
                    subscriber.Dispose();
                    IDisposable cacheToDispose = null;
                    lock (_locker)
                        if (--_refCount == 0)
                        {
                            cacheToDispose = _cache;
                            _cache = null;
                        }
                    if (cacheToDispose != null)
                        cacheToDispose.Dispose();
                });
```
I think I'm sure enough that this algorithm is ok. This is my last thoughts on its correctness:
1) Thanks to locks we have an invariant that `_cache` is `null` iff `_refCount` is `0`.
2) When thread subscribes to `_cache` it is not `null` because `_refCount` is clearly greater than `0` in that case.
3) It seems that there are no places where something bad could happen.
4) Also something good will eventually happen because no threads can wait for each other. In all cases there is at most one process inside one of the locks - so all the other threads can be waiting for him but noone else, and the thread inside the lock is never waiting for someone.

> I will merge the PR as soon as you tell me you are happy

@RolandPheasant I think I'm quite happy now :smile:

PR #63, opened 2017-01-18
Chained filters issue

ltrzesniewski:
This PR exposes an issue with chained filters, please see the failing `ChainFilters` test.

When there are several `.Filter` calls, if reevaluatation of the first filter leads to items becoming available or not available to the second filter, the second filter won't be notified of this and won't reevaluate.

This used to work in DD 4 when using multiple FilterControllers.

### DIFF
--- a/DynamicData.Tests/DynamicData.Tests.csproj
+++ b/DynamicData.Tests/DynamicData.Tests.csproj
@@ -143,6 +143,7 @@
     <Compile Include="CacheFixtures\ExpireAfterFixture.cs" />
     <Compile Include="CacheFixtures\BatchIfWithTimeOutFixture.cs" />
     <Compile Include="CacheFixtures\DeferUntilLoadedFixture.cs" />
+    <Compile Include="ListFixtures\FilterWithObservable.cs" />
     <Compile Include="ListFixtures\FromAsyncFixture.cs" />
     <Compile Include="Kernal\EnumerableEx.cs" />
     <Compile Include="CacheFixtures\EnumerableObservableToObservableChangeSetFixture.cs" />
--- a/DynamicData.Tests/ListFixtures/FilterWithObservable.cs
+++ b/DynamicData.Tests/ListFixtures/FilterWithObservable.cs
@@ -1,0 +1,255 @@
+using System;
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Reactive.Subjects;
+using DynamicData.Controllers;
+using DynamicData.Internal;
+using DynamicData.Aggregation;
+using DynamicData.Tests.Domain;
+using NUnit.Framework;
+using System.Collections.Generic;
+
+namespace DynamicData.Tests.ListFixtures
+{
+    [TestFixture]
+    public class FilterWithObservable
+    {
+        private ISourceList<Person> _source;
+        private ChangeSetAggregator<Person> _results;
+        private BehaviorSubject<Func<Person, bool>> _filter;
+
+        [SetUp]
+        public void Initialise()
+        {
+            _source = new SourceList<Person>();
+            _filter = new BehaviorSubject<Func<Person, bool>>(p => p.Age > 20);
+            _results = _source.Connect().Filter(_filter).AsAggregator();
+        }
+
+        [TearDown]
+        public void Cleanup()
+        {
+            _source.Dispose();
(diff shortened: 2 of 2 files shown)


Hinidu:
It looks like the bug is here: https://github.com/RolandPheasant/DynamicData/blob/c7728c212ae69a4680cf73d21ccaf1f6aa233b79/DynamicData/List/ListEx.cs#L93-L94
```csharp
                    case ListChangeReason.Remove:
                        {
                            var change = item.Item;
                            var wasMatch = predicate(change.Current);
                            if (wasMatch) source.Remove(change.Current);
                            break;
                        }
```
I think we should use `change.Previous.Value` instead of `change.Current`. But I'm not sure and can't check it now because I have only VS 2012 at work and found it just by reading code in GitHub.
@ltrzesniewski maybe you can check my assumption?

ltrzesniewski:
@Hinidu that doesn't fix the issue, it actually makes 6 other tests fail :)

I *think* the issue was introduced in 5c1c61936e82360bfcbf564b2b6362cfc307a0d5, as this used to work fine with `FilterPolicy.ClearAndReplace`

Hinidu:
@ltrzesniewski oops :-)
If you are right the bug was here even before this commit if you used another `FilterPolicy`.

Hinidu:
I doubt that the problem is really in removed `FilterPolicy`. It looks like it affects only reevaluation in the case of changed filter (and that works well according to your other tests). I think the bug is somewhere in processing incoming `Remove` item event in the second `Filter`.

Hinidu:
I was right about the reason of this bug - the method I was pointing about `ListEx.Filter`. To fix your test you can just remove item unconditionally:
```csharp
                    case ListChangeReason.Remove:
                        {
                            var change = item.Item;
                            source.Remove(change.Current);
                            break;
                        }
```
It doesn't break any existing tests. But this is not enough the whole method assumes that the result of `predicate(item)` was not changed, but it can be changed because `item` is mutable (if I understand correctly this method is not called with changed `predicate` so only the changes in `item` are a problem). `Replace` and `RemoveRange` handlers will also work inappropriate (you just don't have testcases to catch it).

I'm not sure whether `ListEx.Filter` should be updated to account mutations of items or should we create another one with `Mutable` suffix and call each one in appropriate places.

MendelMonteiro:
@Hinidu I agree with the logic behind changing the code that you suggest.  If an upstream source has removed an item from the stream it should implicitly be removed from the filtered results.  We should also do the same thing for `RemoveRange`.  However for `Replace`, I think it's handled correctly already.

If a filter is not reevaluated explicitly it should only re-evaluate when its upstream source changes (add/remove/replace).  To implement the mutable filter you evoke the pattern that exists is to use `WhenPropertyChanged`.

Hinidu:
> If an upstream source has removed an item from the stream it should implicitly be removed from the filtered results.

@MendelMonteiro the reason why I'm bothering about additional mutable version because current implementation can be more performant for a large list because evaluation of predicate usually is much faster than removal from the list so if predicate returns `false` we can avoid needless list scan.

> However for Replace, I think it's handled correctly already.

I think that if previous value was modified then we will have the same problems with `Replace`. I will try to write the tests that catches it and `RemoveRange`.

> If a filter is not reevaluated explicitly it should only re-evaluate when its upstream source changes (add/remove/replace). To implement the mutable filter you evoke the pattern that exists is to use WhenPropertyChanged.

This is still very hard for me to decide what we should do with all these possible mutations... Immutable structures are so much simpler to reason about... Maybe it's just to late to think about all this stuff :smile:

RolandPheasant:
At last I have been able to look at the issue.  Mutable changes has been considered, but incorrectly.  

The fix should be reasonably simple

In the class ```MutableFilter<T>``` a list of previous matches is already maintained using the following

```
 var allWithMatch = new List<ItemWithMatch>();
```

This should be used when assessing whether a item was previously a match instead of re-evaluating the filter again

Hinidu:
> In the class MutableFilter<T> a list of previous matches is already maintained using the following
 var allWithMatch = new List<ItemWithMatch>();
This should be used when assessing whether a item was previously a match instead of re-evaluating the filter again

👍 I also thought about that in the morning. I think it should work.
Also there is an `all` list that is not used anymore and can be deleted.

PR #73, opened 2017-04-01
Implement IExtendedList

sfmskywalker:
Having `ObservableCollectionExtended<T>` implement `IExtendedList<T>` ensures the `ListEx.Clone` method invokes the `Move` method.

An alternative solution I considered was introducing a new interface called something like `IMove`, and derive both `IExtendedList<T>` as well as `IObservableCollection<T>` from that, so that `ObservableCollectionExtended` doesn't have to also implement the `AddRange`, `RemoveRange` and `InsertRange` methods. But then again, maybe it's not a bad thing it supports thee operations as well.

Let me know if you have any other suggestions, I'm happy to make changes.


RolandPheasant:
Looks good to me.

PR #89, opened 2017-08-14
Add vs Update

cubed-it:


RolandPheasant:
Thanks for this. It will make it much easier for me to diagnose

PR #95, opened 2017-10-06
Fixed typo in ChangeAwareList's class description

snalesso:


RolandPheasant:
Thanks for that. There's probably quite a few  typos in the lib but one less now

PR #99, opened 2017-10-26
Master

### LINKED ISSUE #98: BatchIf race condition
If a batched set of changes is consumed on a different thread (UI thread in this case) from the producer (Threadpool thread) there is a race condition between the buffered changes being cleared and the consumer consuming the buffered change set, resulting in buffered changes being lost.

from BatchIf.cs...
`                        var resume = bufferSelector.Where(shouldPause => !shouldPause)`
`                            .Subscribe(_ =>`
`                            {`
`                                paused = false;`
`                                //publish changes and clear buffer`
`                                if (buffer.Count == 0) return;`
`                                observer.OnNext(new ChangeSet<TObject, TKey>(buffer));`
`                                buffer.Clear();` **<-- Race between this and the OnNext being consumed**

 `                               //kill off timeout if required`
 `                               timeoutSubscriber.Disposable = Disposable.Empty;`
 `                           });`

I am working on a fix for this

robsonj:
Fixes #98

### DIFF
--- a/DynamicData.Tests/Cache/BatchIfFixture.cs
+++ b/DynamicData.Tests/Cache/BatchIfFixture.cs
@@ -1,5 +1,8 @@
 using System;
+using System.Collections.ObjectModel;
+using System.Reactive.Linq;
 using System.Reactive.Subjects;
+using DynamicData.Binding;
 using DynamicData.Tests.Domain;
 using FluentAssertions;
 using Microsoft.Reactive.Testing;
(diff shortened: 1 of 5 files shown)


robsonj:
BatchIf now supports buffer release on an interval timer as well as the
original features of releasing the buffer on a timer event.

Redesign also incorporates use of Reactive Buffer operator which I don't think existed at the time of writing the original

robsonj:
Failing unit test

PR #102, opened 2017-11-23
potential sorting optimizations

mmalek06:
When my team started putting 5000+ elements into the source cache and sort it, the performance started to drop in a exponentially way. This commit fixes that problem. All your tests are passing, we also checked it "visually" - our dataset is sorted ok after using this code. On the other hand we can see that the changesets being produced by this solution and the one being produced by your code differ in the number of elements, also they contain different moves. Therefore we are not 100% sure in regard to correctness of our solution - test are passing using both approaches.

### DIFF
--- a/DynamicData/Cache/Internal/IndexCalculator.cs
+++ b/DynamicData/Cache/Internal/IndexCalculator.cs
@@ -67,26 +67,25 @@ namespace DynamicData.Cache.Internal
             }
             else
             {
-                int index = -1;
+                var index = -1;
                 var sorted = _list.OrderBy(t => t, _comparer).ToList();
+                var unsortedCache = _list.Select((kvp, idx) => new { kvp, idx }).ToDictionary(o => o.kvp, o => o.idx);
+
                 foreach (var item in sorted)
                 {
-                    KeyValuePair<TKey, TObject> current = item;
                     index++;
 
-                    //Cannot use binary search as Resort is implicit of a mutable change
-                    KeyValuePair<TKey, TObject> existing = _list[index];
-                    var areequal = EqualityComparer<TKey>.Default.Equals(current.Key, existing.Key);
-                    if (areequal)
+                    var oldPos = unsortedCache[item];
+
+                    if (oldPos == index)
                     {
                         continue;
                     }
-                    var old = _list.IndexOf(current);
-                    _list.RemoveAt(old);
-                    _list.Insert(index, current);
 
-                    result.Add(new Change<TObject, TKey>(current.Key, current.Value, index, old));
+                    result.Add(new Change<TObject, TKey>(item.Key, item.Value, index, oldPos));
                 }
+
+                _list = sorted;
             }
 
             return new ChangeSet<TObject, TKey>(result);


RolandPheasant:
This looks very interesting. I will grab the branch and check it out as soon as I get a chance.  I always welcome improvements to dd so thanks for your effort.

Also I do not know whether you knew but one optimisation for sorting is to use the sort reset threshold i.e. ```.Sort(myComparer, resetThreshold=50)```.  Give that a try and let me know how you get on.

DynamicData/Cache/Internal/IndexCalculator.cs
```diff
@@ -67,26 +67,25 @@ public void Reset(ChangeAwareCache<TObject, TKey> cache)
             }
             else
             {
```
RolandPheasant: This looks like a radical improvement to me.

mmalek06:
I will experiment with this resetThreshold stuff then. However, I can't see any documentation or comments on how to use it, so the question is - how do you use it? :D

DynamicData/Cache/Internal/IndexCalculator.cs
mmalek06: It is, but I'm still not sure if it won't actually break your stuff. I figured you've done the sorting the way you did, because there were important reasons, therefore my change may do something nasty, something that wouldn't be immediately obvious...

RolandPheasant: I have been thinking about this. I remember now that the huge complexity with dd is there is no way to predict what a changeset may contain. For example there may be multiple updates to an object with the same key within a single changeset. This is particularly true when a batch operation is applied.  The proposed algorithm will work wonders but will fail in this case.

However there should be a simple solution to account for that case i.e check for duplicates within the same changeset. If there are no duplicates, apply the old algorithm, otherwise apply the new one.

RolandPheasant:
The reset is on the api but alas I spend so much time maintaining and developing dd that I have little headroom for documentation. 

I accept there needs to be more than exists and I have taken efforts to provide as many examples as possible as code talks loudly. Today it was suggested on slack that every operator needs documentation which holds true.  Hopefully I can reach out to the community and find some people to help. A dedicated web site would be awesome but such things are not my strong point.

DynamicData/Cache/Internal/IndexCalculator.cs
mmalek06: So in other words, should I check for uniqueness of objects inside the _list var?

RolandPheasant: On second thoughts, I will take a fresh look at the code when I am on desktop later. I have been looking at this on my phone which is not ideal.

mmalek06: I've just taken a fresh look at your code and I don't get one thing - it seems that except element[0], every other element "old index" is calculated the wrong way. Take a list of A, B and C letters that you want to sort in the inverse order. Boiling it down to a list of old positions, the result will contain information that C moved from 2, which is correct, and that B also moved from position 2 which doesn't seem to be correct. The question is then: should it be like that?

RolandPheasant: I can honestly say I cannot remember why I did it that way. All I remember is sorting was the most complicated operator to get right in dd. Looking at the unit tests, they are rather sparse and require padding out. I suggest I do that first before making any more changes

robsonj: "For example there may be multiple updates to an object with the same key within a single changeset. This is particularly true when a batch operation is applied."

This is a problem I wanna address also when we can move to DD v6. I wanna conflate those changes if possible to keep only the most recent change to an item. Will be happy to write an operator for this when we get to v6

onliner10:
Any updates on this?

RolandPheasant:
Not yet. I am just catching my breath after several months of intense development on this lib.

RolandPheasant:
Btw I mentioned the reset threshold on the sort operator in a previous comment. Some of the overloads can work wonders for sort performance. 

There is an example in dynamic trader. See https://github.com/RolandPheasant/Dynamic.Trader/blob/master/Trader.Client/Views/LiveTradesViewer.cs

alexgula:
We have tried reset threshold in our sort an indeed it changed performance dramatically, thanks, it's not an issue now at all!

On the other hand, it's interesting if it's possible to optimize sorting, not from practical perspective, but as a nice algorithmic task.

RolandPheasant:
I think I will change the sort operator to have a default threshold rather than it being opt in. The only question is what should that threshold be?

onliner10:
Actually, what is the use case for not using reset threshold when sorting ?

RolandPheasant:
There is no use case. It is only that way for historic reasons.

In my latest branch I have changed the default value to be 100.

onliner10:
Actually, I've been thinking that it would be very beneficial (in terms of performance) to introduce this change.. Since the sort reason is actually "Reorder", we shouldn't need to reflect any mutable changes while calculating moves (reorder is stateless) 
This change would improve DD performance greatly

PR #115, opened 2018-05-04
new operator to convert sorting operations resulting in moves into re…

jasonwurzel:
Some UI list controls are not able to cope with Moves and only understand Adds+ Removes. (see UWP ListView, for example).
so in introduced a new extension method on 
`IObservable<ISortedChangeSet<TObject, TKey>> `
called
`TreatMovesAsRemoveAdd`

in the wiki on the sorting page maybe we need to give people a hint as to that this operator exists and when to use it...

### DIFF
--- a/DynamicData/Cache/ObservableCacheEx.cs
+++ b/DynamicData/Cache/ObservableCacheEx.cs
@@ -1663,6 +1663,45 @@ namespace DynamicData
             return new Sort<TObject, TKey>(source, comparer, sortOptimisations, null, resorter, resetThreshold).Run();
         }
 
+	    /// <summary>
+	    /// Converts moves changes to remove + add
+	    /// </summary>
+	    /// <typeparam name="TObject">The type of the object.</typeparam>
+	    /// <typeparam name="TKey">The type of the key.</typeparam>
+	    /// <param name="source">The source.</param>
+	    /// <returns>the same SortedChangeSets, except all moves are replaced with remove + add.</returns>
+	    public static IObservable<ISortedChangeSet<TObject, TKey>> TreatMovesAsRemoveAdd<TObject, TKey>(
+		    this IObservable<ISortedChangeSet<TObject, TKey>> source)
+	    {
+		    if (source == null) throw new ArgumentNullException(nameof(source));
+
+		    return source.Select(changes =>
+		    {
+			    var changesWithoutMoves = changes.SelectMany(change =>
+			    {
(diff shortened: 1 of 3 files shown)


DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -1663,6 +1663,45 @@ public static IObservable<IReadOnlyCollection<T>> StartWithEmpty<T>(this IObserv
...
+	    /// <typeparam name="TObject">The type of the object.</typeparam>
+	    /// <typeparam name="TKey">The type of the key.</typeparam>
+	    /// <param name="source">The source.</param>
+	    /// <returns>the same SortedChangeSets, except all moves are replaced with remove + add.</returns>
+	    public static IObservable<ISortedChangeSet<TObject, TKey>> TreatMovesAsRemoveAdd<TObject, TKey>(
+		    this IObservable<ISortedChangeSet<TObject, TKey>> source)
+	    {
+		    if (source == null) throw new ArgumentNullException(nameof(source));
+
+		    return source.Select(changes =>
```
RolandPheasant: This code solves the problem - good work, thanks.

There can be a small optimisation here by abstracting the SelecyMany into an enumerator by using an enumerator

eg 

```cs
	    public static IObservable<ISortedChangeSet<TObject, TKey>> TreatMovesAsRemoveAdd<TObject, TKey>(
		    this IObservable<ISortedChangeSet<TObject, TKey>> source)
	    {
		    if (source == null) throw new ArgumentNullException(nameof(source));

		    IEnumerable<Change<TObject, TKey>> ReplaceMoves(IChangeSet<TObject, TKey> items)
		    {
			    foreach (var change in items)
			    {
				    if (change.Reason == ChangeReason.Moved)
				    {
					    yield return new Change<TObject, TKey>(
						    ChangeReason.Remove,
						    change.Key,
						    change.Current, change.PreviousIndex);

					    yield return new Change<TObject, TKey>(
						    ChangeReason.Add,
						    change.Key,
						    change.Current,
						    change.CurrentIndex);
				    }
				    else
				    {
					    yield return change;
				    }
			    }
		    }

		    return source.Select(changes => new SortedChangeSet<TObject, TKey>(changes.SortedItems, ReplaceMoves(changes)));
	    }
```

jasonwurzel: true, I'll change it

PR #116, opened 2018-05-08
Changes to allow compilation on VS 7.5+ for Mac

robsonj:
Reordering to put netstandard2.0 at the front of the target frameowkr list. This allows compilation and test execution to succeed on Visual Studio for ac (v7.5+) as VS for Mac picks the first target framework in the list to target currently. This fools it into picking netstandard2.0.

Also adjusted .gitignore to remove annoying OSX artifacts.

### Note in the DD.Test.csproj, does the 452 need replacing with 461 versus just removing?

### DIFF
--- a/DynamicData.Tests/DynamicData.Tests.csproj
+++ b/DynamicData.Tests/DynamicData.Tests.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.0;netcoreapp1.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp1.1;net461</TargetFrameworks>
     <NoWarn>$(NoWarn);CS0618</NoWarn>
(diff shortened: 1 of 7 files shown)


DynamicData.Tests/DynamicData.Tests.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net452;netcoreapp1.1;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp1.1</TargetFrameworks>
```
RolandPheasant: As per you comment on the PR, I suggest we include 461 in the test. 

Also the same should be applied to the corresponding section in DynamicData.csproj

```xml
 <PropertyGroup Condition="'$(TargetFramework)'=='net452'">
    <DefineConstants>P_LINQ</DefineConstants>
  </PropertyGroup>
```
Will need to change to match 461

PR #120, opened 2018-06-01
Avoid unnecessary tolist calls (performance optimization)

onliner10:
While playing around with profiler I noticed a lot of CPU spikes and memory allocations were caused by toList calls in KeyValueCollection constructor. After I made the list Lazy I was able to reduce my test case run time almost 3 times ( I'll also try to submit PR for my benchmarks, soon)

### DIFF
--- a/DynamicData/Cache/Internal/IndexCalculator.cs
+++ b/DynamicData/Cache/Internal/IndexCalculator.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 
 namespace DynamicData.Cache.Internal
@@ -12,7 +13,7 @@ namespace DynamicData.Cache.Internal
     internal sealed class IndexCalculator<TObject, TKey>
     {
         private KeyValueComparer<TObject, TKey> _comparer;
-        private List<KeyValuePair<TKey, TObject>> _list;
+        private ImmutableList<KeyValuePair<TKey, TObject>> _list;
 
         private readonly SortOptimisations _optimisations;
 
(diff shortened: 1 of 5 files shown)


RolandPheasant:
Good idea but there is an issue. Typically the Sort operator is used as follows:

```cs
.Sort(myComparer)
.ObserveOnDispatcher()
.Bind(myBindingTarget)
```

The list passed into ```KeyValueCollection``` is maintained internally and ```Bind``` is on a different thread. There is the possibility (which does occasionally happen) that a later change is being processed within the sort functionality whilst ```Bind``` is processing the result. This is a small window but ```ToList()``` can be enumerating the collection at the same time as the sort is processing. This leads to that dreaded "Collection was modified..." exception 

However I think your idea is great and there should be an easy solution:

If a lock object is passed into the constructor of ```KVC```, the lazy constructor can be modified to 

```
_items = Lazy<List<KeyValuePair<TKey, TObject>>>(() =>
            {
                lock (locker) { return  items.ToList();}
            });
```

And of course the lock object would need to be used inside the sort implementation. 

I never liked that ```ToList()``` and in a previous attempt to get rid of it I tried to implement an immutable sorted array but the performance was dire on busy systems. Now I think your idea coupled with my suggested modification will solve the problem.

glennawatson:
If you are going to have outside locks be introduced you may want to turn off the threading safe guards on the Lazy object since you'll have that protection anyway.

RolandPheasant:
Yes. Also a minor semantic improvement would be to pass ```Func<List<T>>``` into the constructor so the lock need not be exposed outside the sort class.

RolandPheasant:
This looks to be a clean and simple solution. 

The only consideration we need to be sure of is performance. Dynamic Data  lacks benchmarking and I think with changes like this it is essential to measure perf implications.

Would you be happy to help with adding a new benchmarking project to the solution. Probable best as another PR, or is you want maybe so so here

glennawatson:
Most projects i've seen that do benchmarking tend to use this guy https://github.com/dotnet/BenchmarkDotNet

RolandPheasant:
@glennawatson Agreed, BenchmarkDotNet is absolutely the go to tool for bench marking

PR #126, opened 2018-06-17
Use LinkedLists in IndexCalculator, optimise algorithm

onliner10:
IndexCalculator used to cause a lot of memory allocations mainly due to the fact it was using lists. So, for each insert/update it had to perform full array copy twice (RemoveAt, InsertAt) which is expensive. Since we use LinkedList now, this is no longer an issue. Also, list iterations were reduced significantly to a constant of three (one for removals, one for updates, one for adds).
To reduce complexity in IndexCalculator, I also created ChangesReducer which reduces N changes for a given key into 1, which should also alter performance in a positive way.
Additionaly, I modified SortFixture to sort by property instead of a key so that moves are possible.
I tested this solution throughly and, according to my knwoledge, it seems working


onliner10:
I managed to fix algorithm to handle refreshes, therefore fixing failing test

RolandPheasant:
I've merged this is because it looks good, there are lots of unit tests and on my local testing with dynamic trader there was a noticeable side by side performance improvement. 

What we desperately need now is some proper bench marking incorporated into dd

Issue #139, opened 2018-08-17
DistinctValues is not working correctly for items that should be reinserted.


owoudenberg:
When using a DistinctValues on a collection where same values are removed and added again, then re-adding items are not notified in the resulting collection.

for example:
```
var collection = new ObservableCollection<int>();

using (collection.ToObservableChangeSet()
                 .DistinctValues(value => value)
                 .Bind(out var distinctCollection)
                 .Subscribe())
{
    collection.AddRange(new[] { 1, 1, 2, 3 });
    Assert.That(distinctCollection, Is.EquivalentTo(new[] { 1, 2, 3 }));
    collection.Remove(2);
    Assert.That(distinctCollection, Is.EquivalentTo(new[] { 1, 3 }));
    collection.Add(2);
    Assert.That(distinctCollection, Is.EquivalentTo(new[] { 1, 3, 2 })); // <-- this doesn't assert as distinctCollection now contains { 1, 3 }
}
```

PR #140, opened 2018-08-17
Bugfix: DistinctValues didn't re-add an item that should reappear.

owoudenberg:
This is a fix to the issue described in #139


RolandPheasant:
Thanks for the fix

PR #142, opened 2018-08-17
Bugfix: When a changeset doesn't have indexes for removes, then the remove is not propogated

owoudenberg:
I found this bug after creating a combined list and noticed that in the changesets all indexes were set to -1.

### DIFF
--- a/DynamicData.Tests/List/TransformFixture.cs
+++ b/DynamicData.Tests/List/TransformFixture.cs
@@ -58,6 +58,23 @@ namespace DynamicData.Tests.List
         }
 
         [Fact]
+        public void RemoveWithoutIndex()
+        {
+            const string key = "Adult1";
+            var person = new Person(key, 50);
+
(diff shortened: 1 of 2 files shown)


owoudenberg:
I just realise that using ReferenceEquals is not the best solution when you are working with structs.
This is not an unlikely case when using structs, for example ValueTuples as they cab be more performant than classes.

glennawatson:
Potentially we would want to use EqualityComparer<T>.Default.Equals, maybe even passing in a equality comparer into the method and use the default if it's null.

owoudenberg:
I can see the reason for ReferenceEquals in case of classes, instead of EqualityComparer. As you want to search for the source item not by equality but by reference.
I like the idea of passing an equality comparer into the method.

RolandPheasant:
First of all thanks for the fix.

I agree with the points made about equality comparison. And it should indeed use EqualityComparer.Default. It would also be ideal to add an overload to specify the comparer.

Also there are several operators which could do with an equality comparer overload. It’s just one of those things that I have not got around doing.

PR #148, opened 2018-08-31
Add SourceLink

### LINKED ISSUE #147: Use SourceLink to enable debugging from the GitHub  code base
This would be cool. I need to do some research to find out how to do it. If anyone is able to help please do so.

ltrzesniewski:
Adds SourceLink support. Fixes #147.

Here's what I did:

- Added a `Directory.Build.props` file for the common properties across all projects. This is [automatically picked up by MSBuild](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2017).
- Cleaned up the csproj files to remove the properties I moved to `Directory.Build.props`
- Made the build deterministic on AppVeyor, because why not ¯\\\_(ツ)\_/¯
- Disabled SourceLink if you happen to use NCrunch (it just gets in the way)
- Added the pdb files to the NuGet and removed the symbol packages as they are now redundant.
- Set `IsPackable` to false as a default, and set it to true in the two projects that require it. This one is not really needed, but it may be *cleaner* this way.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+
+  <PropertyGroup>
+    <LangVersion>7.2</LangVersion>
+    <DefaultItemExcludes>$(DefaultItemExcludes);*.DotSettings;*.ncrunchproject</DefaultItemExcludes>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Authors>Roland Pheasant</Authors>
+    <Owners>Roland Pheasant</Owners>
+    <PackageLicenseUrl>http://opensource.org/licenses/MIT</PackageLicenseUrl>
+    <PackageProjectUrl>https://github.com/RolandPheasant/DynamicData</PackageProjectUrl>
+    <Copyright>Copyright Roland Pheasant 2011-2018</Copyright>
+    <PackageTags>DynamicData;Dynamic;Data;Rx;Reactive;Observable;Cache;Binding;ObservableCache;ObservableList;ObservableCollection;Collection;Linq</PackageTags>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(APPVEYOR)' == 'True'">
+    <Deterministic>true</Deterministic>
+    <DeterministicSourcePaths>true</DeterministicSourcePaths>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(NCrunch)' == '1'">
+    <EnableSourceLink>false</EnableSourceLink>
+    <EnableSourceControlManagerQueries>false</EnableSourceControlManagerQueries>
+  </PropertyGroup>
+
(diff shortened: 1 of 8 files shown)


RolandPheasant:
Thanks for such a fast and awesome response.  This PR rocks.

I will take a close look tomorrow.

Directory.Build.props
```diff
@@ -0,0 +1,36 @@
...
+  <PropertyGroup>
+    <Authors>Roland Pheasant</Authors>
+    <Owners>Roland Pheasant</Owners>
+    <PackageLicenseUrl>http://opensource.org/licenses/MIT</PackageLicenseUrl>
+    <PackageProjectUrl>https://github.com/RolandPheasant/DynamicData</PackageProjectUrl>
+    <Copyright>Copyright Roland Pheasant 2011-2018</Copyright>
+    <PackageTags>DynamicData;Dynamic;Data;Rx;Reactive;Observable;Cache;Binding;ObservableCache;ObservableList;ObservableCollection;Collection;Linq</PackageTags>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
```
glennawatson: Probably want <GenerateDocumentationFile>true</GenerateDocumentationFile> here

ltrzesniewski: On `GeneratePackageOnBuild` ? It was false previously and I left it that way. The `appveyor.yml` script packs the NuGet explicitly in a separate step.

This can be changed, of course. I'd suggest this if you want to simplify the build script:

```XML
<GeneratePackageOnBuild>False</GeneratePackageOnBuild>
<GeneratePackageOnBuild Condition="'$(Configuration)' == 'Release'">True</GeneratePackageOnBuild>
```

glennawatson: Just helps with intellsense normally.

ltrzesniewski: Ooh, `GenerateDocumentationFile`. I didn't notice you've edited your comment.

If you move it to `Directory.Build.props`, you'll get a *ton* of warnings:

> warning CS1591: Missing XML comment for publicly visible type or member ...

So let's keep `GenerateDocumentationFile` in the two projects that need it. No need to document tests anyway 😉

RolandPheasant:
One further observation, I have noticed System.Reactive have the following 2 properties set in Directory.build.props

```xml
<DebugType>embedded</DebugType>
<EmbedUntrackedSources>true</EmbedUntrackedSources>
```

Could these being omitted be the issue?

ltrzesniewski:
@RolandPheasant weird.

I just tried it with that package and it worked for me with the `netstandard2.0` target:

image

But it failed with the `net46` target.

You don't need `EmbedUntrackedSources`. That includes files which are not tracked by git, and I don't think you have any of those.

I think you might need to use the portable PDB format for SourceLink to work. I'll find out.

Thanks for pointing it out, I'm pretty sure I must have the same issue in one of my multi-targeted libs.

RolandPheasant:
Again thanks for your help, and I appreciate the effort which you are making

glennawatson:
Be the debugtype embedded why it works for Rx.net https://github.com/ctaggart/SourceLink/issues/155

ltrzesniewski:
@glennawatson thanks for the info, that confirms what I thought. I wonder why embedded is preferred over portable though. Do you have any info on this? I'd prefer to use portable over embedded.

ltrzesniewski:
@RolandPheasant OK I confirmed it works with portable PDBs. I don't really like the idea of embedded debug info, separate debug info seems better to me so I chose that.

I made a one-line PR: #149, and now I'll push the same fix to my other projects. 😄

PR #149, opened 2018-09-03
Use portable PDBs to make SourceLink work on the .NET Framework

ltrzesniewski:
Fixes https://github.com/RolandPheasant/DynamicData/pull/148#issuecomment-418180968

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,6 +4,7 @@
   <PropertyGroup>
     <LangVersion>7.2</LangVersion>
     <DefaultItemExcludes>$(DefaultItemExcludes);*.DotSettings;*.ncrunchproject</DefaultItemExcludes>
+    <DebugType>portable</DebugType>
   </PropertyGroup>
 
   <PropertyGroup>


Directory.Build.props
```diff
@@ -4,6 +4,7 @@
   <PropertyGroup>
     <LangVersion>7.2</LangVersion>
     <DefaultItemExcludes>$(DefaultItemExcludes);*.DotSettings;*.ncrunchproject</DefaultItemExcludes>
+    <DebugType>portable</DebugType>
```
RolandPheasant: For my understanding, why did you choose portable and not embedded?

ltrzesniewski: That's more of a personal preference. That way you can actually choose if you want to ship the debug info or not along with the binaries. This could matter for people who try to minimize their product's download size.
Both options should work the same, functionally speaking.

RolandPheasant: Ok, let's run with this and we can always change to embedded if required

RolandPheasant:
@ltrzesniewski I am about to do a release and before doing so tried to debug from Dynamic.Trader. However with portable, the symbols are not loaded. I am tempted to change it to embedded.

What's your view?

ltrzesniewski:
Ugh. That's interesting. I'd like to know why it doesn't work. What version did you test? v6.5.0.2468? I'll try to reproduce.

I'm fine with changing the symbols to embedded if you think that's better though.

RolandPheasant:
I am using that version. Before doing anything I will let you take a look

ltrzesniewski:
Ok, I **swear** it worked 3 days ago. But right now I can't get it to work with portable PDBs *at all*, not even in my test project. VS just says it can't find the PDB file.

I honestly have no clue what's going on, but it's time to switch to embedded, I guess.
 Sorry for that mess. Fixed in #151.

RolandPheasant:
It's no mess whatsoever. I am delighted with your awesome contribution

Issue #158, opened 2018-10-04
TransformMany does not work with IObservableList


ronaldvdv:
I'm creating a Parent class that exposes a "Children" property of type `IObservableList<TChild>`. Then when I have an `IObservable<Parent>`, I would like to create another observable that represents the relation to the children such that a change of the Parent or a change in the collection of children are propagated.

For this I would like to use `parentObservable.ToObservableChangeSet().TransformMany(parent => parent.Children)` but as far as I can see, the TransformMany does not contain an overload that recognizes a `Func<TSource, IObservableList<TDestination>>`. Therefore the `Func<TSource, IEnumerable<TDestination>>` is used instead which does not propagate changes in the child collection.

Sample code:

```
class Parent {
  private SourceList<Child> _children;
  public IObservableList<Child> Children => _children.AsObservableList();
}

class Child { }

IObservable<Parent> parent; // Set elsewhere
var subscription = parent.ToObservableChangeSet().TransformMany(p => p.Children).Bind(out list).Subscribe();
```

RolandPheasant:
The desired result can be achieved  using the logical collection operations

```.Transform(p => p.Children).Or();```

See [Composite Collections](https://github.com/RolandPheasant/DynamicData/wiki/Composite-Collections)

ronaldvdv:
Thanks for following up on this! However, I struggle to get your solution to work:

- ToObservableChangeSet gives a result of type IObservable<IChangeSet<Parent>> 
- Transform then gives a result of type IObservable<IChangeSet<IObservableList<Child>>>
- Or then gives me the same result type

Also looking at the source code (List/ObservableListEx.cs), I cannot find overloads for the Or operator that accept IObservable<IChangeSet<IObservableList<Child>>> and return IObservable<IChangeSet<Child>> ; they either combine multiple sources, or expect an observable list as the outer type.

Did I misunderstand your solution or would I need a different approach to get this to work?

Also, looking at the source code for TransformMany, it seems odd that there is no overload that accepts a manyselector of type Func<TSource, IObservableList<TDestination>>  ... wouldn't that be the most straight-forward approach? Happy to help out with a PR if you agree this is the better approach instead.

RolandPheasant:
Change 

```_children.AsObservableList```

to

```_children.Connect()```

RolandPheasant:
Also a PR for the other overload would be welcome

ronaldvdv:
[...]

PR #160, opened 2018-10-06
Allow TransformMany with a Func<TSource, IObservableList<TDestination>>

ronaldvdv:
Hi!

This PR solves #158 
- Added new constructor for TransformMany and corresponding overload in ObservableListEx
- Duplicated and adjusted the existing unit tests "TransformManyObservableCollectionFixtures" to test the new overload

Note: I'm not sure the "CreateCollectionFromList" method is the nicest way, especially since it cannot store the subscription to the child changes anywhere. I'm thinking if a larger change is needed: If there would be a more flexible ManyContainer, perhaps subclassed separately for the existing and the new use case, the subscription could be managed there. For that I would like some feedback from you before digging deeper.

Thanks!

### DIFF
--- a/DynamicData/List/Internal/TransformMany.cs
+++ b/DynamicData/List/Internal/TransformMany.cs
@@ -50,6 +50,28 @@ namespace DynamicData.List.Internal
         {
         }
 
+        private static ReadOnlyObservableCollection<TDestination> CreateCollectionFromList(IObservableList<TDestination> list)
+        {
+            ReadOnlyObservableCollection<TDestination> result = null;
+            list.Connect().Bind(out result).Subscribe();
+            return result;
+        }
+
+        public TransformMany(IObservable<IChangeSet<TSource>> source,
+            Func<TSource, IObservableList<TDestination>> manyselector,
+            IEqualityComparer<TDestination> equalityComparer = null)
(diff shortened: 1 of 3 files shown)


DynamicData/List/Internal/TransformMany.cs
```diff
@@ -50,6 +50,28 @@ internal sealed class TransformMany<TSource, TDestination>
         {
         }
 
+        private static ReadOnlyObservableCollection<TDestination> CreateCollectionFromList(IObservableList<TDestination> list)
```
RolandPheasant: I agree with your comment on the PR. There must be a cleaner way to achieve the goal, without any need to use bind.

ronaldvdv: I was afraid so ;-) Does the proposal above make sense? Happy to spend some time on this, but since I'm relatively new to this library and Rx in general, I could use a little bit of help :-) Thanks!

RolandPheasant: I will get involved on this one as there is clearly a missing transform many overload for both list and cache parts or the system. The import thing is not to resolve the issue quickly but to find a good solution with good performance characteristics

ronaldvdv: Makes sense, thanks for that! If there's anything I can do, let me know.

Issue #164, opened 2018-10-16
Dynamic Filter with SourceList


Liklainy:
Hello!
I am new to DynamicData and cannot understand why there is no update after removing the items from filtered SourceList.

```csharp
using DynamicData;
using System;
using System.Collections.Generic;
using System.Linq;
using System.Reactive.Linq;
using System.Reactive.Subjects;

namespace DynamicDataTests
{
    class UserPermission
    {
        public int Id;
        public int UserId;
        public int RoleId;
    }

    class Program
    {
        static void Main(string[] args)
        {
            var subject = new Subject<HashSet<int>>();
            var dynamicFilter = subject.Select(ids => (Func<UserPermission, bool>)(x => ids.Contains(x.UserId)));

            var list = new SourceList<UserPermission>();

            list.Connect()
                //.Filter(dynamicFilter)
                .ToCollection()
                .Subscribe(x => Console.WriteLine($"Changed: {string.Join(",", x.Select(up => up.Id))}"));

            list.Edit(innerList =>
            {
                innerList.Add(new UserPermission { Id = 0, UserId = 0, RoleId = 0 });
                innerList.Add(new UserPermission { Id = 1, UserId = 0, RoleId = 0 });
                innerList.Add(new UserPermission { Id = 2, UserId = 0, RoleId = 1 });
                innerList.Add(new UserPermission { Id = 3, UserId = 0, RoleId = 1 });
                Console.WriteLine($"Items in collection: {innerList.Count}");
            });

            subject.OnNext(new HashSet<int>(new[] { 0 }));

            list.Edit(innerList =>
            {
                var toRemove = innerList.Where(x => x.RoleId == 1 && x.UserId == 0);
                Console.WriteLine($"Remove: {string.Join(",", toRemove.Select(up => up.Id))}");
                innerList.RemoveMany(toRemove);
                Console.WriteLine($"Items in collection: {innerList.Count}");
            });

            Console.ReadKey();
        }
    }
}
```
What I got:
```
Items in collection: 4
Changed: 0,1,2,3
Remove: 2,3
Items in collection: 2
```
What I expected to see:
```
Items in collection: 4
Changed: 0,1,2,3
Remove: 2,3
Items in collection: 2
Changed: 0,1
```
Why there is no update after removing items?

RolandPheasant:
It looks to be working for me. `ToCollection()` materialises inner contents and does not report changes. For the actual changes you need to remove this operator.

[...]

PR #167, opened 2018-10-20
Fix list's filter

Liklainy:
There is an error with dynamic filter applied to SourceList - #164 

The error exists because `IsMatch` might be not actual in `filteredResult`.
Transform is not called on remove, and is not called when filter is changed, so internally it saves previous state of the matched item, which is returned on remove.
And if it was updated on remove, `IsMatch` might be set to `false` because item was internally changed and must be filtered off, but item will not be removed.

The fast fix is to remove the optimization check on `IsMatch` when remove is called.


RolandPheasant:
Thanks for the effort with the investigation and fix

PR #168, opened 2018-10-23
Fix list's connect

Liklainy:
Predicate in the `Connect` method doesn't work as expected.
Internal `Filter` is not initialized properly when there were items in the list.
So removing items in some cases can cause an out of bounds exception in `Transform`.
There is a fast fix, which moves `Filter` outside.

### DIFF
--- a/DynamicData.Tests/List/FilterFixture.cs
+++ b/DynamicData.Tests/List/FilterFixture.cs
@@ -224,5 +224,18 @@ namespace DynamicData.Tests.List
             _results.Data.Count.Should().Be(0, "Should nothing cached");
         }
 
+        [Fact]
+        public void AddSubscribeRemove()
+        {
+            var people = Enumerable.Range(1, 100).Select(l => new Person("Name" + l, l)).ToArray();
+            var source = new SourceList<Person>();
+            source.AddRange(people);
+
+            var results = source.Connect(x => x.Age > 20).AsAggregator();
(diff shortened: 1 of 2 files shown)


Liklainy:
@RolandPheasant, can you, please, assist on this?
There is a test in the first commit which fails.
I think, the correct fix is to remove predicate from `Connect`, but it will break existing API.

RolandPheasant:
The original API having the filter provides an optimisation for large record sets. I know for sure some people depend on this. I wonder whether the issue can be fixed in the filter and transform methods.

Liklainy:
Will it be a better way to add new special filter without transform?
Which will not track item's IsMatch, and it will only check predicate when item is added.

RolandPheasant:
It probably would do. Accounting for every permutation is what makes DD so difficult to implement correctly.

Liklainy:
Added `FilterStatic`
Based on Filter, but I removed Refresh case and changed Replace.
Predicated is checked when item is added or replaced.

Is it enough?
Or there is a need to save `filtered` items by wrapping `ItemWithIndex` from the base list?
(to change by index and not by item)

RolandPheasant:
I am inclined to leave it as simple as possible. At least for now anyway.

I will sit on this for a day or two before merging as dealing with list collections, it is very easy to omit permutations.  That way I will have time to think.

Additionally I intend to have a round of optimisation for the list side of DD soon and I think filter is a prime candidate for me to look at with Greg's eyes.

PR #174, opened 2018-11-01
Fix TransformMany

### LINKED ISSUE #173: TransformMany with Refresh
Hello, I have an issue with TransformMany:
```csharp
[Fact]
public void Refresh()
{
    var relations = new List<Person> { new Person("Child1", 1) };
    var parent = new PersonWithChildren("parent", 50, relations);
    _source.AddOrUpdate(parent);

    relations.Add(new Person("Child2", 2));
    _source.Refresh(parent); //_source.AddOrUpdate(parent);

    _results.Data.Count.Should().Be(2, "Should be 2 in the cache");
    _results.Data.Lookup("Child1").HasValue.Should().BeTrue();
    _results.Data.Lookup("Child2").HasValue.Should().BeTrue();
}
```
The test will fail.
Because TransformMany on `Refresh` doesn't add or remove items it only propagates `Refresh`.
TransformMany internally uses Transform with `transformOnRefresh = false` by default.

I can see that changing `Relations` from `IEnumerable` to `ObservableCollection` will fix it and there will be no need for explicit `Refresh`.
```csharp
public ObservableCollection<Person> Relations { get; }
parent.Relations.Add(new Person("Child2", 2));
```

Are there other ways to do it?
Probably add `transformOnRefresh` parameter to TransformMany?

Liklainy:
Fixes #173
Added `Transform` interface with `transformOnRefresh` to cache extensions (similar to list)
Changed TransformMany in SourceCache:
- enable `transformOnRefresh` if without `childChanges`
- added `ToList()` to force calculate `Select`, else Previous items and Current items will be equal
- in case when there is `childChanges` there was strange applying of `lock` - `Select` was iterated outside locked area

Do I need to make `transformOnRefresh` optional for TransformMany and add it to interface?
Probably similar stuff must be done for SourceList.

I think, that I have messed up with `lock`

### DIFF
--- a/DynamicData.Tests/Cache/TransformManyRefreshFixture.cs
+++ b/DynamicData.Tests/Cache/TransformManyRefreshFixture.cs
@@ -1,0 +1,79 @@
+﻿using DynamicData.Tests.Domain;
+using FluentAssertions;
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace DynamicData.Tests.Cache
+{
+
+    public class TransformManyRefreshFixture: IDisposable
+    {
+        private readonly ISourceCache<PersonWithFriends, string> _source;
+        private readonly ChangeSetAggregator<PersonWithFriends, string> _results;
+
+        public TransformManyRefreshFixture()
+        {
+            _source = new SourceCache<PersonWithFriends, string>(p => p.Key);
+
+            _results = _source.Connect()
+                .AutoRefresh()
+                .TransformMany(p => p.Friends, p => p.Name)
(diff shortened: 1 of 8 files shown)


Liklainy:
Added fix for list's `TransformMany`

For `SourceList`:
Does list's `ManyContainer` need the same logic by wrapping destination in `Func` is needed?
For now iterating `IEnumerable` is done without `lock`.

RolandPheasant:
Super awesome effort. Much appreciated. I will be back from vacation soon and will take a very detailed look at this.

Issue #180, opened 2018-11-08
AutoRefreshOnObservable followed by Transform does not trigger a refresh


samirem:
The `AutoRefreshOnObservable` does not trigger a refresh in the following snippet
```
MyObservableCache.Connect()
    .AutoRefreshOnObservable(_ => refreshObservable)
    .Transform(x =>...)
    .AsObservableCache();
```
I've tried setting breakpoints and it hits the one in the `refreshObservable` but not the `Transform`.

Tried with version 6.6.1.2516 of the library.

grokys:
I too am having this problem. #185 was closed saying:

>  You can use the all items observable as it is but instead of using the `AutoRefresh` operator, call the `Refresh` method on the source cache

However I am using a `SourceList` and that doesn't have a `Refresh` method. What should I do in this case?

Claye-L:
The problem mentioned in the first post is easily solved by doing ```.Transform(x => someItem, true)``` which enables the "transformOnRefresh".

grokys:
Ah yes, `transformOnRefresh` helps for the `SourceList` case too, although I assume it's still causing a subscription for each item which isn't ideal performance-wise.

PR #185, opened 2018-11-19
AutoRefreshOnObservable fails when items share a subscription

samirem:
A complement to issue #180. Added test to show that `AutoRefreshOnObservable` fails when items share a subscription

### DIFF
--- a/DynamicData.Tests/Cache/AutoRefreshFixture.cs
+++ b/DynamicData.Tests/Cache/AutoRefreshFixture.cs
@@ -94,6 +94,53 @@ namespace DynamicData.Tests.Cache
         }
 
         [Fact]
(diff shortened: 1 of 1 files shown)


samirem:
You're right that it's inefficient, but basically what I want is a `.ResetOnObservable(...)` type of statement where the entire collection is forced to reset/refresh.

My specific use case is that I store a lot of calculated results, and whenever some input data to that calculation changes, all results are invalidated and has to be refreshed/reset. Do you think that's a valid use case or could I solve it in a better way?

RolandPheasant:
There is a very efficient way of doing it.  You can use the all items observable as it but instead of using the ```AutoRefresh``` operator, call the ```Refresh``` method on the source cache

PR #193, opened 2019-01-09
Add overload for TransformMany to support IObservableList

ronaldvdv:
- Added an extra constructor in TransformMany that accepts a `Func<TParent, IObservableList<TChild>>` as its `manyselector`. Since we want to be able to call the original `manyselector` again when the parent is removed (instead of remembering the value of `manyselector(x).Items` when the parent was added), I also added a small `ManySelectorFunc` class that captures the original selector.
- Added an overload of the TransformMany extension method too
- Duplicated and fixed the unit test for TransformMany to also test the new overloads

### DIFF
--- a/DynamicData.Tests/List/TransformManyObservableCollectionFixture.cs
+++ b/DynamicData.Tests/List/TransformManyObservableCollectionFixture.cs
@@ -7,8 +7,7 @@ using FluentAssertions;
 using Xunit;
 
 namespace DynamicData.Tests.List
-{
-    
+{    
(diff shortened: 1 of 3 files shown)


ronaldvdv:
Solves #158

RolandPheasant:
Thanks for this. I will leave the issue open as I will add some optimisations. Additionally I will add the same operator for the cache side of dd

ronaldvdv:
Sounds good, thanks!

PR #196, opened 2019-02-04
Expire after bug

### LINKED ISSUE #187: ExpireAfter in ToObservableChangeSet with a key does only remove items once
I believe I have found a bug in the ToObservableChangeSet operator.

It seems that objects are removed ony once if the expireafter operator is specified. This affects only the overload with a key.

Since the problem is hard to explain, I added some Unit Tests to a branch here:

https://github.com/Peter-B-/DynamicData/tree/ExpireAfterBug

The unit test called ExpireAfterTimeDynamicWithKey fails and shows the behaviour.

I also creaded an LinqPad script to see the bug in action:

http://share.linqpad.net/d33j3u.linq

In this query, new objects should be added every 0.1 seconds and be removed after 1 second. The count should never exceed 10 (or maybe 11). But you can see, that only one element is removed and from then on the count continues to grow.

I tried to find a solution to the problem, but it seems I am not fluent enough with RX to fix that.

Peter-B-:
This PR is an attempt to resolve #187. 

I think the root cause of this issue was, that MergeMany propagates OnCompleted from all items to the outside observer. So when the first item's remove is executed and the observable started with Observable.Timer is completed, the overall stream is completed an the remove of subsequent items will not be executed.

Please review my changes to MergeMany. I am ignoring OnCompleted and OnError and I am not sure, if this is the correct thing to do.

### DIFF
--- a/DynamicData.Tests/Cache/MergeManyWithKeyOverloadFixture.cs
+++ b/DynamicData.Tests/Cache/MergeManyWithKeyOverloadFixture.cs
@@ -25,6 +25,16 @@ namespace DynamicData.Tests.Cache
                 _changed.OnNext(value);
             }
 
+            public void CompleteObservable()
+            {
+                _changed.OnCompleted();
+            }
+
+            public void FailObservable(Exception ex)
+            {
+                _changed.OnError(ex);
+            }
+
             public IObservable<bool> Observable => _changed.AsObservable();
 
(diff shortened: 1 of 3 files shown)


RolandPheasant:
You broke nothing. My thought is whether the issue can also exist with subscribe many. Accounting for your comments it almost certainly does not

PR #205, opened 2019-02-13
Add FilterOnObservable

oysteinkrog:
New operator that I based on AutoRefresh that allows for filtering on an object's observable property.
This is unlike the existing Filter which only works for normal properties or INPC properties.
I am not completely happy with the implementation, it generates N refresh changes for N object, if anyone has an idea for how to prevent this it would be great :)

### DIFF
--- a/DynamicData.Tests/Domain/PersonObs.cs
+++ b/DynamicData.Tests/Domain/PersonObs.cs
@@ -1,0 +1,122 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reactive.Subjects;
+using DynamicData.Annotations;
+using DynamicData.Binding;
+
+namespace DynamicData.Tests.Domain
+{
+    public class PersonObs : IEquatable<PersonObs>
+    {
+        public string ParentName { get; }
+        public string Name { get; }
+        public string Gender { get; }
+        public string Key => Name;
+        [NotNull]
+        private readonly BehaviorSubject<int> _age;
+
+        public PersonObs(string firstname, string lastname, int age, string gender = "F", string parentName = null)
+            : this(firstname + " " + lastname, age, gender, parentName)
+        {
+        }
+
+        public PersonObs(string name, int age, string gender = "F", string parentName = null)
+        {
+            Name = name;
+            _age = new BehaviorSubject<int>(age);
+            Gender = gender;
+            ParentName = parentName ?? string.Empty;
+        }
+
+        public IObservable<int> Age => _age;
+
+        public void SetAge(int age)
+        {
+            _age.OnNext(age);
+        }
+
+        #region Equality Members
+
+        public bool Equals(PersonObs other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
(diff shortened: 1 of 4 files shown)


oysteinkrog:
Updated to fix merge conflicts.

RolandPheasant:
The reason for there being so many updates is the way that the observable values are received.  Inherent to the nature of observables you cannot predict when you are going to received a value. Subsequently when each item is subscribed to a value is produced and the filter is applied.  Particularly for the initial load, it is performance intensive as the number of change set notifications is multiplied by the number items in the cache which clearly is not scaleable.  

The problem goes away when you apply the buffer. If you change the test class to something like this:

```cs
private class FilterPropertyStub : IDisposable
{
    public ISourceList<PersonObs > Source { get; } = new SourceList<PersonObs >();
    public ChangeSetAggregator<PersonObs > Results { get; }

    public TestScheduler Scheduler { get; } = new TestScheduler();

    public FilterPropertyStub()
    {
        Results = new ChangeSetAggregator<PersonObs >(Source.Connect().FilterOnObservable(p => p.Age, (o, p) => p > 18,TimeSpan.FromMilliseconds(250),Scheduler));
    }

    public void Dispose()
    {
        Source.Dispose();
        Results.Dispose();
    }
}
```

then this test passes.

```cs
[Fact]
public void InitialValues()
{
    var people = Enumerable.Range(1, 100).Select(i => new PersonObs ("Name" + i, i)).ToArray();
    using (var stub = new FilterPropertyStub())
    {
        stub.Source.AddRange(people);
        stub.Scheduler.AdvanceBy(TimeSpan.FromMilliseconds(250).Ticks);

        1.Should().Be(stub.Results.Messages.Count);
        82.Should().Be(stub.Results.Data.Count);
        stub.Results.Data.Items.ShouldAllBeEquivalentTo(people.Skip(18));
    }
}
```

So the question is for this operator should we enforce a buffer? From a pure design approach I would rather not impose restrictions on consumers. However without doing so in this case I am sure there would be many support issues raise.

oysteinkrog:
Yeah I don't know if we should enforce a buffer or not, I feel like if we do there should be overloads for all the Buffer versions, not just the time-based one.
I think it might be possible to optimize the implementation a bit. It seems to me that many of the refresh messages are redundant, e.g. when the property observable/filter always returns true there should not be any refresh at all (since the default for the filter operator is of course to only refresh/remove filtered items).

RolandPheasant:
One solution to avoid enforcing a buffer would be to pass a flag in to indicate that the observable always returns and initial value (or not). That way, logic can be applied accordingly. I suggest this flag is mandatory to enforce consumers to think about the implications. This would need to be coupled with good comments on the extension.

As for optimisations of the operator let's leave that until we decide on the API

oysteinkrog:
I would argue that if the observable does not return a value, the operator should _not_ filter it.

oysteinkrog:
I pushed a commit that both simplifies the API.
The operator now takes just ```Func<TObject, IObservable<bool>>``` instead of two arguments (property selector and filter) and I was also able to change the implementation to prevent generating refresh messages for no-op filter updates.

PR #211, opened 2019-03-03
TwoStageRWLock for lists

Wouterdek:
This pull request is a smaller chunk of PR #199. It implements the new locking mechanism for lists only. See PR #199 for more details.

### DIFF
--- a/DynamicData/Kernel/InternalEx.cs
+++ b/DynamicData/Kernel/InternalEx.cs
@@ -103,5 +103,12 @@ namespace DynamicData.Kernel
                 scheduleNext(next);
             });
         }
-    }
+
+        public static void Swap<TSwap>(ref TSwap t1, ref TSwap t2)
+        {
+	        TSwap temp = t1;
+	        t1 = t2;
+	        t2 = temp;
+        }
+	}
 }
--- a/DynamicData/Kernel/InternalEx.cs
+++ b/DynamicData/Kernel/InternalEx.cs
@@ -104,11 +104,11 @@ namespace DynamicData.Kernel
             });
         }
 
-        public static void Swap<TSwap>(ref TSwap t1, ref TSwap t2)
+        internal static void Swap<TSwap>(ref TSwap t1, ref TSwap t2)
         {
-	        TSwap temp = t1;
-	        t1 = t2;
-	        t2 = temp;
+            TSwap temp = t1;
+            t1 = t2;
+            t2 = temp;
         }
-	}
+    }
 }
--- a/DynamicData/Kernel/TwoStageRWLock.cs
+++ b/DynamicData/Kernel/TwoStageRWLock.cs
@@ -1,149 +1,151 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Text;
 using System.Threading;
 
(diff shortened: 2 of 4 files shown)


RolandPheasant:
I am in the process of adding bench marks to dd. When I am done with that I will run them against this, then assuming the is no significant performance hit will merge this PR.

RolandPheasant:
I have added some benchmarks to dd and compared the performance difference between the previous locking mechanism and the changes here. The good news is I see is no significant difference is performance. So let's get this in. First I will add some review notes.

DynamicData/Kernel/TwoStageRWLock.cs
```diff
@@ -0,0 +1,149 @@
...
+	/// <summary>
+	/// Similar to ReaderWriterLockSlim, but still allows reading after disposal.
+	/// All operations involving entering and exiting the lock in read mode still work after disposal.
+	/// All other operations will throw an NullReferenceException.
+	/// </summary>
+	/// <remarks>
+	/// Before disposing this object, all locks should be released.
+	/// Failure to do so will result in a SynchronizationLockException.
+	/// </remarks>
+	public class TwoStageRWLock : IDisposable
```
RolandPheasant: This should be made internal otherwise I may end up having to support it to the public at large

DynamicData/Kernel/TwoStageRWLock.cs
```diff
@@ -0,0 +1,149 @@
+﻿using System;
+using System.Collections.Generic;
```
RolandPheasant: using System.Collections.Generic;
using System.Text;

are not used, so please remove

DynamicData/Kernel/InternalEx.cs
```diff
@@ -103,5 +103,12 @@ public static IDisposable ScheduleRecurringAction(this IScheduler scheduler, Fun
                 scheduleNext(next);
             });
         }
-    }
+
+        public static void Swap<TSwap>(ref TSwap t1, ref TSwap t2)
```
RolandPheasant: Should be internal

DynamicData/Kernel/TwoStageRWLock.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Text;
+using System.Threading;
+
+namespace DynamicData.Kernel
```
RolandPheasant: Although it does not show in this editor, can you remove tabs. This is easily achieved in vs. Select all text, Edit->Advanced->Untabbify Selected Lines. Please do the same for all the files which you have changed.

NB: I have never set up fx cop rules because I find it so intrusive and slows down editing.

Wouterdek:
> I have added some benchmarks to dd and compared the performance difference between the previous locking mechanism and the changes here. The good news is I see is no significant difference is performance. So let's get this in. First I will add some review notes.

Great! I have addressed your comments in my latest commit. Let me know if there is anything else that needs tweaking.

RolandPheasant:
Thanks. I suggest 2 more PRs. The first to add the new lock to source cache and the second to add Preview to list and cache.

I am sure what you are doing will be a great contribution to this library.

PR #222, opened 2019-04-25
Implementing AutoRefreshAllExcept and WhenAnyPropertyChangedExcept

sa-he:
This is an attempt to implement AutoRefreshAllExcept and WhenAnyPropertyChangedExcept as wished in this issue: https://github.com/RolandPheasant/DynamicData/issues/214

Sorry, my Visual Studio created a bunch of unnecessary white-space changes :|

### DIFF
--- a/DynamicData.Tests/List/AutoRefreshFixture.cs
+++ b/DynamicData.Tests/List/AutoRefreshFixture.cs
@@ -9,7 +9,6 @@ using Xunit;
 
 namespace DynamicData.Tests.List
 {
-    
     public class AutoRefreshFixture
     {
         [Fact]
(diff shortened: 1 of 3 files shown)


sa-he:
@RolandPheasant do you need more than I provided to accept the merge request? Change in a readme file or anything? Reverting the white-space changes?

RolandPheasant:
I've been busy and forgetful recently. I will take a look at this later today.

RolandPheasant:
A better solution would be to add an overload which allows the consumer to filter on the property.

Something like this

```cs
        public static IObservable<TObject> WhenAnyPropertyChanged<TObject>([NotNull] this TObject source, Func<string, bool> propertyFilter)
            where TObject : INotifyPropertyChanged
```

That way allows complete flexibility for filtering on property names.

This extra overload would also have to be added to source list and cache when property changed operators.

PR #242, opened 2019-07-09
Move files from base directory to /src directory

glennawatson:
This will make it easier for azure devops in the next round of changes.


glennawatson:
@RolandPheasant do this review first, it will preserve your file history and also make the other review dramatically easier.

RolandPheasant:
I am happy for the files to be moved. It is only structured like it was for historic reasons

PR #250, opened 2019-07-11
Add IsEmpty / NotEmpty overloads to CountEx

modplug:

It adds overloads to check if the underlying collection has items or not.

The same functionality can be achieved by doing 
```collection.Connect().Count().Select(count => count == 0) ```

**What is the new behavior?**

The new overload adds these two overloads
```collection.Connect().IsEmpty() ```
```collection.Connect().NotEmpty()```

**What might this PR break?**

Nothing as far as I can see

**Other information**:

### DIFF
--- a/src/DynamicData/Aggregation/CountEx.cs
+++ b/src/DynamicData/Aggregation/CountEx.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for full license information.
 
 using System;
+using System.Reactive.Linq;
 
 namespace DynamicData.Aggregation
 {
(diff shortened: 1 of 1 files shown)


wdkr:
Sounds like useful operators to have. I'm quite new to DynamicData so I don't know the API that well, but wouldn't an Any() method make sense instead of IsEmpty() / NotEmpty()?

It's what I'd look for anyway, since I'm used to the LINQ operator: https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.any?view=netstandard-2.0

glennawatson:
If it went down the any route you would have to add the lambda form I guess.

Issue #264, opened 2019-08-07
[BUG] Removing an item from an ObservableList with TrnasformMany and a projection


dchaib:
**Describe the bug**
As shown in RolandPheasant/DynamicData.Snippets#5, I have set up an `ObservableList` with `TransFormMany` and a projection inside. When I remove an item from the `ObservableList`, it is not removed from the output. It works with an `ObservableCache`.

As Roland guessed, the projected item is not removed from the list because it is not found. I believe the culprit is here: https://github.com/RolandPheasant/DynamicData/blob/67367d0068531a94c9ec8a3e197a0b0055f5eca4/DynamicData/List/ChangeAwareList.cs#L519. The call to IndexOf is not using the equalityComparer passed when using TransformMany. I haven't found a way to pass it properly (yet).

**Steps To Reproduce**
Run the unit tests in https://github.com/RolandPheasant/DynamicData.Snippets.

I'm also happy to send a PR with unit tests here.

**Expected behavior**
The items should be removed from the output.

**Screenshots**
n/a

**Environment**
- OS: Windows
- Version: tested in 6.8 (the issue is still present in the master branch today)
- Device: n/a

**Additional context**

glennawatson:
could use https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.findindex?redirectedfrom=MSDN&view=netframework-4.8#System_Collections_Generic_List_1_FindIndex_System_Predicate__0__ and use the equality comparer

dchaib:
The issue is more how to pass the predicate to `ChangeAwareList` than using it ;-)

dchaib:
I may have found something. If that's OK, I'll submit a PR tomorrow and we can see if I'm headed in the right direction.

RolandPheasant:
A PR is welcome. Thanks

dchaib:
I can confirm that this is fixed in 6.13.1 by #265.

PR #265, opened 2019-08-09
Remove from transform many with projection

dchaib:
Bug fix for #264 

When removing an item from a `SourceList` with a `TransformMany` transfomartion using a projection, the items are never removed from the "output". See #264 for more details.

**What is the new behavior?**
`IChangeSet` now has an equality comparer that can be used when looking for items to remove.

**What might this PR break?**
Hopefully nothing!

**Other information**:
This is a draft. If I'm headed in the right direction, I'll refactor and add proper comments when needed.
It might also be necessary to extend the usage of the newly added equality comparer for consistency and maybe fixing other cases.

### DIFF
--- a/src/DynamicData/List/ChangeSet.cs
+++ b/src/DynamicData/List/ChangeSet.cs
@@ -17,6 +17,7 @@ namespace DynamicData
     /// <typeparam name="T">The type of the object.</typeparam>
     public class ChangeSet<T> : IChangeSet<T>
     {
+        private readonly IEqualityComparer<T> _equalityComparer;
         private int _adds;
         private int _removes;
         private int _replaced;
@@ -40,8 +41,9 @@ namespace DynamicData
         /// Initializes a new instance of the <see cref="ChangeSet{T}" /> class.
         /// </summary>
         /// <param name="items">The items.</param>
+        /// <param name="equalityComparer">An optional equality comparer.</param>
         /// <exception cref="System.ArgumentNullException">items</exception>
-        public ChangeSet([NotNull] IEnumerable<Change<T>> items)
+        public ChangeSet([NotNull] IEnumerable<Change<T>> items, IEqualityComparer<T> equalityComparer = null)
         {
             if (items == null)
             {
@@ -55,6 +57,8 @@ namespace DynamicData
             {
                 Add(change, true);
             }
+
+            _equalityComparer = equalityComparer;
         }
 
         /// <summary>
@@ -162,6 +166,8 @@ namespace DynamicData
         ///     The total number if individual item changes
         /// </summary>
         public int TotalChanges => Adds + Removes + Replaced + Moves;
+
+        public IEqualityComparer<T> EqualityComparer => _equalityComparer;
 
         #region Enumeration
 
--- a/src/DynamicData/List/ChangeSet.cs
+++ b/src/DynamicData/List/ChangeSet.cs
@@ -17,7 +17,6 @@ namespace DynamicData
     /// <typeparam name="T">The type of the object.</typeparam>
     public class ChangeSet<T> : IChangeSet<T>
     {
-        private readonly IEqualityComparer<T> _equalityComparer;
         private int _adds;
         private int _removes;
         private int _replaced;
@@ -41,9 +40,8 @@ namespace DynamicData
         /// Initializes a new instance of the <see cref="ChangeSet{T}" /> class.
         /// </summary>
         /// <param name="items">The items.</param>
-        /// <param name="equalityComparer">An optional equality comparer.</param>
         /// <exception cref="System.ArgumentNullException">items</exception>
-        public ChangeSet([NotNull] IEnumerable<Change<T>> items, IEqualityComparer<T> equalityComparer = null)
+        public ChangeSet([NotNull] IEnumerable<Change<T>> items)
         {
             if (items == null)
             {
@@ -57,8 +55,6 @@ namespace DynamicData
             {
                 Add(change, true);
             }
-
-            _equalityComparer = equalityComparer;
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/DynamicData/List/ChangeSet.cs
```diff
@@ -55,6 +57,8 @@ public ChangeSet([NotNull] IEnumerable<Change<T>> items)
             {
                 Add(change, true);
             }
+
+            _equalityComparer = equalityComparer;
```
glennawatson: I would add a null colesing operator and default to the default equality operator for the null case

dchaib: I see! And I should probably do the same in the other constructor, just above. That would also mean it's never null and I could simplify the rest of the code.

src/DynamicData/List/IChangeSet.cs
```diff
@@ -24,5 +24,7 @@ public interface IChangeSet<TObject> : IEnumerable<Change<TObject>>, IChangeSet
         /// The total count of items changed
         /// </summary>
         int TotalChanges { get; }
+
```
RolandPheasant: I see why you have added the equality comparer to the change set.  However,  I do not see this as being the correct solution to pass the equality comparer into the clone method.  The change set exists for the purpose of transmitting changes only with a very well defined API. To add this new property to the public API for the purpose of a single function is overkill as it is irrelevant for the rest of the system.

I am sure there is a functional way of passing the comparer. I have grabbed your branch and will take a look to see if I can find an obvious alternative.

dchaib: I understand. That's exactly why I didn't go too far in this PR.

RolandPheasant:
I think I have worked it out.  In TansformMany.cs you need to change the following code:


```cs
return _source.Transform(item => new ManyContainer(_manyselector(item).ToArray()), true)
	.Select(changes => new ChangeSet<TDestination>(new DestinationEnumerator(changes, _equalityComparer), qualityComparer))
	.NotEmpty();
```

to 
```cs
return Observable.Create<IChangeSet<TDestination>>(observer =>
    {
        var result = new ChangeAwareList<TDestination>();

        return _source.Transform(item => new ManyContainer(_manyselector(item).ToArray()), true)
            .Select(changes =>
            {
                var destinationChanges = new ChangeSet<TDestination>(new DestinationEnumerator(changes, _equalityComparer));
                result.Clone(destinationChanges, _equalityComparer);
                return result.CaptureChanges();
            })

            .NotEmpty()
            .SubscribeSafe(observer);
    }
);
```

Also `ListEx.Clone` should be changed to 
```cs
void Clone<T>(this IList<T> source, IChangeSet<T> changes, IEqualityComparer<T> equalityComparer = null)
```

and inside that method change 
```cs
Clone(source,  item, changes.EqualityComparer);
```

to
```cs
Clone(source,  item, equalityComparer ?? EqualityComparer<T>.Default);
```
With these changes, your unit tests pass, which means you can remove the equality comparer from the change set

dchaib:
Awesome! I'll revert my commit and proceed with your proposal! Thanks a lot!

src/DynamicData/List/ListEx.cs
```diff
@@ -45,6 +45,23 @@ internal static bool MovedWithinRange<T>(this Change<T> source, int startIndex,
...
+        /// <typeparam name="T"></typeparam>
+        /// <param name="source">The source.</param>
+        /// <param name="changes">The changes.</param>
+        /// <param name="equalityComparer">An equality comparer to match items in the changes.</param>
+        /// <exception cref="System.ArgumentNullException">
+        /// source
+        /// or
+        /// changes
+        /// </exception>
+        public static void Clone<T>(this IList<T> source, IChangeSet<T> changes, IEqualityComparer<T> equalityComparer)
```
dchaib: I have added an overload instead of an optional parameter because of https://github.com/dchaib/DynamicData/blob/e4e0327fd38c909f293e1bd9731049fdd78a73e0/src/DynamicData/List/ObservableListEx.cs#L1246
Let me know if I should fix this differently!

src/DynamicData/List/ChangeSet.cs
dchaib: Done where it is now relevant :)

src/DynamicData/List/IChangeSet.cs
dchaib: PR updated with your suggestion

dchaib:
Should `equalityComparer` be used in more cases in the `Clone` method? For example, it might be needed for `ListChangeReason.RemoveRange`.

Should the equality comparer be passed in more cases to the `Clone` method? Maybe for `TransformMany` in `CreateWithChangeset`:
* https://github.com/dchaib/DynamicData/blob/e4e0327fd38c909f293e1bd9731049fdd78a73e0/src/DynamicData/List/Internal/TransformMany.cs#L141
* https://github.com/dchaib/DynamicData/blob/e4e0327fd38c909f293e1bd9731049fdd78a73e0/src/DynamicData/List/Internal/TransformMany.cs#L149

RolandPheasant:
Yes, good spot

glennawatson:
I would recommend incrementing the number of the minor in version.json in the root directory before merging this commit

dchaib:
@RolandPheasant do you know how I can trigger `ListChangeReason.RemoveRange` and a call to `RemoveMany`? I'm trying to add a unit test to cover this.

RolandPheasant:
RemoveMany is not produced by the transform many operator

dchaib:
Thanks, that explains why I couldn't figure out how to produce it. :)

Then, I guess I'm done as far as I'm concerned.

glennawatson:
Just bump the version.json minor version would be the last thing

RolandPheasant:
I agree with @glennawatson about the minor version. Please bump then then I will merge this PR

dchaib:
Done!

PR #276, opened 2019-09-21
BufferIf:  Adjusted API to not collide

Noggog:
If only a pause observable was provided, the method signatures were ambiguous

API improvement to avoid ambiguous calls for BufferIf

If BufferIf is called with only a pause observable provided, the compiler complains about ambiguous calls.

**What is the new behavior?**
Happy compiler

**What might this PR break?**
I don't think it will break much, as any users of those extra parameters will still see them as API options.  This should only improve users of BufferIf that only provide a pause observable.

### DIFF
--- a/src/DynamicData/List/ObservableListEx.cs
+++ b/src/DynamicData/List/ObservableListEx.cs
@@ -1713,7 +1713,7 @@ namespace DynamicData
         /// <exception cref="System.ArgumentNullException">source</exception>
         public static IObservable<IChangeSet<T>> BufferIf<T>([NotNull] this IObservable<IChangeSet<T>> source,
             [NotNull] IObservable<bool> pauseIfTrueSelector,
-            bool intialPauseState = false,
+            bool intialPauseState,
             IScheduler scheduler = null)
         {
             if (source == null)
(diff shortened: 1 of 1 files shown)


Noggog:
Heya!  Not too familiar /w doing pull requests as I haven't done too many before.  Let me know if you need anything tweaked or addressed.  8)

glennawatson:
Hey @noggog. Thanks for doing the pr. Yeah they have a inbuilt way for us to do a review and if any problems do a review and show the block of code we are talking about. You can add a new commit to your branch/fork and we will automatically see it. If the affected line of code we indicated feedback has changed GitHub will tell us. 

We could also mark for approval which indicates we are happy with the work. I often do this for dynamic data for example if I want to indicate to @rolandpheasant that I'm happy with the work. 

We can merge which adds the code to the main product.  

Final case that can happen is we reject. So this happens if we are planning a similar feature or if it just didn't fit with the vision of the product. Often these types of issues can be avoided by submitting an issue first and waiting for the maintainer to respond. 

Bit of an overview since you said you are new to the process.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1713,7 +1713,7 @@ public static IObservable<IChangeSet<T>> FlattenBufferResult<T>(this IObservable
         /// <exception cref="System.ArgumentNullException">source</exception>
         public static IObservable<IChangeSet<T>> BufferIf<T>([NotNull] this IObservable<IChangeSet<T>> source,
```
glennawatson: There doesn't seem to be a version with just the `pauseIfTrueSelector` and maybe the scheduler?

Noggog: It's just hiding unmodified on L1696

glennawatson:
Thanks, will let @RolandPheasant review as well. But makes sense to me. (@RolandPheasant  is the original author of DynamicData)

PR #277, opened 2019-09-21
Buffer if clearing fix

Noggog:
A hopeful fix for ObservableList BufferIf usage being more reliable downstream, especially when used /w delaying calls like ObserveOn.

The change makes a new list for the internal buffer when firing, rather than reusing the same list that was provided to ChangeSet.

Current behavior reuses the same internal buffer list for the outgoing ChangeSet that it is firing.  This meant that the subsequent Clear() call on the buffer list would clear the ChangeSet as well, as it was the same list.  This only became a problem if the ChangeSet consumption was delayed by an ObserveOn call or something similar.

**What is the new behavior?**
When executing an event, it passes the internal buffer to the ChangeSet to be used directly.  A new list is then created for the internal buffer variable to be filled with new data.

An alternate route could have been constructing a new list and copying over the buffer's data to be given to the ChangeSet, and then clearing the buffer.  The current setup felt like less processing with the same results.

**What might this PR break?**
ObservableList BufferIf calls in general.  Might not be a full fix for all possible threading scenarios.  Might be some oddball edge cases not yet considered.

### DIFF
--- a/src/DynamicData/List/Internal/BufferIf.cs
+++ b/src/DynamicData/List/Internal/BufferIf.cs
@@ -72,7 +72,7 @@ namespace DynamicData.List.Internal
                                                        }
 
                                                        observer.OnNext(new ChangeSet<T>(buffer));
-                                                       buffer.Clear();
+                                                       buffer = new List<Change<T>>();
 
                                                        //kill off timeout if required
                                                        timeoutSubscriber.Disposable = Disposable.Empty;
--- a/src/DynamicData/List/ObservableListEx.cs
+++ b/src/DynamicData/List/ObservableListEx.cs
@@ -1713,7 +1713,7 @@ namespace DynamicData
         /// <exception cref="System.ArgumentNullException">source</exception>
         public static IObservable<IChangeSet<T>> BufferIf<T>([NotNull] this IObservable<IChangeSet<T>> source,
             [NotNull] IObservable<bool> pauseIfTrueSelector,
-            bool intialPauseState = false,
+            bool intialPauseState,
             IScheduler scheduler = null)
         {
             if (source == null)
@@ -1742,7 +1742,7 @@ namespace DynamicData
         /// <exception cref="System.ArgumentNullException">source</exception>
         public static IObservable<IChangeSet<T>> BufferIf<T>(this IObservable<IChangeSet<T>> source,
             IObservable<bool> pauseIfTrueSelector,
-            TimeSpan? timeOut = null,
+            TimeSpan? timeOut,
             IScheduler scheduler = null)
         {
             return BufferIf(source, pauseIfTrueSelector, false, timeOut, scheduler);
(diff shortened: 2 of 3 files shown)


src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1713,7 +1713,7 @@ public static IObservable<IChangeSet<T>> FlattenBufferResult<T>(this IObservable
         /// <exception cref="System.ArgumentNullException">source</exception>
         public static IObservable<IChangeSet<T>> BufferIf<T>([NotNull] this IObservable<IChangeSet<T>> source,
             [NotNull] IObservable<bool> pauseIfTrueSelector,
-            bool intialPauseState = false,
```
glennawatson: I suspect this is the same as #276 -- a tip for avoiding inclusion code in your pull requests is to use the git branch feature. So I tend to keep "master" matching the third party software version of master, then just branch off what I'm working on from there. For this PR it just means it should probably come after #276 is merged so we don't duplicate up.

src/DynamicData/List/Internal/BufferIf.cs
```diff
@@ -72,7 +72,7 @@ public IObservable<IChangeSet<T>> Run()
                                                        }
 
                                                        observer.OnNext(new ChangeSet<T>(buffer));
-                                                       buffer.Clear();
+                                                       buffer = new List<Change<T>>();
```
glennawatson: I wonder if `ChangeSet<T>`'s constructor should be generating a new list in the `List<T>` based parameter. At the moment it just sets a reference, and does ToList() only in the non-`List<T>` parameter

src/DynamicData/List/ObservableListEx.cs
Noggog: Yeah, I'll make sure to avoid cross pollination going forward

src/DynamicData/List/Internal/BufferIf.cs
Noggog: I felt like there might be other scenarios where you know the list you're giving the ChangeSet is safe to use directly, so we would not want it to make another copy again right after for no reason.

The way it is now leaves it more up to the creator of the ChangeSet<T> to decide if/when the copy needs to happen, which I think is more appropriate.

RolandPheasant: @glennawatson `ChangeSet<T>` needs to be optimised - the list copying is not desirable. A while back, I changed `ChangeSet<TObject, TKey>` so it is inherited directly from a list.  The same needs to be applied across the board for list functions

RolandPheasant: `buffer = new List<Change<T>>();` makes sense

src/DynamicData/List/ObservableListEx.cs
RolandPheasant: I am happy for this one to go through as I do not foresee any issues

PR #278, opened 2019-09-23
List ChangeSet modified to match Cache ChangeSet patterns

Noggog:
The list-based ChangeSet class inherits from List<T> now, and performs counter calculations on access.  This mirrors the cache-based ChangeSet systems.

An internal List is passed in and used, and counters are tallied on construction of the ChangeSet.

**What is the new behavior?**
ChangeSet is itself the list, now, and counters are tallied when the user accesses their properties.

**What might this PR break?**
Anything interacting with the list-based ChangeSet

**Other information**:
I did not add any tests, but I did run the existing ones.  Let me know if we think additional unit tests should be added

### DIFF
--- a/src/DynamicData/List/ChangeSet.cs
+++ b/src/DynamicData/List/ChangeSet.cs
@@ -15,14 +15,8 @@ namespace DynamicData
     /// A set of changes which has occured since the last reported change
     /// </summary>
     /// <typeparam name="T">The type of the object.</typeparam>
-    public class ChangeSet<T> : IChangeSet<T>
+    public class ChangeSet<T> : List<Change<T>>, IChangeSet<T>
     {
-        private int _adds;
-        private int _removes;
-        private int _replaced;
-        private int _moves;
-        private int _refreshes;
-
         /// <summary>
         /// An empty change set
         /// </summary>
@@ -33,7 +27,6 @@ namespace DynamicData
         /// </summary>
         public ChangeSet()
         {
-            Items = new List<Change<T>>();
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/DynamicData/List/ChangeSet.cs
```diff
@@ -42,145 +35,79 @@ public ChangeSet()
...
-                    _replaced++;
-                    break;
-                case ListChangeReason.Remove:
-                    _removes++;
-                    break;
-                case ListChangeReason.RemoveRange:
-                    _removes = _removes + item.Range.Count;
-                    break;
-                case ListChangeReason.Refresh:
-                    _removes = _refreshes++;
```
Noggog: I did notice this potential bug here, where Refresh events were contributing to the remove counter instead of the refresh counter.  This should be implicitly fixed in this PR's refactor.

glennawatson:
Might be worth running the benchmark project and providing some benchmark timings. Do before change and after change. Since this is very PC specific make sure you run both benchmark runs on the same PC. Just with the properties getting tallied and not cached anymore just concern it might add performance hits.

Noggog:
Here's the results, glenn
Results
There doesn't seem to be too much of a difference.  I see small + and - both ways here and there.

I'm somewhat wary of them for two reasons:
1. Lots of warnings that the observed iteration time was too small to be accurate, even for the larger data sets.
2. While I would expect a faster time for these tests like AddRange/Insert, it's not flexing the other side where we're now paying the cost.  The cost added is during typical usage where queries/transformations make use of the Adds/Removes properties that are now calculations.   There'd need to be benchmarks of "typical" usage in order to measure that, but even that would be not too informative as everyone's usage is different.

Noggog:
Also, this PR would make the just approved #277 need some maintenance, as its internal buffer variable could likely just be swapped to a ChangeSet directly, rather than being a List that gets passed to a ChangeSet ctor, that is copied over internally causing unnecessary work.  There's probably a few other places that do the same that we could shark around and optimize.  

Will do that after this PR, if it's approved.

glennawatson:
Benchmarks can show if something dramatic has changed. The delta I see in your example are very minimal that PC usage at the time can explain the change.

RolandPheasant:
There will definitely be an improvement with this change as a `ToList()` (which can be very bad) has been removed. There are many reasons why there is not much difference showing but overall so many change sets are created that this must be a good thing.

Also one plan I have for a future major version change is to get rid of the IChangeSet interface altogether as the interface automatically means allocations when enumerating, whereas when you enumerate a list there is no allocation. This change is a small step to that very large improvement.

@Noggog further to changing the buffer (in the last PR) directly to a change set, I suspect there are quite a few places throughout the system where similar improvements can be made.

PR #279, opened 2019-09-23
IObservableList implements IReadOnlyList.  ISourceList implements IList

Noggog:
IObservableList implements IReadOnlyList.  ISourceList implements IList.

I find that I write a lot of convenience code via extension methods, and I try to hook them onto the most generic objects that I can.  These would be functions similar to the ones found in DynamicData itself like `ListEx.AddRange(this IList<T> source, IEnumerable<T> items)`;

Because DynamicData's list functionality didn't inherit from the List interfaces meant its collections didn't hook onto those extension methods.  I had to duplicate the code for them to work with DynamicData, which is verbose and hindered refactorability as both versions needed to be maintained with the same updates.

I didn't see anything glaring as to why DynamicData's lists couldn't implement the interfaces, so I gave it a stab that seems to have worked out alright for my fork, so I thought I'd migrate it upwards for everyone else.

SourceList and its interfaces implement list patterns, without actually implementing IList<T>/IReadOnlyList<T>

**What is the new behavior?**
SourceList and its interfaces implement IList<T>/IReadOnlyList<T>

**What might this PR break?**
This could break wiring that currently works around the fact that SourceList doesn't implement IList.  I had to do a bit of maintenance myself on ObservableListEx, for example, adding another extension method that took ISourceList directly to clear the ambiguous call between the `IObservable<IChangeSet<T>>` and ` ICollection<IObservable<IChangeSet<T>>>` versions.

I believe I've handled all the cases in DynamicData itself, but users of the library might have made their own workarounds for the lack of IList that might need to be adjusted or dialed back.

**Other information**:
I ran the unit tests, which passed.  I couldn't think of any obvious unit tests to add, but let me know if anyone thinks of some we could write to flex any of the new behavior.

### DIFF
--- a/src/DynamicData/Cache/ObservableCacheEx.cs
+++ b/src/DynamicData/Cache/ObservableCacheEx.cs
@@ -2195,6 +2195,24 @@ namespace DynamicData
         }
 
         /// <summary>
+        /// Dynamically apply a logical And operator between the items in the outer source list.
+        /// Items which are in all of the sources are included in the result
+        /// </summary>
+        /// <typeparam name="TObject">The type of the object.</typeparam>
+        /// <typeparam name="TKey">The type of the key.</typeparam>
+        /// <param name="sources">The source.</param>
+        /// <returns></returns>
+        public static IObservable<IChangeSet<TObject, TKey>> And<TObject, TKey>(this ISourceList<IObservable<IChangeSet<TObject, TKey>>> sources)
+        {
+            if (sources == null)
+            {
+                throw new ArgumentNullException(nameof(sources));
+            }
+
+            return And((IObservableList<IObservable<IChangeSet<TObject, TKey>>>)sources);
+        }
+
+        /// <summary>
         /// Dynamically apply a logical And operator between the items in the outer observable list.
         /// Items which are in all of the sources are included in the result
         /// </summary>
@@ -2297,6 +2315,24 @@ namespace DynamicData
             }
 
             return sources.Combine(CombineOperator.Or);
+        }
+
+        /// <summary>
+        /// Dynamically apply a logical Or operator between the items in the outer source list.
+        /// Items which are in any of the sources are included in the result
+        /// </summary>
+        /// <typeparam name="TObject">The type of the object.</typeparam>
+        /// <typeparam name="TKey">The type of the key.</typeparam>
+        /// <param name="sources">The source.</param>
+        /// <returns></returns>
+        public static IObservable<IChangeSet<TObject, TKey>> Or<TObject, TKey>(this ISourceList<IObservable<IChangeSet<TObject, TKey>>> sources)
+        {
+            if (sources == null)
+            {
+                throw new ArgumentNullException(nameof(sources));
+            }
+
+            return Or((IObservableList<IObservable<IChangeSet<TObject, TKey>>>)sources);
         }
 
         //public static IObservable<IChangeSet<TDestination, TDestinationKey>> Or<TDestination, TDestinationKey, TSource, TSourceKey>(this IObservable<IObservableCache<TSource, TSourceKey>> source,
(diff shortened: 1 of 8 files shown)


src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -2194,6 +2194,24 @@ public static IObservable<IReadOnlyCollection<T>> StartWithEmpty<T>(this IObserv
             return sources.Combine(CombineOperator.And);
         }
 
+        /// <summary>
```
Noggog: These functions are needed, as the two functions above them collide now that ISourceList implements them both.  This function is a clarifying function that the compiler can use to remove the ambiguity.  It internally calls the IObservableList version instead of the ICollection version.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -2120,6 +2120,19 @@ public static IObservable<IChangeSet<T>> Or<T>([NotNull] this ICollection<IObser
             return sources.Combine(CombineOperator.Or);
         }
 
+        /// <summary>
```
Noggog: These functions are needed, as the two functions above them collide now that ISourceList implements them both.  This function is a clarifying function that the compiler can use to remove the ambiguity.  It internally calls the IObservableList version instead of the ICollection version.

src/DynamicData/List/IObservableList.cs
```diff
@@ -37,10 +37,5 @@ public interface IObservableList<T> : IDisposable
         /// Items enumerable
         /// </summary>
         IEnumerable<T> Items { get; }
-
```
Noggog: This property was removed, as IReadOnlyList provides it now

src/DynamicData/List/IObservableList.cs
```diff
@@ -37,10 +37,5 @@ public interface IObservableList<T> : IDisposable
...
-        /// <summary>
-        /// Gets the count.
-        /// </summary>
-        int Count { get; }
```
Noggog: This property was removed, as IReadOnlyList provides it now

src/DynamicData/List/ISourceList.cs
```diff
@@ -12,12 +13,36 @@ namespace DynamicData
     /// as well as data access methods
     /// </summary>
     /// <typeparam name="T"></typeparam>
-    public interface ISourceList<T> : IObservableList<T>
+    public interface ISourceList<T> : IObservableList<T>, IList<T>
```
RolandPheasant: I am not sure whether it is desirable to fully implement `IList<T>`. I have always resisted doing so since I actively want to discourage people thinking that it should be treated in the same way as a normal list.

The reason  being that most list operations on source list are expensive in terms of inner locks and array copying. What I can be sure of is if consumers see Ilist then they will use it as a normal IList and this will surely lead to complaints about performance.

But then again maybe this is a needless worry of mine.

RolandPheasant: I am interested in other opinions out there

Noggog: I'll just throw in my two cents, but I'm also interested in hearing some other's thoughts.

I agree with your sentiment that it's good to help guide people from using expensive objects in ways that feel cheap that might deceive them.  I'm totally on board with that up until it starts to hinder other important patterns and goals of users.  For me, the one that's being broken here is encouraging hefty amounts of duplicate code, which is a bigger architectural offense/concern for me.  Avoiding duplicate code heavily outweighs potentially helping newcomers avoid usage traps their first time around.

Not to mention, I think the familiar IList traps are all still there via the SourceListEditConvenienceEx class, even if they don't "officially" come from an IList implementation.  I know that I already personally have fallen into it not a few days ago, lol.  I was populating a SourceList, and using Add() one by one like an idiot, which was astronomically slow.  This is what made me create #277, as I opted to try out the BufferIf functionality rather than using AddRange.

RolandPheasant: Having sat on this one for a day, and looking at the code again I am beginning to think that there may be no harm in implementing IList, after all as you point out most of the methods were available on the convenience methods.

However I think it wise for me to sit on this for a few days more as once an API it is available I have to support it forever.

src/DynamicData/List/ISourceList.cs
```diff
@@ -12,12 +13,36 @@ namespace DynamicData
...
+        //
+        // Returns:
+        //     The element at the specified index.
+        //
+        // Exceptions:
+        //   T:System.ArgumentOutOfRangeException:
+        //     index is not a valid index in the ISourceList`1.
+        //
+        //   T:System.NotSupportedException:
+        //     The property is set and the ISourceList`1 is read-only.
```
worldbeater: Looks like something went wrong with the XML documentation, probably worth adjusting it a bit.

Noggog: I copied the comment from IList's comment for its indexer property directly, and just swapped `IList` for `ISourceList`.  The odd looking `1 is in the original comments, too, as it's a result of how generics are toString()'ed by default.   I do agree it's an odd way to display generics, but it is standard, I'd say.  I don't mind swapping it out for something we all agree we like better, though.

glennawatson: I think @worldbeater is referring it's not in xmldoc format with the `///` and XML tags.

glennawatson: In the current form it won't be picked up by intellisense.

Noggog: Ah, my bad.  I'm following now.
Swapping it to the xml format.  Thanks!

glennawatson:
Spoken to @RolandPheasant we are inclined to decline this PR.

The rationale is that it could have performance implications to the user if they start using it as a IReadOnlyList<T>.

The main concept of the SourceList/Cache is they are meant to be used binding to visible containers.

PR #280, opened 2019-09-25
Reduced some extra ChangeSet instantiations + content copies

Noggog:
Cuts some extra instantiations and list copies.  After ChangeSet<T> was updated to be itself a List (#278), there was some room for potential optimizations.  I did a find references on code that utilized the ChangeSet(IEnumerable<Change<T>>) constructor, and looked very briefly around those areas for potential improvements.  These are the half dozen I spotted that I thought were quick tweaks.

Some code snippets created extra middlemen lists or changesets that were then passed to the final returning changeset.

**What is the new behavior?**
Code tries to avoid extra internal lists or copies by using ChangeSet's list functionality as directly as possible.

**What might this PR break?**
Could always be a decent amount of regression bugs if there are some edge cases not considered in the refactors presented.

**Other information**:
I ran the unit tests, which passed.  I did not write any new tests, as no new functionality was added.  Only existing functionality was slightly optimized.

### DIFF
--- a/src/DynamicData/List/ChangeAwareList.cs
+++ b/src/DynamicData/List/ChangeAwareList.cs
@@ -21,7 +21,7 @@ namespace DynamicData
     {
         private readonly object _lockObject = new object();
         private readonly List<T> _innerList;
-        private List<Change<T>> _changes = new List<Change<T>>();
+        private ChangeSet<T> _changes = new ChangeSet<T>();
 
         /// <summary>
         /// Create a change aware list with the specified capacity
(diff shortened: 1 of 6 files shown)


src/DynamicData/List/ChangeAwareList.cs
```diff
@@ -84,7 +84,7 @@ public IChangeSet<T> CaptureChanges()
                     return ChangeSet<T>.Empty;
                 }
 
-                returnValue = new ChangeSet<T>(_changes);
+                returnValue = _changes;
```
Noggog: We can now fill a changeset up directly, and pass it as the return value when we want a capture.  We're always doing a clear afterwards, so downstream users will not have reference to our internal changeset after this call.

Noggog:
Something got weird.  Some files marked as full removals and readditions.  Going to delete this PR and start a new one

PR #281, opened 2019-09-25
Reduced some extra ChangeSet instantiations + content copies

Noggog:
Cuts some extra instantiations and list copies. After ChangeSet was updated to be itself a List (#278), there was some room for potential optimizations. I did a find references on code that utilized the ChangeSet(IEnumerable<Change>) constructor, and looked very briefly around those areas for potential improvements. These are the half dozen I spotted that I thought were quick tweaks.

Edit:  I also moved a call to be inside a lock where it wasn't before.  See related comment

Some code snippets created extra middlemen lists or changesets that were then passed to the final returning changeset.

**What is the new behavior?**
Code tries to avoid extra internal lists or copies by using ChangeSet's list functionality as directly as possible.

In the ChangeAwareList, during captures, the change to the locking should make it safer by ensuring the clear occurs after the capture before any other changes from another thread might occur.

**What might this PR break?**
Could always be a decent amount of regression bugs if there are some edge cases not considered in the refactors presented.

**Other information**:
I ran the unit tests, which passed. I did not write any new tests, as no new functionality was added. Only existing functionality was slightly optimized.

### DIFF
--- a/src/DynamicData/List/ChangeAwareList.cs
+++ b/src/DynamicData/List/ChangeAwareList.cs
@@ -21,7 +21,7 @@ namespace DynamicData
     {
         private readonly object _lockObject = new object();
         private readonly List<T> _innerList;
-        private List<Change<T>> _changes = new List<Change<T>>();
+        private ChangeSet<T> _changes = new ChangeSet<T>();
 
         /// <summary>
         /// Create a change aware list with the specified capacity
@@ -67,7 +67,7 @@ namespace DynamicData
 
             if (copyChanges)
             {
-                _changes = new List<Change<T>>(list._changes);
+                _changes = new ChangeSet<T>(list._changes);
             }
         }
 
@@ -84,7 +84,7 @@ namespace DynamicData
                     return ChangeSet<T>.Empty;
                 }
 
-                returnValue = new ChangeSet<T>(_changes);
+                returnValue = _changes;
 
                 //we can infer this is a Clear
                 if (_innerList.Count == 0 && returnValue.Removes == returnValue.TotalChanges && returnValue.TotalChanges > 1)
@@ -106,7 +106,7 @@ namespace DynamicData
         {
             lock (_lockObject)
             {
-                _changes = new List<Change<T>>();
+                _changes = new ChangeSet<T>();
             }
         }
 
--- a/src/DynamicData/List/ChangeAwareList.cs
+++ b/src/DynamicData/List/ChangeAwareList.cs
@@ -92,9 +92,9 @@ namespace DynamicData
                     var removed = returnValue.Unified().Select(u => u.Current);
                     returnValue = new ChangeSet<T> { new Change<T>(ListChangeReason.Clear, removed) };
                 }
-            }
 
-            ClearChanges();
+                ClearChanges();
+            }
 
             return returnValue;
         }
--- a/src/DynamicData/List/Internal/TransformMany.cs
+++ b/src/DynamicData/List/Internal/TransformMany.cs
@@ -101,7 +101,7 @@ namespace DynamicData.List.Internal
                 return _source.Transform(item => new ManyContainer(_manyselector(item).ToArray()), true)
                     .Select(changes =>
                     {
-                        var destinationChanges = new ChangeSet<TDestination>(new DestinationEnumerator(changes, _equalityComparer));
+                        var destinationChanges = new DestinationEnumerator(changes, _equalityComparer);
                         result.Clone(destinationChanges, _equalityComparer);
                         return result.CaptureChanges();
                     })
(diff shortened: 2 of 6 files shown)


src/DynamicData/List/ListEx.cs
```diff
@@ -62,7 +62,24 @@ public static void Clone<T>(this IList<T> source, IChangeSet<T> changes)
...
+        /// <typeparam name="T"></typeparam> 
+        /// <param name="source">The source.</param> 
+        /// <param name="changes">The changes.</param> 
+        /// <param name="equalityComparer">An equality comparer to match items in the changes.</param> 
+        /// <exception cref="System.ArgumentNullException"> 
+        /// source 
+        /// or 
+        /// changes 
+        /// </exception> 
+        public static void Clone<T>(this IList<T> source, IEnumerable<Change<T>> changes, IEqualityComparer<T> equalityComparer)
```
Noggog: Added to allow changes like the one found on TransformMany to clone without needing to create an intermediary ChangeSet

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -408,8 +408,7 @@ public static IObservable<IChangeSet<object>> CastToObject<T>(this IObservable<I
         {
             return source.Select(changes =>
             {
-                var items = changes.Transform(t => (object)t);
-                return new ChangeSet<object>(items);
+                return changes.Transform(t => (object)t);
```
Noggog: Transform creates a new ChangeSet internally, so no need to wrap it again

src/DynamicData/List/SourceList.cs
```diff
@@ -160,10 +160,13 @@ public IObservable<IChangeSet<T>> Connect(Func<T, bool> predicate = null)
             {
                 lock (_locker)
                 {
-                    var initial = new ChangeSet<T>(new[] {new Change<T>(ListChangeReason.AddRange, _readerWriter.Items)});
-                    if (initial.TotalChanges > 0)
+                    if (_readerWriter.Items.Length > 0)
```
Noggog: Can short circuit if the current items are empty, rather than making a changeset and filling it first

src/DynamicData/List/Internal/TransformMany.cs
```diff
@@ -101,7 +101,7 @@ public IObservable<IChangeSet<TDestination>> Run()
                 return _source.Transform(item => new ManyContainer(_manyselector(item).ToArray()), true)
                     .Select(changes =>
                     {
-                        var destinationChanges = new ChangeSet<TDestination>(new DestinationEnumerator(changes, _equalityComparer));
+                        var destinationChanges = new DestinationEnumerator(changes, _equalityComparer);
```
Noggog: Passed directly as an enumerable to Clone().  No need to make a changeset to hold the results first.

src/DynamicData/List/ChangeAwareList.cs
```diff
@@ -92,9 +92,9 @@ public IChangeSet<T> CaptureChanges()
                     var removed = returnValue.Unified().Select(u => u.Current);
                     returnValue = new ChangeSet<T> { new Change<T>(ListChangeReason.Clear, removed) };
                 }
-            }
 
-            ClearChanges();
+                ClearChanges();
```
Noggog: I think this should be inside the lock.  There was the slight possibility that a change could've happened after firing the changset, but before the lock was acquired for the clear.   Let me know if anyone disagrees with this change, as lock side effects can always be weird 8)

Noggog:
Hmm, was surprised to see the DynamicData.Tests.Cache.SizeLimitFixture.AddMoreThanLimitInBatched unit test failing.  I'm uncertain it's related to the changes I made, as I only touched List systems, not cache ones.  To try to find out for sure, I tried putting breakpoints in all the code I touched, and ran just that one unit test.  I did not hit any of my breakpoints.  I did run it twice, and got a pass one of the times, and the failure mentioned the other time.

Here's a pic of the exception happening, with some local variables if happens to be helpful.
UnitTestError

It looks like just a non-deterministic test that might need some tweaks?

glennawatson:
hmmm.. seems it's likely one of the class variables that might be an issue.

RolandPheasant:
@Noggog did you ever get to the bottom of failing unit test?

Noggog:
@RolandPheasant I did not look into it much further.  It felt to me like an unrelated non-deterministic unit test that coincidentally failed, so I didn't know what to do to look into it more.
1)  I put breakpoints in all of my modified code.  The failing test hit none of the breakpoints, and so I don't think my modified code touched that unit test's logic path.  (Perhaps I did this poorly?)
2)  When I re-ran the tests myself, I saw both failures on some runs, and successes on others.  Feels non-deterministic at that point.

glennawatson:
@dependabot rebase

glennawatson:
I tested these changes for a couple hours today. Seems to work fine. Also spoke to @RolandPheasant so we've decided to merge it in.

Thanks for the contribution @Noggog and sorry it took so long.

PR #327, opened 2020-01-28
[impl] remove items not in enumerable anymore

CaseNewmark:
It changes the behavior of the `toObservableChangeSet()` method when passing an `IEnumerable`.

Items that have been in the `IEnumerable` once, will stay in the cache forever (or until expiration).

**What is the new behavior?**
`IEnumerable` entries are compared to the cache. If there are keys in the cache not in the `IEnumerable` anymore, they get removed.

**What might this PR break?**
This changes the behavior of this use case for the `toObservableChangeSet` in general. Not sure if it's a good idea to do it like this.

- [X ] Tests for the changes have been added (for bug fixes / features)

**Other information**:
To be discussed also in https://github.com/reactiveui/DynamicData/issues/326

### DIFF
--- a/src/DynamicData.Tests/Cache/EnumerableObservableToObservableChangeSetFixture.cs
+++ b/src/DynamicData.Tests/Cache/EnumerableObservableToObservableChangeSetFixture.cs
@@ -33,6 +33,41 @@ namespace DynamicData.Tests.Cache
         }
 
         [Fact]
+        public void OnNextProducesAnAddAndRemoveChangeForEnumerableSource()
+        {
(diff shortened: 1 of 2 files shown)


CaseNewmark:
First attempt breaks one test where only a single item is introduced into the source. Thought that would be handled already, but the constructor of ToObservableChangeSet makes everything a list!?
Handling this now with a flag :(

RolandPheasant:
I have reverted this code.

See  #604 for description + the PR #603 for how I changed the unit test to use a different methodology to achieve the desired result

PR #368, opened 2020-06-08
Bump fluent assertions to 5.10.3. 

RolandPheasant:
This is a manual upgrade for fluent assertions.

I have to do a find and replace for collections asserts as the old system of `AllShouldBeEquivalentTo` was removed in v5


RolandPheasant:
this supercedes #339

PR #369, opened 2020-06-11
Bind to observable list

ryanholden8:
This is primarily viewed as a bug fix since it was not expected that the existing code did not already behave this way.

There is no way to bind out to a read-only derived data bucket and have that bucket continue to emit refresh notifications from the upstream source. 

**What is the new behavior?**
Adds a new API for `IChangeSet<T>` & `IChangeSet<T, K>`: `.BindToObservableList(out observableList)`. Same idea as `.Bind(out readOnlyCollection)`.

Provides a way to bind out to a read-only derived data bucket. The new part is that this read-only derived data bucket emits refreshes from the source rather than ignore them in the case of `ReadOnlyObservableCollection`

**What might this PR break?**
This mostly adds new APIs to bind to an `IObservableList`. Changes have been made to `RemoveKeyEnumerator<TObject, TKey>` to support Refresh and Move changes. Any code that uses `RemoveKeyEnumerator<TObject, TKey>` has the potential to be effected. However, the changes should only be improvements rather than changing an existing expected result.

- [x] Docs have been added / updated (for bug fixes / features)

**Other information**:

### DIFF
--- a/src/DynamicData.Tests/Binding/BindingListToChangeSetFixture.cs
+++ b/src/DynamicData.Tests/Binding/BindingListToChangeSetFixture.cs
@@ -1,8 +1,8 @@
-﻿using DynamicData.Binding;
-using FluentAssertions;
-using System;
+﻿using System;
 using System.ComponentModel;
 using System.Linq;
+using DynamicData.Binding;
+using FluentAssertions;
 using Xunit;
 
 namespace DynamicData.Tests.Binding
@@ -88,6 +88,45 @@ namespace DynamicData.Tests.Binding
             _collection.Add(1);
 
             _results.Messages.Count.Should().Be(1);
+        }
+
+        [Fact]
+        public void RefreshCausesReplace()
+        {
+            // Arrange
+            var sourceCache = new SourceCache<Item, Guid>(item => item.Id);
+
+            var item1 = new Item(name: "Old Name");
+
+            sourceCache.AddOrUpdate(item1);
+
+            var collection = new TestBindingList<Item>();
+
+            var sourceCacheResults = sourceCache
+                .Connect()
+                .AutoRefresh(item => item.Name)
+                .Bind(collection)
+                .AsAggregator();
+
+            var collectionResults = collection.ToObservableChangeSet().AsAggregator();
+
+            item1.Name = "New Name";
+
+            // Source cache received add and refresh
+            sourceCacheResults.Messages.Count.Should().Be(2);
+            sourceCacheResults.Messages.First().Adds.Should().Be(1);
+            sourceCacheResults.Messages.Last().Refreshes.Should().Be(1);
+
+            // List receives add and replace instead of refresh
+            collectionResults.Messages.Count.Should().Be(2);
+            collectionResults.Messages.First().Adds.Should().Be(1);
+            collectionResults.Messages.First().Refreshes.Should().Be(0);
+            collectionResults.Messages.Last().Replaced.Should().Be(1);
+            collectionResults.Messages.Last().Refreshes.Should().Be(0);
+
+            sourceCache.Dispose();
+            sourceCacheResults.Dispose();
+            collectionResults.Dispose();
         }
 
         private class TestBindingList<T> : BindingList<T>
--- a/src/DynamicData.Tests/Binding/BindingListToChangeSetFixture.cs
+++ b/src/DynamicData.Tests/Binding/BindingListToChangeSetFixture.cs
@@ -42,7 +42,7 @@ namespace DynamicData.Tests.Binding
 
             _results.Data.Count.Should().Be(9);
             _results.Data.Items.Contains(3).Should().BeFalse();
-            _results.Data.Items.ShouldAllBeEquivalentTo(_collection);
+            _results.Data.Items.Should().BeEquivalentTo(_collection);
         }
 
         [Fact]
@@ -60,7 +60,7 @@ namespace DynamicData.Tests.Binding
             _collection.AddRange(Enumerable.Range(1, 10));
             _collection[8] = 20;
 
-            _results.Data.Items.ShouldBeEquivalentTo(new[] { 1, 2, 3, 4, 5, 6, 7, 8, 20, 10 });
+            _results.Data.Items.Should().BeEquivalentTo(new[] { 1, 2, 3, 4, 5, 6, 7, 8, 20, 10 });
         }
 
         [Fact]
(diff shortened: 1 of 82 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/reactiveui/DynamicData?pullRequest=369) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/reactiveui/DynamicData?pullRequest=369) before we can accept your contribution.<br/><br/>:x: ryanholden8 [sign now](https://cla.dotnetfoundation.org/reactiveui/DynamicData?pullRequest=369)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/reactiveui/DynamicData?pullRequest=369) it.</sub>

RolandPheasant:
Looks great at a glance. I will look in more detail tomorrow.

src/DynamicData/Binding/IObservableListEx.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// </summary>
+        /// <typeparam name="TObject">The type of the object.</typeparam>
+        /// <typeparam name="TKey">The type of the key.</typeparam>
+        /// <param name="changeSetWithKey">The source change set</param>
+        /// <param name="list">
+        /// An optional list, if provided it allows the refresh from a key based cache to find the index for the resulting list based refresh.
+        /// If not provided a refresh will dropdown to a replace which may ultimately result in a remove+add change downstream.
+        /// </param>
+        /// <returns>The downcasted <see cref="IChangeSet{TObject}"/></returns>
+        private static IChangeSet<TObject> Convert<TObject, TKey>(this IChangeSet<TObject, TKey> changeSetWithKey, IExtendedList<TObject> list = null)
```
RolandPheasant, line 127: Have you considered using, or adapting the `RemoveKey` operator.

This serves a similar purpose to what you are doing here.  Internally remove key uses `RemoveKeyEnumerator<TObject, TKey>` which is close to what you want. The only obvious difference is that an update is treated as an add / remove rather than a replace. Also we would need to ensure it handles the refresh.

src/DynamicData/Binding/IObservableListEx.cs
```diff
@@ -0,0 +1,197 @@
...
+                case ChangeReason.Remove:
+                    // If not sorted
+                    if (change.CurrentIndex == -1)
+                    {
+                        return new Change<TObject>(ListChangeReason.Remove, change.Current);
+                    }
+                    return new Change<TObject>(ListChangeReason.Remove, change.Current, index: change.CurrentIndex);
+
+                case ChangeReason.Update:
+                    // If not sorted
```
RolandPheasant, line 184: Semantically, I always intended a replace to be the list equivalent of `list[index]=value` rather than the direct equivalent `dict[key]=value`. I am not sure whether treating these the same for all down stream operators will produce consistent results.

src/DynamicData/Binding/IObservableListEx.cs
```diff
@@ -0,0 +1,197 @@
...
+
+        /// <summary>
+        /// Binds the results to the specified <see cref="IObservableList{T}"/>. Unlike
+        /// binding to a <see cref="ReadOnlyObservableCollection{T}"/> which loses the
+        /// ability to refresh items, binding to an <see cref="IObservableList{T}"/>.
+        /// allows for refresh changes to be preserved and keeps the list read-only.
+        /// </summary>
+        /// <typeparam name="TObject">The type of the object.</typeparam>
+        /// <typeparam name="TKey">The type of the key.</typeparam>
+        /// <param name="source">The source.</param>
```
RolandPheasant, line 88: Incorrect param name "destination" on the comments. Similar else where in this class

src/DynamicData.Tests/Binding/BindingListToChangeSetFixture.cs
```diff
@@ -1,8 +1,8 @@
-﻿using DynamicData.Binding;
-using FluentAssertions;
-using System;
+﻿using System;
```
RolandPheasant, line 1: Excellent and encouraging to see loads of unit tests

ryanholden8:
Hey guys, thank you for the quick responses! I should be able to look these change requests early next week.

src/DynamicData/Binding/IObservableListEx.cs
ryanholden8: Thanks for this tip, we were trying to find something like this in the codebase but just couldn't seem to find it.

We adapted to use `RemoveKeyEnumerator<TObject, TKey>`. Our unit tests for refresh started to fail. Additionally, we wrote a unit test for move, which also started to fail. So we improved `RemoveKeyEnumerator<TObject, TKey>` to handle these two change reasons. We also improved it to pass through the index since this would be more performant when cloning out to the list.

src/DynamicData/Binding/IObservableListEx.cs
ryanholden8: I see - when adapting `RemoveKeyEnumerator<TObject, TKey>` we left it as a remove/add. Let us know if there is any adjustments we still need to make. Thanks!

src/DynamicData/Binding/IObservableListEx.cs
ryanholden8: Good catch, this has been fixed.

src/DynamicData.Tests/Binding/BindingListToChangeSetFixture.cs
ryanholden8: Gotta love those unit tests! :)

glennawatson:
I will run through testing this branch this week and see if I run into any problems. If not and haven't heard any disagreement from @RolandPheasant  will merge it in. 

Want to run it through it's paces just to make sure it works before merging.

src/DynamicData/Cache/Internal/RemoveKeyEnumerator.cs
```diff
@@ -32,14 +40,61 @@ public IEnumerator<Change<TObject>> GetEnumerator()
                 switch (change.Reason)
                 {
                     case ChangeReason.Add:
-                        yield return new Change<TObject>(ListChangeReason.Add, change.Current);
+                        // If not sorted
```
RolandPheasant, line 43: There is no need to check for the current index being -1.

```cs
 yield return new Change<TObject>(ListChangeReason.Add, change.Current, change.CurrentIndex);
```

does exactly the same as 

```cs
                        if (change.CurrentIndex == -1)
                        {
                            yield return new Change<TObject>(ListChangeReason.Add, change.Current);
                        }
                        else
                        {
                            yield return new Change<TObject>(ListChangeReason.Add, change.Current, index: change.CurrentIndex);
                        }
```

PR #373, opened 2020-06-21
SourceCache optional key EqualityComparer parameter

Noggog:
Allows the user to specify if they want a SourceCache to utilize a special key comparer internally.  

Example usage:
```
var cache = new SourceCache<MyObject, string>(
    keySelector: o => o.MyString,
    keyEqualityComparer: StringComparer.OrdinalIgnoreCase);
cache.AddOrUpdate(new MyObject()
{
    MyString = "lowercase"
});
Assert.True(cache.Lookup("LOWERCASE").HasValue);
```

No way to specify custom comparers in a SourceCache.

**What is the new behavior?**
Optional parameter on SourceCache allows for any comparer to be specified for the keys.

**What might this PR break?**
Internally, within the `ReaderWriter` class that actually uses the comparer, there might be some mismatches in the special logic within the preview functionality.  I added specific tests for this code section to help ensure correctness.

**Other notes**
The custom comparer supplied to a SourceCache in no way trickles down to any .Connect() observable chains.  But it's not certain that would be desirable anyway.

### DIFF
--- a/src/DynamicData/Cache/Internal/ReaderWriter.cs
+++ b/src/DynamicData/Cache/Internal/ReaderWriter.cs
@@ -11,12 +11,19 @@ namespace DynamicData.Cache.Internal
     internal sealed class ReaderWriter<TObject, TKey>
     {
         private readonly Func<TObject, TKey> _keySelector;
-        private Dictionary<TKey,TObject> _data = new Dictionary<TKey, TObject>(); //could do with priming this on first time load
+        private Dictionary<TKey, TObject> _data;
         private CacheUpdater<TObject, TKey> _activeUpdater;
+        private readonly IEqualityComparer<TKey> _keyEqualityComparer;
 
         private readonly object _locker = new object();
 
-        public ReaderWriter(Func<TObject, TKey> keySelector = null) => _keySelector = keySelector;
+        public ReaderWriter(Func<TObject, TKey> keySelector = null, IEqualityComparer<TKey> keyEqualityComparer = null)
+        {
+            _keySelector = keySelector;
+            _keyEqualityComparer = keyEqualityComparer ?? EqualityComparer<TKey>.Default;
+            //could do with priming this on first time load
+            _data = new Dictionary<TKey, TObject>(_keyEqualityComparer);
+        }
 
         #region Writers
 
@@ -56,7 +63,7 @@ namespace DynamicData.Cache.Internal
             {
                 if (previewHandler != null)
                 {
-                    var copy = new Dictionary<TKey, TObject>(_data);
+                    var copy = new Dictionary<TKey, TObject>(_data, _keyEqualityComparer);
                     var changeAwareCache = new ChangeAwareCache<TObject, TKey>(_data);
 
                     _activeUpdater = new CacheUpdater<TObject, TKey>(changeAwareCache, _keySelector);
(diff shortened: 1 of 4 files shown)


src/DynamicData/Cache/Internal/ReaderWriter.cs
```diff
@@ -11,12 +11,19 @@ namespace DynamicData.Cache.Internal
...
+        private readonly IEqualityComparer<TKey> _keyEqualityComparer;
 
         private readonly object _locker = new object();
 
-        public ReaderWriter(Func<TObject, TKey> keySelector = null) => _keySelector = keySelector;
+        public ReaderWriter(Func<TObject, TKey> keySelector = null, IEqualityComparer<TKey> keyEqualityComparer = null)
+        {
+            _keySelector = keySelector;
+            _keyEqualityComparer = keyEqualityComparer ?? EqualityComparer<TKey>.Default;
+            //could do with priming this on first time load
```
glennawatson, line 24: Think it wants // and then a space.

Noggog:
Ah, my bad.  Pushed /w unsupported C# language features.  One sec

glennawatson:
Had a quick play around, seems to work nicely.

src/DynamicData/Cache/SourceCache.cs
```diff
@@ -23,15 +23,16 @@ public class SourceCache<TObject, TKey> : ISourceCache<TObject, TKey>
         /// Initializes a new instance of the <see cref="SourceCache{TObject, TKey}"/> class.
         /// </summary>
         /// <param name="keySelector">The key selector.</param>
+        /// <param name="keyEqualityComparer">The comparer to use to determine if two keys are equivalent</param>
```
glennawatson, line 26: Just one request though, can you add a full stop, and also just document what comparer is used if null.

RolandPheasant:
Sorry for being a bit late to this party.

Whilst this PR looks good, the issue is that many downstream operators also use a dictionary to cache state. The issue is that whilst the source cache will respect the comparer, subsequent operators know nothing about the equality comparer.  For this solution to work we would have to transmit the comparer as well as the changes. This would inevitably mean extending the change set to hold the comparer and to trawl the code base to ensure operators internally use the comparer. This would be a very large and risky PR, and heavy weight solution which is why I have never account for this.

Case insensitive string keys are the only use case which I have every found which requires a comparer, in which case the solution I came up with was to use a custom struct - for an example, see [CaseInsensitiveString.cs](https://github.com/RolandPheasant/TailBlazer/blob/master/Source/TailBlazer.Domain/CaseInsensitiveString.cs)

An example usage is 

```cs
var cache = new SourceCache<MyObject, CaseInsensitiveString>(o => (CaseInsensitiveString)o.MyString);
```

glennawatson:
I've used it for custom inclusion/exclusion rules in the past this type of operator. 

The struct idea at least avoids most allocs but feels a bit clunky so I might do a search through the code base and identify the other operators.

Noggog:
> the issue is that many downstream operators also use a dictionary to cache state
Yeah, this was the one edge case I was worried might throw a wrench into things.  I was hoping bringing the topic up might spur some discussion towards a solution.. but yeah there's no clean way to communicate the comparer to use downstream.  

Now thinking on it more, even this simple case would fail:
```
SourceCache<MyObject, string> source = new SourceCache<MyObject, string>(StringComparer.OrdinalIgnoreCase);

IObservableCache<MyOtherObject, string> derivativeCache = source.Connect()
    .Transform(i => new MyOtherObject())
    .AsObservableCache();
```

A query into the two caches would then disagree on the presence of certain strings.

-------------

I do agree the CaseInsensitiveString seems to be the cleanest solution, unfortunately.    Unless someone has a clever idea, I think this PR pitch is probably non viable.

glennawatson:
Yeah let's close this one then and consider it in the vNext of DD.

PR #377, opened 2020-06-26
Thread safety in ObservableCache and SourceList. Fixes #376.

### LINKED ISSUE #376: [BUG] ObservableCache Connect subscription omits concurrent updates
**Describe the bug**

There are two related bugs in `ObservableCache<TObject, TKey>` which cause the changeset sequence emitted by `Connect(Func<TObject, bool>)` sometimes to miss updates between the initial changeset and the update changesets which follow.

### Bug 1: `_changes` escapes lock scope in `Connect(Func<TObject, bool>)`
https://github.com/reactiveui/DynamicData/blob/1e61faa7e1db28e4ce1004f99c7d26cc59050a34/src/DynamicData/Cache/ObservableCache.cs#L196-L208

First, note that this method uses Observable.Defer, not Observable.Create. Observable.Defer returns an observable that when subscribed to:
1) executes the delegate provided (the "callback") to obtain a new observable; then
2) subscribes to it.

In ObservableCache, access to the Subject field `_changes` is (apparently) guarded by `_locker`. But the code above violates this constraint: within the callback's lock scope (executed in step 1) we simply read the `_changes` field and return this reference within a Concat observable. It is not until sometime after that the calling code subscribes to it (in _or after_ step 2), at which point the lock is no longer held. There is therefore a window where any updates to the cache occurring in another thread will appear neither in the initial changeset nor in any subsequent one, i.e. be lost.

An alternative implementation which does not exhibit this problem is as follows:
```cs
    public IObservable<IChangeSet<TObject, TKey>> Connect(Func<TObject, bool> predicate = null)
    {
        return Observable.Create<IChangeSet<TObject, TKey>>(o =>
        {
            lock (_locker)
            {
                var initial = GetInitialUpdates(predicate);
                var updates = predicate == null ? _changes : _changes.Filter(predicate);

                return Observable.Return(initial, ImmediateScheduler.Instance) // immediate ensures that `updates` is subscribed before we exit lock scope
                    .Concat(updates)
                    .NotEmpty()
                    .Subscribe(o);
            }
        });
    }
```
Here, the initial changeset construction *and the subscription to `_changes`* occur together atomically within the scope of the lock.

Note that this also fixes a relatively minor bug (call it 1b) in the original where if a predicate is given, the already-filtered initial changeset is redundantly filtered a second time.

### Bug 2: UpdateFromSource/UpdateFromIntermediate access to `_changes` without lock

https://github.com/reactiveui/DynamicData/blob/1e61faa7e1db28e4ce1004f99c7d26cc59050a34/src/DynamicData/Cache/ObservableCache.cs#L91
https://github.com/reactiveui/DynamicData/blob/1e61faa7e1db28e4ce1004f99c7d26cc59050a34/src/DynamicData/Cache/ObservableCache.cs#L122

In these excerpts from UpdateFromIntermediate and UpdateFromSource respectively `_readerWriter.Write`'s bool parameter `collectChanges` is determined by the value of `_changes.HasObservers`. If true, `Write` returns a non-empty changeset which is propagated to change observers (after correctly acquiring the `_locker` lock). `HasObservers` internally uses a volatile read to retrieve the very latest state.

This mechanism appears to exist to avoid the materialisation of changesets redundantly when no-one is interested.

However, a concurrent caller to `Connect` *is* interested in any changes that occur after the point of generating the initial changeset, including any that occur before the subscription to `_changes` has completed (which will cause HasObservers to become true). There is therefore a window during which `collectChanges` is erroneously determined to be false, resulting in lost changesets.

A simple solution with minimal downside (i.e. no new locking) would be to introduce a volatile flag field:
```cs
    private volatile bool _connectInProgress;
```
and to take it into account when determining `collectChanges`:
```cs
        var collectChanges = _changes.HasObservers || _connectInProgress;
        changes = _readerWriter.Write(updateAction, previewHandler, collectChanges);
```

Then it just needs to be set and unset within the `Connect` callback. Thus:
```cs
    public IObservable<IChangeSet<TObject, TKey>> Connect(Func<TObject, bool> predicate = null)
    {
        return Observable.Create<IChangeSet<TObject, TKey>>(o =>
        {
            lock (_locker)
            {
                _connectInProgress = true;
                try
                {
                    var initial = GetInitialUpdates(predicate);
                    var updates = predicate == null ? _changes : _changes.Filter(predicate);

                    return Observable.Return(initial, ImmediateScheduler.Instance) // immediate ensures that `updates` is subscribed before we exit lock scope
                        .Concat(updates)
                        .NotEmpty()
                        .Subscribe(o);
                }
                finally
                {
                    _connectInProgress = false;
                }
            }
        });
    }
```

**Steps To Reproduce**

Missing updates can be observed by subscribing to a cache that is undergoing frequent concurrent updates from another thread.

I think it likely that #319 and #333 are manifestations of this problem.

**Expected behavior**

The changeset sequence emitted by the observable returned by Connect should form a contiguous sequence, with no "gaps".

**Environment**
- Version: 6.14.14

pmg23:

Eliminates (some, possibly all) race conditions in ObservableCache and SourceList.

See #376, including comments.

**What is the new behavior?**

Observables returned by `Connect` and `CountChanged` are contiguous with respect to their initial state and subsequent updates.

**What might this PR break?**

It might make performance slower in applications with a high frequency of concurrent updates and subscriptions. (But note that these are applications that are especially prone to incorrect behaviour caused by the existing race conditions.)

**Other information**:

None

### DIFF
--- a/src/DynamicData/Cache/ObservableCache.cs
+++ b/src/DynamicData/Cache/ObservableCache.cs
@@ -21,7 +21,6 @@ namespace DynamicData
         private readonly ReaderWriter<TObject, TKey> _readerWriter;
         private readonly IDisposable _cleanUp;
         private readonly object _locker = new object();
-        private readonly object _writeLock = new object();
 
         private int _editLevel; // The level of recursion in editing.
 
@@ -80,7 +79,7 @@ namespace DynamicData
                 throw new ArgumentNullException(nameof(updateAction));
             }
 
-            lock (_writeLock)
+            lock (_locker)
             {
                 ChangeSet<TObject, TKey> changes = null;
 
@@ -111,7 +110,7 @@ namespace DynamicData
                 throw new ArgumentNullException(nameof(updateAction));
             }
 
-            lock (_writeLock)
+            lock (_locker)
             {
                 ChangeSet<TObject, TKey> changes = null;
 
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/reactiveui/DynamicData?pullRequest=377) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/reactiveui/DynamicData?pullRequest=377) before we can accept your contribution.<br/><br/>:x: pmg23 [sign now](https://cla.dotnetfoundation.org/reactiveui/DynamicData?pullRequest=377)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/reactiveui/DynamicData?pullRequest=377) it.</sub>

pmg23:
Thanks @RolandPheasant and @glennawatson.

May I ask when this might be released please? (Sorry, I'm not sure how much effort is involved on your part - but I'm keen now to upgrade to a version that includes this.)

glennawatson:
I will push out a new release tomorrow my time. Just a matter of changing version.json file and then hitting the release button.

PR #386, opened 2020-07-31
Optimize ListEx.Clone by using the is operator

jcummings2:
Fix inconsistent indentation
Add unit test for ListEx.Clone for Move on a plain List

Optimization/simplification and code cleanup.

As expected

**What is the new behavior?**
Same but with cleaner/optimized code and more unit tests to prove it works.

**What might this PR break?**
Nothing

**Other information**:


glennawatson:
The IL code that gets generated is the same between the two versions but your PR is worth it in terms of code simplification and the extra unit tests.

Thanks for submitting.

PR #401, opened 2020-09-02
Fix BindToObservableList: Previous values not being removed & sorting resets being ignored.

ryanholden8:
This is two bug fixes for the `BindToObservableList` extension method.
Improves unit tests to cover these situations.

1. Currently if you make a change to a model in an observable list that causes the updated model to move to a new index, the previous model is not removed from its old index.

2. When using `.Sort` and the reset threshold is hit, the observable list does not update it's data set because it did not receive a list of changes.

**What is the new behavior?**
1. The old model at the old index is now removed as expected.
2. SortReason.Reset is now detected to clear and re-apply using the SortedItems property.

**What might this PR break?**
This should only fix.

- [x] Docs have been added / updated (for bug fixes / features)

**Other information**:


RolandPheasant (approved):
This makes perfect sense to me. and good to see loads of unit tests

PR #440, opened 2020-11-17
feature: Update to rx 5.0.0 and .net 5

glennawatson:


RolandPheasant:
@glennawatson great work. One issue, is why are there so many whitespace changes?

glennawatson:
Brought into line with the same analyzer suite we use for the rest of the RXUI projects.

PR #460, opened 2021-02-28
FirstOrOptional extension fix to return Option.None instead of default on value type enumerable

Gotcha7770:
It would be great to get Option.None instead of the default, for example when searching through an int enumerable. If i get 0 i can't figure out if this is the default or the value that satisfies the condition.


RolandPheasant:
This makes sense

PR #464, opened 2021-03-08
SourceList should be unsealed

1R053:
### Problem
SourceList is currently a sealed class, which disallows inheriting from it. This makes it unnecessarily complex to add behaviours to lists, that should be reused.

For example I wanted to have a Copy function for specific variants of the list, that would provide type specific deep / shallow copying of the items.

Btw, the SourceCache class is not sealed. There it is easy to add those behaviours.

### Solution
Simply remove the sealed attribute... and as it then is required add a Dispose(bool) method.

### Impact / Risk
none - all current code should still run. Users should be responsible themselves when inheriting from SourceList

### DIFF
--- a/src/DynamicData/List/SourceList.cs
+++ b/src/DynamicData/List/SourceList.cs
@@ -17,7 +17,7 @@ namespace DynamicData
     /// An editable observable list.
     /// </summary>
     /// <typeparam name="T">The type of the object.</typeparam>
-    public sealed class SourceList<T> : ISourceList<T>
+    public class SourceList<T> : ISourceList<T>
     {
         private readonly ISubject<IChangeSet<T>> _changes = new Subject<IChangeSet<T>>();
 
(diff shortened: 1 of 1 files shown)


1R053:
What is the suggested approach to locally build & test? The debug build went successfully for this change, also running the test project did not show issues.

RolandPheasant:
Almost all objects in Dynamic Data are sealed by design as I am a firm advocate of composition over inheritance.   In cases, like the source cache not being sealed is a mistake.

1R053:
@RolandPheasant count me in on the firm composition advocates as well! However, C# is lacking in functionality here, mainly because traits/mixins are not well support. There are workarounds, sure, usually with extensions. But sometimes they are either complex to build/understand for new people in a project. Sometimes it simply does not meet requirements. And there are cases, where inheritance meets requirements better than C# composition.

That's why I favor to have both open and let the developers decide. There is nothing to loose imo for the project.

glennawatson:
Going to close this one. Spoke to Roland and the usage is deliberate

PR #392, opened 2020-08-04
[BUG] SortedObservableCollectionAdaptor does Remove -> Add instead of Replace when index not changed

radupetrisel:
**What kind of change does this PR introduce?**
Bug fix

**What is the current behavior?**
#391 

**What is the new behavior?**

SortedObservableCollectionAdaptor.DoUpdate uses Replace if the updated item's index does not change from the update.
The behavior stays the same if the index is changed.

**What might this PR break?**
Uses that rely on #391 

**Please check if the PR fulfills these requirements**
- [x] Tests for the changes have been added (for bug fixes / features)
- `not needed` Docs have been added / updated (for bug fixes / features)

PR #489, opened 2021-06-28
fix list sorting to replace without indexof

hutterm:
#392 introduced a change that slows down sorting.
The `list.Replace(...)` internally uses `IndexOf` to find the index of the item.
This is unnecessary as the index is known from the change


RolandPheasant:
This is kind of obvious - good spot and thanks.

Issue #393, opened 2020-08-05
[BUG] ObservableChangeSet with AutoRefresh and Observable filter predicate has unexpected behaviours


EmilienDup:
**Describe the bug**
Derived collection obtained from chaining `AutoRefresh` and `Filter` to an `ObservableChangeSet` has unexpected behavior.
In some situations the derived list is missing some items, in some other situations the derived collection has more items than the source collection.

**Steps To Reproduce**
I've created two unit tests to demonstrate the problem: https://github.com/EmilienDup/DynamicData/blob/main/src/DynamicData.Tests/AutoRefreshFilter.cs
(Forgive me for the poor unit test naming)

I can reproduce both bugs consistently on my machine.

I've tried to change the source collection type from `SourceList` to `ObservableCollection` (changing of course the way to get an `ObservableChangeSet` from it), and to bind the resulting `ObservableChangeSet` to a binding collection (through the `Bind(...)` extension).
I got the exact same results.

Do not hesitate to ask for any information, or more tests on my side.

RolandPheasant:
So sorry that this one  completely escaped my attention!  Well it's fixed now.

The issue came from the filter having to maintain the state of immutable objects and trying to track whether the item previously matched the filter or not and after a change, whether it now matches the filter.

EmilienDup:
No worries.
I'm glad that the pb could have been fixed.
Thanks a lot 👍

PR #514, opened 2021-10-05
List filter fix

RolandPheasant:
Fixes issue with mutable state filters for observable list - #393


RolandPheasant:
There are some code cleaning changes inadvertently added to this PR. I will however allow those to stand as they are minor changes (including some minor optimisations for the join operators).

PR #516, opened 2021-10-15
feat: Added DisposeManyStrict.

### LINKED ISSUE #512: [FEAT] Add IDisposable constaint to DisposeMany
Add `where TObject : IDisposable` to avoid unnecessary calls at compile time. In addition, during training, I was confused about the meaning of using `DisposeMany` for non-IDisposable types.

Pros:
- Avoids memory leaks when `DisposeMany` is applied to some Wrapper for an object that implements `IDisposable`.

For cases where the collection is of a base type (such as `object`) and the resulting classes can implement `IDisposable`, you can implement a separate statement that repeats the current behavior.

HavenDV:
Closes #512


HavenDV:
Yes, I think PR doesn't make sense, we need to think about adding an alternative operator with a strict restriction, like in PR.

RolandPheasant:
Closing because I do not think this brings value

PR #550, opened 2022-01-27
fix order of actual and expected values in asserts

rastislavsvoboda:
some of unit tests have switched actual and expected values, like
`3.Should().Be(_result.Data.Count);`
so switching it to
`_result.Data.Count.Should().Be(3);`
for correct message when test fails


RolandPheasant:
I really appreciate you doing this. Thanks very much.

PR #568, opened 2022-03-09
fix: Improve efficiency of AddKeyEnumerator

### LINKED ISSUE #563: Why does ToObservableChangeSet<T,TKey>() produce two changes when there is a replace?
I noticed that `ToObservableChangeSet<TObject, TKey>(this ObservableCollection<TObject> source, Func<TObject, TKey> keySelector)`
will produce an observable change set that behaves in what I would regard an odd way.

To me the idea of providing a keySelector indicates that I wish to nominate an **identity** value for every TObject. 

Consider this example:
```
        class Human
        {
            public Human(string name, int age)
            {
                Name = name;
                Age = age;
            }

            public string Name { get; }
            public int Age { get; }
        }
        ...
            var collection = new ObservableCollection<Human>();
            var observable = collection.ToObservableChangeSet(x => x.Name);
            using (observable.Subscribe(cs => Console.WriteLine(string.Join(",", cs.Select(c => c.Reason)))))
            {
                collection.Add(new Human("Fred", 43));
                collection[0] = new Human("Fred", 42);
            }
```
This code produces the output:
```
    Add
    Remove,Add
```
Given the identity selector that was provided, I would have expected it to produce:
```
    Add
    Update
```

After some digging it appears the behaviour emanates from AddKeyEnumerator.cs, which contains the following code:
```
                        {
                            // replace is a remove and add
                            var previous = change.Item.Previous.Value;
                            var previousKey = _keySelector(previous);
                            yield return new Change<TObject, TKey>(ChangeReason.Remove, previousKey, previous);

                            var current = change.Item.Current;
                            var currentKey = _keySelector(current);
                            yield return new Change<TObject, TKey>(ChangeReason.Add, currentKey, current);

                            break;
                        }
```

Would it not make sense here to extract both the keys here, and if they are equal then to treat the Change as an Update, rather than as sequential Add and Remove changes? 

For instance the following code produces the expected output, which is what I would regard as a more efficient observable:

```
                        {
                            // replace is a remove and add, IFF the keys do not match
                            var previous = change.Item.Previous.Value;
                            var previousKey = _keySelector(previous);

                            var current = change.Item.Current;
                            var currentKey = _keySelector(current);

                            if (Equals(currentKey, previousKey))
                            {
                                yield return new Change<TObject,TKey>(ChangeReason.Update, currentKey, current, previous);
                            }
                            else
                            {
                                yield return new Change<TObject, TKey>(ChangeReason.Remove, previousKey, previous);

                                yield return new Change<TObject, TKey>(ChangeReason.Add, currentKey, current);
                            }

                            break;
                        }
```

Am I missing something?

(BTW, thanks for the awesome library!)

pdegenhardt:
A replacement in an observable changeset with a key (i.e. cache style collection) should only produce an Add/Remove when the keys do not match. Otherwise, it should produce an Update.

Resolves #563

### DIFF
--- a/src/DynamicData/List/Linq/AddKeyEnumerator.cs
+++ b/src/DynamicData/List/Linq/AddKeyEnumerator.cs
@@ -55,7 +55,7 @@ internal class AddKeyEnumerator<TObject, TKey> : IEnumerable<Change<TObject, TKe
 
                 case ListChangeReason.Replace:
                     {
-                        // replace is a remove and add, IFF the keys do not match
+                        // replace is a remove and add, if and only if, the keys do not match
                         var previous = change.Item.Previous.Value;
                         var previousKey = _keySelector(previous);
 
--- a/src/DynamicData/List/Linq/AddKeyEnumerator.cs
+++ b/src/DynamicData/List/Linq/AddKeyEnumerator.cs
@@ -55,14 +55,23 @@ internal class AddKeyEnumerator<TObject, TKey> : IEnumerable<Change<TObject, TKe
 
                 case ListChangeReason.Replace:
                     {
-                        // replace is a remove and add
+                        // replace is a remove and add, IFF the keys do not match
                         var previous = change.Item.Previous.Value;
                         var previousKey = _keySelector(previous);
-                        yield return new Change<TObject, TKey>(ChangeReason.Remove, previousKey, previous);
 
                         var current = change.Item.Current;
                         var currentKey = _keySelector(current);
-                        yield return new Change<TObject, TKey>(ChangeReason.Add, currentKey, current);
+
+                        if (Equals(currentKey, previousKey))
+                        {
+                            yield return new Change<TObject, TKey>(ChangeReason.Update, currentKey, current, previous);
+                        }
+                        else
+                        {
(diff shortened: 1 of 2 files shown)


src/DynamicData/List/Linq/AddKeyEnumerator.cs
```diff
@@ -55,14 +55,23 @@ public AddKeyEnumerator(IChangeSet<TObject> source, Func<TObject, TKey> keySelec
 
                 case ListChangeReason.Replace:
                     {
-                        // replace is a remove and add
+                        // replace is a remove and add, IFF the keys do not match
```
glennawatson, line 58: `IF` not `IFF`?

pdegenhardt: if, and only if, (https://en.wikipedia.org/wiki/If_and_only_if)

glennawatson: Probably worth not using abbreviations in that case.

RolandPheasant (approved):
This is excellent. Thanks

PR #612, opened 2022-06-24
Vanquish the curse of transform async, but for List now

tomasfil:
Fixes parallel transform async, like it was done for cache in #565


RolandPheasant (approved):
That's awesome thanks

Issue #623, opened 2022-07-29
[Bug]: ArgumentOutOfRangeException/Index out of range exception when using TransformAsync


oysteinkrog:
### Describe the bug 🐞

```Exception: System.ArgumentOutOfRangeException
Message: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
Source: System.Private.CoreLib
TargetSite: Void ThrowArgumentOutOfRange_IndexException()
StackTrace:
   at void DynamicData.ChangeAwareList<T>.InsertItem(int index, T item) in /_/src/DynamicData/List/ChangeAwareList.cs:line 445
   at async Task DynamicData.List.Internal.TransformAsync<TSource, TDestination>.Transform(ChangeAwareList<TransformedItemContainer> transformed, IChangeSet<TSource> changes) in /_/src/DynamicData/List/Internal/TransformAsync.cs:line 78
   at async Task<ChangeAwareList<TransformedItemContainer>> DynamicData.List.Internal.TransformAsync<TSource, TDestination>+<>c__DisplayClass3_0.<Run>b__1(?)+(?) => { } in /_/src/DynamicData/List/Internal/TransformAsync.cs:line 52
   at void System.Reactive.PlatformServices.ExceptionServicesImpl.Rethrow(Exception exception) in /_/Rx.NET/Source/src/System.Reactive/Internal/ExceptionServicesImpl.cs:line 19
   at void System.Reactive.ExceptionHelpers.Throw(Exception exception) in /_/Rx.NET/Source/src/System.Reactive/Internal/ExceptionServices.cs:line 16
   at void System.Reactive.Stubs.Throw(Exception ex) in /_/Rx.NET/Source/src/System.Reactive/Internal/Stubs.cs:line 16
   at void System.Reactive.AnonymousSafeObserver<T>.OnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/AnonymousSafeObserver.cs:line 62
   at void System.Reactive.Sink<TTarget>.ForwardOnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/Internal/Sink.cs:line 60
   at void System.Reactive.Concurrency.Synchronize<TSource>+_.OnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/Concurrency/Synchronization.Synchronize.cs:line 50
[...]

PR #625, opened 2022-08-02
feature: Change TransformAsync implementation to be more similar to normal Transform

oysteinkrog:
- Reuse TransformedItemContainer.
- Add more overloads, like normal Transform
- Add transformOnRefresh, like normal Transform

### DIFF
--- a/src/DynamicData.Tests/List/TransformAsyncFixture.cs
+++ b/src/DynamicData.Tests/List/TransformAsyncFixture.cs
@@ -125,8 +125,12 @@ public class TransformAsyncFixture : IDisposable
         _results.Data.Count.Should().Be(0, "Should be nothing cached");
     }
 
+    /// <summary>
+    /// This test is disabled as it was flaky.
+    /// https://github.com/reactivemarbles/DynamicData/pull/625
+    /// </summary>
     [Fact]
-    public void TransformOnRefresh()
+    private void TransformOnRefresh()
     {
         var items = Enumerable.Range(1, 100).Select(i => new Person("Person" + i, 1)).ToArray();
 
--- a/src/DynamicData.Tests/List/TransformAsyncFixture.cs
+++ b/src/DynamicData.Tests/List/TransformAsyncFixture.cs
@@ -121,4 +121,41 @@ public class TransformAsyncFixture : IDisposable
     //    _results.Messages[1].Removes.Should().Be(100, "Should be 80 removes");
     //    _results.Data.Count.Should().Be(0, "Should be nothing cached");
     //}
+
+    // [Fact]
+    // public void TransformOnRefresh()
+    // {
+    //     var items = Enumerable.Range(1, 100).Select(i => new Person("Person" + i, 1)).ToArray();
+    //
+    //     //result should only be true when all items are set to true
+    //     using var list = new SourceList<Person>();
+    //     using var results = list.Connect().AutoRefresh(p => p.Age)
+    //         .TransformAsync(Task.FromResult, transformOnRefresh: true).AsAggregator();
+    //     list.AddRange(items);
+    //
+    //     results.Data.Count.Should().Be(100);
+    //     results.Messages.Count.Should().Be(1);
+    //
+    //     items[0].Age = 10;
+    //     results.Data.Count.Should().Be(100);
+    //     results.Messages.Count.Should().Be(2);
+    //
+    //     results.Messages[1].First().Reason.Should().Be(ListChangeReason.Replace);
+    //
+    //     //remove an item and check no change is fired
+    //     var toRemove = items[1];
+    //     list.Remove(toRemove);
+    //     results.Data.Count.Should().Be(99);
+    //     results.Messages.Count.Should().Be(3);
+    //     toRemove.Age = 100;
+    //     results.Messages.Count.Should().Be(3);
+    //
+    //     //add it back in and check it updates
+    //     list.Add(toRemove);
+    //     results.Messages.Count.Should().Be(4);
+    //     toRemove.Age = 101;
+    //     results.Messages.Count.Should().Be(5);
+    //
+    //     results.Messages.Last().First().Reason.Should().Be(ListChangeReason.Replace);
+    // }
 }
--- a/src/DynamicData.Tests/List/TransformAsyncFixture.cs
+++ b/src/DynamicData.Tests/List/TransformAsyncFixture.cs
@@ -1,7 +1,10 @@
 using System;
+using System.Linq;
 using System.Threading.Tasks;
 
 using DynamicData.Tests.Domain;
+using FluentAssertions;
+using Xunit;
 
 namespace DynamicData.Tests.List;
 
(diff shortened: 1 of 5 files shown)


oysteinkrog:
The failing test works on my computer, and does not seem to be using any of the code I changed..

RolandPheasant:
Thanks, l am on holiday right now (at 3.5k altitude in the Alps as I type ), so can't look into it. I'll get straight on the case as soon as I am home.

src/DynamicData.Tests/List/TransformAsyncFixture.cs
```diff
@@ -35,90 +38,127 @@ public void Dispose()
...
-    //public void Clear()
-    //{
-    //    var people = Enumerable.Range(1, 100).Select(l => new Person("Name" + l, l)).ToArray();
-
-    //    _source.AddRange(people);
-    //    _source.Clear();
-
-    //    _results.Messages.Count.Should().Be(2, "Should be 2 updates");
-    //    _results.Messages[0].Adds.Should().Be(100, "Should be 80 addes");
-    //    _results.Messages[1].Removes.Should().Be(100, "Should be 80 removes");
```
glennawatson, line 121: Try not to have old code floating around.

oysteinkrog: Hm yeah, this entire test fixture was commented out. I added a new test, but commented it out too, as I did not want to risk hanging the CI (the documented reason for why the test was commented out).

src/DynamicData.Tests/List/TransformAsyncFixture.cs
```diff
@@ -35,90 +38,127 @@ public void Dispose()
...
+
+    [Fact]
+    public void TransformOnRefresh()
+    {
+        var items = Enumerable.Range(1, 100).Select(i => new Person("Person" + i, 1)).ToArray();
+
+        //result should only be true when all items are set to true
+        using var list = new SourceList<Person>();
+        using var results = list.Connect().AutoRefresh(p => p.Age)
+            .TransformAsync(async v => v, transformOnRefresh: true).AsAggregator();
```
RolandPheasant, line 136: v => Task.FromResult(v) will fix this warning

src/DynamicData/List/Internal/TransformAsync.cs
```diff
@@ -88,25 +97,44 @@ private async Task Transform(ChangeAwareList<TransformedItemContainer> transform
...
 
+                case ListChangeReason.Refresh:
+                    {
+                        var change = item.Item;
+                        if (_transformOnRefresh)
+                        {
+                            Optional<TDestination> previous = transformed[change.CurrentIndex].Destination;
+                            var container = await _containerFactory(change.Current, previous, change.CurrentIndex)
+                                .ConfigureAwait(false);
+                            transformed.Refresh(container, change.CurrentIndex);
```
RolandPheasant, line 115: `transformed.Refresh(container, change.CurrentIndex); `

should be 

` transformed[change.CurrentIndex] = container;`

because technically transform on refresh is an instruction that the we we need to re-apply a transform which creates (or normally creates) a new item.

On my PC, this fixes the  List.TransformAsyncFixture.TransformOnRefresh test

RolandPheasant (changes requested):
Thanks for this PR. A couple minor changes required then it can be merged.

src/DynamicData/List/Internal/TransformAsync.cs
oysteinkrog: 👍

src/DynamicData.Tests/List/TransformAsyncFixture.cs
oysteinkrog: 👍

oysteinkrog:
Thanks for review!
PR updated with suggested changes :)

oysteinkrog:
@RolandPheasant With your suggested change, the test I added fails. I'm not sure if the test is wrong or the implementation...

RolandPheasant:
> @RolandPheasant With your suggested change, the test I added fails. I'm not sure if the test is wrong or the implementation...

That's really odd. I just run it again locally and it passes!  Does it run locally for you too?

oysteinkrog:
> 

It passes for me locally :P

oysteinkrog:
> > @RolandPheasant With your suggested change, the test I added fails. I'm not sure if the test is wrong or the implementation...
> 
> That's really odd. I just run it again locally and it passes! Does it run locally for you too?

It passes for me locally :P

RolandPheasant:
> It passes for me locally :P

I have seen such issues before when running code on a build server (when using Tasks within RX).   It's a bit of a guess but I suspect the issue comes as a result of parallelised running of tests where for some reason task awaiting behaves oddly. Perhaps a context switch due to internal `.ConfigureAwait(false)` usage plays weirdly with the test runner - if that's the case it would just be a timing issue which is not an actual issue but is a test issue.

I am happy the code works and suggest we ignore the test by making it private, not public. Add a comment with a link to this discussion.

oysteinkrog:
> > It passes for me locally :P
> 
> I have seen such issues before when running code on a build server (when using Tasks within RX). It's a bit of a guess but I suspect the issue comes as a result of parallelised running of tests where for some reason task awaiting behaves oddly. Perhaps a context switch due to internal `.ConfigureAwait(false)` usage plays weirdly with the test runner - if that's the case it would just be a timing issue which is not an actual issue but is a test issue.
> 
> I am happy the code works and suggest we ignore the test by making it private, not public. Add a comment with a link to this discussion.

Hm ok, I did as you suggested, but I am concerned that this problem indicates there is a deeper bug somewhere in the implementation (not in the test).

RolandPheasant:
@oysteinkrog I've worked out the issue. 

I tried running all the tests on your branch and found there was several failures including with AutoRefresh and Transform. This struck me as odd, then i realised you'd branched from an old branch and there's been a load of changes since.

So I grabbed a later branch and manually copied all your code over to the clean branch and all the tests run.

Before on your branch.

image

Code cloned to clean branch

image

So the choice is either you repeat what I did on a new PR, or alternatively I could check the changes in on my branch.  What do you think?

oysteinkrog:
> @oysteinkrog I've worked out the issue.
> 
> I tried running all the tests on your branch and found there was several failures including with AutoRefresh and Transform. This struck me as odd, then i realised you'd branched from an old branch and there's been a load of changes since.
> 
> So I grabbed a later branch and manually copied all your code over to the clean branch and all the tests run.
> 
> Before on your branch.
> 
> image
> 
> Code cloned to clean branch
> 
> image
> 
> So the choice is either you repeat what I did on a new PR, or alternatively I could check the changes in on my branch. What do you think?

I'm not sure I understand this, I just rebased the PR to latest main branch. Is there another branch I should base my PR on?
I don't mind either way, I really just want to get this in, as for me (in our product) it also fixes #623 :)

RolandPheasant:
I can't entirely explain why myself but sometimes merge, rebases or force push does not always behave perfectly. However what I also cannot understand is why the AutoRefresh tests plus one of the Transform (not TransformAsync!) fails on your branch (and also on the build server).  That's why for an experiment I manually copied the code over to a clean branch and the results are clear .

I suggest leaving this PR open for now.  I'll raise a PR with the cloned changes and lets see what happens. It may be that the TransformAsync.TransformOnRefresh stills fails, but there will be no harm trying.

src/DynamicData/List/Internal/Transformer.cs
```diff
@@ -82,15 +82,16 @@ private void Transform(ChangeAwareList<TransformedItemContainer> transformed, IC
...
                 case ListChangeReason.Refresh:
                     {
+                        var change = item.Item;
                         if (_transformOnRefresh)
                         {
-                            var change = item.Item;
                             Optional<TDestination> previous = transformed[change.CurrentIndex].Destination;
-                            transformed[change.CurrentIndex] = _containerFactory(change.Current, previous, change.CurrentIndex);
+                            var newItem = _containerFactory(change.Current, previous, change.CurrentIndex);
+                            transformed.Refresh(newItem, change.CurrentIndex);
```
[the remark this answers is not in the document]
oysteinkrog: I've changed this back, I understand what you are saying, I don't remember why I even did this change.
My PR is updated now

RolandPheasant (approved):

PR #633, opened 2022-08-22
list/TransformAsync: Add locking to fix ArgumentOutOfRangeException

oysteinkrog:
Fixes issue #623

I am not confident this is the correct way to fix this, please advise.

### DIFF
--- a/src/DynamicData/List/Internal/TransformAsync.cs
+++ b/src/DynamicData/List/Internal/TransformAsync.cs
@@ -57,7 +57,15 @@ internal class TransformAsync<TSource, TDestination>
             transformed =>
             {
                 var changed = transformed.CaptureChanges();
-                return changed.Transform(container => container.Destination);
+                try
+                {
+                    asyncLock.Wait();
+                    return changed.Transform(container => container.Destination);
+                }
+                finally
+                {
+                    asyncLock.Release();
+                }
             });
     }
 


src/DynamicData/List/Internal/TransformAsync.cs
```diff
@@ -57,7 +57,15 @@ private IObservable<IChangeSet<TDestination>> RunImpl()
             transformed =>
             {
                 var changed = transformed.CaptureChanges();
-                return changed.Transform(container => container.Destination);
+                try
```
RolandPheasant, line 60: I don't think this is the solution.  Have you been able to replicate the issue in test code?

oysteinkrog: Ok.
No I have not tried to create a test.
The code/pattern in our application is this:
```            _availableLocations = possibleEndPoints.Connect()
                .AutoRefreshOnObservable(_ => Observable.Timer(5.s(), 5.s(), schedulers.Default)
                    .SwitchWithEmpty(enablePolling))
                // TransformAsync does not (yet) have transformOnRefresh parameter, so we need this hack
                .RewriteRefreshToAddRemove()
                .TransformAsync(v => _pingService.TryPingAsync(v))
                .WhereNotNull()
                .TransformAsync(v => _tcpConnectionTester.TryConnectAsync(v))
                .WhereNotNull()
                .DistinctValues(v => v)
                .AsObservableList();
```
We kind of abuse TransformAsync as a filter, and I created this RewriteRefreshToAddRemove() extensions since TransformAsync did not yet have the transformOnRefresh parameter.

oysteinkrog: Our testing seems to indicate that the changes in this PR fixes #623 _for us_. Obviously that is not to say that it is actually a good fix, just one datapoint.

RolandPheasant: Normally (except in the async case), locking would be achieved as such:

```cs
.Select(tasks => tasks.ToObservable()).SelectMany(items => items)
.Synchronize(someMontitorLock)
.Select(transformed =>
```
However in this case being as we already have a semaphore lock, maybe it makes more sense to use it.  I'll sit on this a little to consider whether this is best.

RolandPheasant: Alternative I also wonder whether the actual issue is caused by the ConfigureAwait(false) clauses meaning that every result effectively continues on a different thread.

oysteinkrog: After more extensive testing we find that the proposed changes in this PR does not actually fix the problem, it just makes it less likely to occur.

RolandPheasant: I think we should both try and produce a replication

PR #636, opened 2022-08-31
[Feature] TransformWithInlineUpdate 

### LINKED ISSUE #323: Feature Request: TransformWithInlineUpdate
**Is your feature request related to a problem? Please describe.**
Well not exactly a problem but it could benefit everyone that's using the DD on mobile (and other platforms too): Shuffling on the lists and unnecessary update of  whole item in the UI due to the replacement of items even when their `TKey` is the same as the previous cached version. That means the items are equal (not the same reference) but only a property has changed.

The issue I'm talking about is mentioned in this issue:
#240 

The real life case:
In my most recent projects I've been using `Akavache` as fetcher of data and caching mechanism, and `DynamicData` to provide the data through my services connecting them through the viewmodels. That's a nice duo, but I'm constantly being hit by the combination of `GetAndFetchLatest` (gets from the cache and from the remote) + `AddOrUpdate`/`EditDiff` because the replacement of items is making my lists to shuffle when a property of the object gets changed on the remote.

**Describe the solution you'd like**
As already discussed in the issue #240 above there are two proposed solutions:
- one is to provide a `Transform` that receives an update action as argument, so that could be implemented for each specific case before presenting the data at the viewmodel level (for my case at least).
- the second one is to provide something closer to the cache itself, like the `EditDiff` and `AddOrUpdate`, making use of the `Edit` to perform the update action as passed in.

I think both of these mechanisms would be great additions to the framework. For my case the **second one** would be the perfect addition, to not have to rely on it as an extension snippet that I copy and paste into my project. 
With them in place I imagine it would cover most of the cases people are using, having this optimization out of the box to be used. With that we could fine grain how the items would get updated and when.

**Describe alternatives you've considered**
No others for now.

**Describe suggestions on how to achieve the feature**
They are already proposed on the issue #240

**Additional context**

tomasfil:
Adds TransformWithInline update. 

Resolves #323

### DIFF
--- a/src/DynamicData/Cache/ObservableCacheEx.cs
+++ b/src/DynamicData/Cache/ObservableCacheEx.cs
@@ -5050,6 +5050,83 @@ public static class ObservableCacheEx
     }
 
     /// <summary>
+    /// Projects each update item to a new form using the specified transform function and when an update is received, allows the preservation of the previous instance.
+    /// </summary>
+    /// <typeparam name="TDestination">The type of the destination.</typeparam>
+    /// <typeparam name="TSource">The type of the source.</typeparam>
+    /// <typeparam name="TKey">The type of the key.</typeparam>
+    /// <param name="source">The source.</param>
+    /// <param name="transformFactory">The transform factory.</param>
+    /// <param name="updateAction">The update action.</param>
+    /// <returns>
+    /// A transformed update collection.
+    /// </returns>
+    /// <exception cref="System.ArgumentNullException">source
+    /// or
+    /// transformFactory.</exception>
+    public static IObservable<IChangeSet<TDestination, TKey>> TransformWithInlineUpdate<TDestination, TSource, TKey>(this IObservable<IChangeSet<TSource, TKey>> source, Func<TSource, Optional<TSource>, TKey, TDestination> transformFactory, Action<TDestination, TSource> updateAction)
+        where TKey : notnull
+    {
+        if (source is null)
+        {
+            throw new ArgumentNullException(nameof(source));
+        }
+
+        if (transformFactory is null)
+        {
+            throw new ArgumentNullException(nameof(transformFactory));
+        }
+
+        if (updateAction is null)
+        {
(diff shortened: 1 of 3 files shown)


RolandPheasant (changes requested):
Sorry for the delay in responding. Thanks for this, and would you mind adding some unit tests

tomasfil:
Sorry for delay from me as well. I am very busy past few days. I will try my best to add unit tests asap as I am able to

src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -5049,6 +5049,86 @@ public static IObservable<IReadOnlyCollection<T>> StartWithEmpty<T>(this IObserv
...
+    /// <param name="source">The source.</param>
+    /// <param name="transformFactory">The transform factory.</param>
+    /// <param name="updateAction">The update action.</param>
+    /// <returns>
+    /// A transformed update collection.
+    /// </returns>
+    /// <exception cref="System.ArgumentNullException">source
+    /// or
+    /// transformFactory.</exception>
+    public static IObservable<IChangeSet<TDestination, TKey>> TransformWithInlineUpdate<TDestination, TSource, TKey>(this IObservable<IChangeSet<TSource, TKey>> source, Func<TSource, Optional<TSource>, TKey, TDestination> transformFactory, Action<TDestination, TSource> updateAction)
```
RolandPheasant, line 5067: You could combine these 2 overloads into one by making the error handler an optional param.

SuperJMN:
> Adds TransformWithInline update.
> 
> Resolves #323

I've created a sample to see this working: https://github.com/SuperJMN-Tutorials/TreeDataGrid-InlineUpdate-WithGrouping

tomasfil:
Hello @RolandPheasant, I did manage to add some tests and I hope it is sufficient.

RolandPheasant:
This is great thanks.

SuperJMN:
Thanks! I'll use this ASAP.

PR #639, opened 2022-09-06
Serialize TransformAsync using Concat

RolandPheasant:
Use Concat to ensure serialized sequence for List.TransformAsync.

Hopefully a fix for #623, and attempts to address #633 

@oysteinkrog I tried to replicate the issue in a unit test and could not do so.  This may work but either way it is a more correct way of handling tasks within an observable.  I'll probably just merge it and get you to test.


RolandPheasant:
@oysteinkrog can you try 7.10.2 and see whether it fixes the issue

PR #664, opened 2022-11-18
Remove try/finally from SelectMany in TransformAsync, which was causing race condition for some reason

### LINKED ISSUE #663: [Bug]: When using TransformAsync, then Bind sometimes does not remove item from final collection
### Describe the bug 🐞

With TransformAsync and Bind behind that, when item is rapidly added and then removed, then once SourceCache is empty, for some reason and only sometimes there are left items in the destination ReadOnlyObservableCollection.

### Step to reproduce

I have created breaking Unit test in repro repo branch TransformAsyncWithBind

### Reproduction repository

https://github.com/tomasfil/DynamicData/tree/TransformAsyncWithBind

### Expected behavior

This should not happen...

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

_No response_

### Version

_No response_

### Device

_No response_

### DynamicData Version

_No response_

### Additional information ℹ️

_No response_

tomasfil:
Fixes #663

### DIFF
--- a/src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
+++ b/src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
@@ -64,43 +64,6 @@ public class TransformAsyncFixture
     }
 
     [Fact]
-    public async Task RemoveFlowsToTheEnd()
-    {
-        int transform = 0;
-        int count = 500;
-        ReadOnlyObservableCollection<Person> collection;
-
-        using var stub = new TransformStub();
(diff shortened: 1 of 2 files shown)


tomasfil:
This PR might be terrible, or just a workaround, someone smarter might identify the cause of the race condition, but as is seen in the issue #663, the actual issue is wild and someone might understand why that happens better

RolandPheasant (approved):
Let's hope this fixes it for all time !

PR #689, opened 2023-03-04
feature: Add overload of TransformMany that works with child instances of IObservableCache

John-MartinMalone:
Add overload of TransformMany that works with child instances of IObservableCache


RolandPheasant (changes requested):
Just one thing, can you update the version found here:

[version.json#L2](https://github.com/reactivemarbles/DynamicData/blob/d1ca57296711630d7bb2e837d1888c3408953c45/version.json#L2
)

to 7.14

glennawatson (approved):

glennawatson:
@RolandPheasant I've updated the version indicator for @John-MartinMalone

PR #694, opened 2023-04-04
Make BindingList adaptors replace the changed item instead of re-adding

Metadorius:
Before it was deleting and adding the item again, resulting in two notifications for deleting and adding an item. This resulted in UI deselecting the updated item if it was selected, which is an undesired behavior. After this change it doesn't do that anymore.


RolandPheasant (approved):
Technically if the item is not found (ie index -1) an exception should be thrown. However I have simply ignored  those cases in binding. But food for thought for future consideration.

Issue #692, opened 2023-03-23
[Bug]: Change<T> does not work correctly with null items


PhilipCavanagh:
### Describe the bug 🐞

- When an item in a `SourceList<T>` is replaced, the creation of the `Change<T>` requires that the parameter `Optional<T> previous` has a value.
- If `T` is nullable, the previous value may be `null`.
- It is not possible to create an `Optional<T>` with a value of `null`.

Either `Change<T>` should mandate that `T` be `notnull`, or `Optional<T>` should not assume that `null` means "no value".

### Step to reproduce

```
var source = new SourceList<object?>();

using (source.Connect().Subscribe())
{
    source.Add(null);
    source.ReplaceAt(0, new object()); // This line throws an `ArgumentException` from `Change<T>` constructor
}
```

### Reproduction repository

https://github.com/PhilipCavanagh/DynamicData/blob/main/src/DynamicData.Tests/List/SourceListBug.cs

### Expected behavior

The single `null` item in the `SourceList` should be replaced with the new object.

### DynamicData Version

7.13.1

RolandPheasant:
Not allowing nulls is by design.

PhilipCavanagh:
In that case, could all the type parameters have `notnull` constraints added?

glennawatson:
`notnull` makes sense to add. Makes sense it doesn't have it due to it being a newer C# feature.

RolandPheasant:
Fixed by #706

PR #706, opened 2023-06-12
Bugfix/forbid null items

PhilipCavanagh:
In response to #692, this PR adds a `where t : notnull` constraint to `Option<T>`.
That change proliferates out to `IChangeSet` and into a lot of extension methods.

I think this constitutes a breaking change, but code that used nullable types may not have worked correctly anyway.

### DIFF
--- a/src/DynamicData/Kernel/Optional.cs
+++ b/src/DynamicData/Kernel/Optional.cs
@@ -17,6 +17,7 @@ namespace DynamicData.Kernel;
 [SuppressMessage("Naming", "CA1716:Identifiers should not match keywords", Justification = "Deliberate usage.")]
 [SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1402:File may only contain a single type", Justification = "Class names the same, generic differences.")]
 public readonly struct Optional<T> : IEquatable<Optional<T>>
+    where T : notnull
 {
     private readonly T? _value;
 
@@ -185,7 +186,9 @@ public static class Optional
     /// </summary>
     /// <typeparam name="T">The type of the item.</typeparam>
     /// <returns>The optional value.</returns>
-    public static Optional<T> None<T>() => Optional<T>.None;
+    public static Optional<T> None<T>()
+        where T : notnull
+        => Optional<T>.None;
 
     /// <summary>
     /// Wraps the specified value in an Optional container.
@@ -193,5 +196,7 @@ public static class Optional
     /// <typeparam name="T">The type of the item.</typeparam>
     /// <param name="value">The value.</param>
     /// <returns>The optional value.</returns>
-    public static Optional<T> Some<T>(T value) => new(value);
+    public static Optional<T> Some<T>(T? value)
+        where T : notnull
+        => new(value);
 }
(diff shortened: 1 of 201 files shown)


glennawatson:
Due to the size of the change and integration testing required it won't be merged straight away. Just a heads up.

PhilipCavanagh:
> Due to the size of the change and integration testing required it won't be merged straight away. Just a heads up.

I imagined as such. I don't think it should change any behaviour, but it could break existing code if it uses nullable types.

RolandPheasant:
I apologise for a long delay. I'd been pondering on whether we should do this, then I forgot !

The one thing which concerns me is that there is potentially a use case for null.  A "fix" was introduced to enable it in the case of a transform to allow nulls.  I believe the idea was to immediately filter the null out.  Doing it this way enabled being able to suppress transforms in certain circumstances.

@oysteinkrog can you remember why you did this, and why I approved - See https://github.com/reactivemarbles/DynamicData/pull/216

Regardless, nulls are still a bad idea and I wonder whether we should push this and perhaps introduce a mechanism to Dynamic Data where a null returned by a transform results in a remove event rather than a null update.

src/DynamicData/Kernel/Optional.cs
```diff
@@ -185,13 +186,17 @@ public static class Optional
...
+        => Optional<T>.None;
 
     /// <summary>
     /// Wraps the specified value in an Optional container.
     /// </summary>
     /// <typeparam name="T">The type of the item.</typeparam>
     /// <param name="value">The value.</param>
     /// <returns>The optional value.</returns>
-    public static Optional<T> Some<T>(T value) => new(value);
+    public static Optional<T> Some<T>(T? value)
```
RolandPheasant, line 199: Value could be null as the constructor of optional directly casts a null to a None<T>.

RolandPheasant: Doh. Ignore me as I just noticed the T?

oysteinkrog:
> 



> I apologise for a long delay. I'd been pondering on whether we should do this, then I forgot !
> 
> The one thing which concerns me is that there is potentially a use case for null. A "fix" was introduced to enable it in the case of a transform to allow nulls. I believe the idea was to immediately filter the null out. Doing it this way enabled being able to suppress transforms in certain circumstances.
> 
> @oysteinkrog can you remember why you did this, and why I approved - See #216
> 
> Regardless, nulls are still a bad idea and I wonder whether we should push this and perhaps introduce a mechanism to Dynamic Data where a null returned by a transform results in a remove event rather than a null update.

Hmm yes, we used to do this as we have a Transform that is essentially using a Try-pattern, meaning that it may fail.
The code we had that did this has been rewritten now and is now essentially using an Optional<> wrapper to represent the success/failure of the Transform, with an accompanying Filter afterwards, just like you describe.
I support this PR, it's a good change. We were using RX/DD for years before we truly realized that nulls are _not_ supported in RX, and I think making DD consistent with the RX API/contract is a good idea.
This change would have prevented other null-related errors in our RX+DD code :)

RolandPheasant:
Thanks for the input @oysteinkrog, that settles it.  This PR is good to go.  

@glennawatson I'll approve this, but before we release I'd like to do a little house keeping within the codebase.  I'll be done with that within a few days.

RolandPheasant:
Also @oysteinkrog I may additionally provide the option within the transform to return a null, in which case the item would be removed.

RolandPheasant (approved):
This is a an excellent contribution to the library,

PR #726, opened 2023-09-21
Adds optional useReplaceForUpdates to SortedObservableCollectionAdaptor

ccopsey:
In a similar vein to how `useReplaceForUpdates` in `ObservableCollectionAdaptor` toggles between `Replace` and `Remove/Add` update types, this pull request proposes adding the same capability to `SortedObservableCollectionAdaptor`.

Default to `true` for backward compatibility, although if a new major version is in the offing, maybe this could be `false` to match `ObservableCollectionAdaptor`.

The test has been updated to match the equivalent test in `ObservableCollectionBindCacheFixture`.

### DIFF
--- a/src/DynamicData.Tests/Binding/ObservableCollectionBindCacheSortedFixture.cs
+++ b/src/DynamicData.Tests/Binding/ObservableCollectionBindCacheSortedFixture.cs
@@ -196,22 +196,42 @@ public class ObservableCollectionBindCacheSortedFixture : IDisposable
     [Fact]
     public void UpdateToSourceSendsReplaceIfSortingIsNotAffected()
     {
-        var person1 = new Person("Adult1", 10);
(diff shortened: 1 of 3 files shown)


ccopsey:
The failing test has nothing to do with me.  I will wait to pull a fix for that.

RolandPheasant:
> The failing test has nothing to do with me. I will wait to pull a fix for that.

Years ago I made a test private as it was flaky, occasionally failed and was a false negative.  For some strange reason the test runs after a minor version test library update, so I have removed the [Fact] annotation altogether to ignore the test.

Just to warn you when you rebase, there are a load of changes as  I merged a PR which applies a not null restraint to cache and list items. This may lead to conflicts.

RolandPheasant (approved):
Thanks

Issue #461, opened 2021-03-01
Add async support for TransformMany


JunkyXL86:
There is an async version TransformAsync() accepting a Task<T> as return value.
Similarly there should be an async version for TransformMany() to achieve the same with collections.

Is there an actual workaround to get this working?

RolandPheasant:
Ideally, this should be implemented

RolandPheasant:
Closed by the mentioned PR

dwcullop:
Sorry but that PR does not address this.  I was just being too chatty and mentioned what other issues I was considering addressing.  My apologies.

dwcullop:
`TransformManyAsync` and `TransformManySafeAsync` where implemented by #824.

Issue #409, opened 2020-09-30
Feature: Consistent FilterOnObservable, SortOnObservable, GroupOnObservable for List and Cache like FilterOnObservable


KeilFelix:
Please consider providing these additional extension methods:

- FilterOnObservable (already available for IChangeSet&lt;TObject&gt; but not for IChangeSet&lt;TObject, TKey&gt;
- SortOnObservable
- GroupOnObservable

If you think the desired behavior is not clear from the method names, I can specify it further.

Further suggestion is to ditch the "OnObservable" and just have them as overloads.

RolandPheasant:
Filter and Sort already have these overloads (without the OnObservable).

Regarding group, I am not sure how the c# type system could handle grouping on different keys?

KeilFelix:
@RolandPheasant Maybe I don't understand how to achieve the same functionality with the overloads you mentioned. But none of the overloads uses a `Func<TObject, IObservable<bool>> objectFilterObservable` like the FilterOnObservable-method.

Sample imagine an `ISourceList<IDevice>` or  Cache

```
    public interface IDevice
    {
        IObservable<bool> IsOnline { get; }
        IObservable<int> Priority { get; }
    }
```

I want to Sort/Filter/Group based on the properties inside

RolandPheasant:
I am considering marking FilterOnObservable as obsolete as it an awful operator beset with problems.  There is nothing it does which cannot be achieved by other means.

So closing this.

PR #735, opened 2023-10-05
Feature: Add FilterOnObservable for Cache ChangeSets

dwcullop:
### Description
This PR implements the `FilterOnObservable` operator for Cache ChangeSets (the operator already exists exists for List ChangeSets).  It filters a ChangeSet by using the given factory function to create an `IObservable<bool>` for each item that controls whether or not that item passes the filter.  No items will pass the filter until the associated observable fires `true`.

This operator enables filtering changesets by something more complicated than a simple property value.  For example, if the items in the ChangeSet have a sub-cache, you can use it to filter them by the contents of their individual sub-caches.

I originally had cloned the code from `AutoRefresh`, but I changed to just using the `AutoRefreshOnObservable` operator to reduce the code size because it is exactly what is needed.

### Examples

Filter by main cache by the number of items in the sub-cache:

```C#
interface IPerson
{
    int Id { get; }
    string Name { get; }
    IObservableCache<IPerson, int> Children { get; }
}

var peopleDirectory = new SourceCache<IPerson, int>(p => p.Id);

// Observe the source cache for people with at least 2 children
using var cleanup = peopleDirectory.Connect()
                                   .FilterOnObservable(p => p.Children.CountChanged.Select(count => count >= 2))
                                   .OnItemAdded(p => Console.WriteLine($"{p.Name} has at least 2 kids!"))
                                   .Subscribe();
```

Create a view of the SourceCache based on how long ago the items were added to the cache (useful if you don't want stale items but don't want to remove them for the main cache).

```C#

IObservable<bool> CreateTimerFilter(TimeSpan timeSpan, bool filterResult) =>
    Observable.Timer(timeSpan).Select(_ => filterResult).StartWith(!filterResult);

// Special handling of entries older than 5 minutes
using var cleanup1 = observableCache.Connect()
                                    .FilterOnObservable(_ => CreateTimerFilter(TimeSpan.FromMinutes(5), true))
                                    .OnItemAdded(entry => HandleStaleEntry(entry))
                                    .Subscribe();

// Show brand new entries in a special UI list
using var cleanup2 = observableCache.Connect()
                                    .FilterOnObservable(_ => CreateTimerFilter(TimeSpan.FromSeconds(30), false))
                                    .ObserveOnDispatcher()
                                    .Bind(var out newEntryList)
                                    .Subscribe();
```

### Unit Tests

I've added unit tests but please let me know if there are any use-cases that I've missed.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -34,7 +34,7 @@ jobs:
         echo "##vso[task.prependpath]$(vs where release --prop=InstallationPath)\MSBuild\Current\Bin"
         
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         fetch-depth: 0
         lfs: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,7 @@ jobs:
         echo "##vso[task.prependpath]$(vs where release --prop=InstallationPath)\MSBuild\Current\Bin"
        
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         fetch-depth: 0
 
--- a/src/DynamicData.Tests/Binding/ObservableCollectionBindCacheSortedFixture.cs
+++ b/src/DynamicData.Tests/Binding/ObservableCollectionBindCacheSortedFixture.cs
@@ -196,22 +196,42 @@ public class ObservableCollectionBindCacheSortedFixture : IDisposable
     [Fact]
     public void UpdateToSourceSendsReplaceIfSortingIsNotAffected()
     {
-        var person1 = new Person("Adult1", 10);
-        var person2 = new Person("Adult2", 11);
+        RunTest(true);
+        RunTest(false);
 
-        NotifyCollectionChangedAction action = default;
-        _source.AddOrUpdate(person1);
-        _source.AddOrUpdate(person2);
 
-        var person2Updated = new Person("Adult2", 12);
-
-        using (_collection.ObserveCollectionChanges().Select(change => change.EventArgs.Action).Subscribe(act => action = act))
+        void RunTest(bool useReplace)
         {
-            _source.AddOrUpdate(person2Updated);
-        }
+            var collection = new ObservableCollectionExtended<Person>();
 
-        action.Should().Be(NotifyCollectionChangedAction.Replace, "The notification type should be Replace");
-        _collection.Should().Equal(person1, person2Updated);
+            using var source = new SourceCache<Person, string>(p => p.Name);
+            using var binder = source.Connect().Sort(_comparer, resetThreshold: 25).Bind(collection, new ObservableCollectionAdaptor<Person, string>(useReplaceForUpdates: useReplace)).Subscribe();
+
+            var person1 = new Person("Adult1", 10);
+            var person2 = new Person("Adult2", 11);
+
+            NotifyCollectionChangedAction action = default;
+            source.AddOrUpdate(person1);
+            source.AddOrUpdate(person2);
+
+            var person2Updated = new Person("Adult2", 12);
+
+            using (collection.ObserveCollectionChanges().Select(x => x.EventArgs.Action).Subscribe(updateType => action = updateType))
+            {
+                source.AddOrUpdate(person2Updated);
+            }
+
+            if (useReplace)
+            {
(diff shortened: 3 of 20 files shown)


dwcullop:
This implements one of the requests from #409.

RolandPheasant:
> This implements one of the requests from #409, but that issue has been closed by @RolandPheasant because the list version has issues. I'd like to understand the nature of the issues to make sure this PR doesn't have the same problems. If it doesn't, this implementation might be easily ported to support Lists.

If my memory serves my right,  the issues were mainly down to performance and scalability.   I think the main reason is due to the difficulty with optimising observables of unknow nature. However, my thought is if it exists with the list side of the system then it should also exist here.   

Also maybe we should create some additional wiki pages expressing scalability concerns of certain operators.  Although I suspect in most cases people are not using 50k records in a dataset and scalability is not everyone's concern.

dwcullop:
> If my memory serves my right, the issues were mainly down to performance and scalability. I think the main reason is due to the difficulty with optimising observables of unknow nature. However, my thought is if it exists with the list side of the system then it should also exist here.

Cache is always going to give better performance, so it might not be an issue.  This implementation uses `AutoRefreshOnObservable`.  If Unknown Observables are okay for that operator, then they should be okay here.  I can add some performance testing to see if there is an issue.

> Also maybe we should create some additional wiki pages expressing scalability concerns of certain operators. Although I suspect in most cases people are not using 50k records in a dataset and scalability is not everyone's concern.

I agree with you 100% on this.  Just because *some* people might use it with heavy operators and end up with scalability issues when they have a large dataset, it doesn't mean that it won't be useful to some people who use simple observables and/or smaller datasets.  All of these things are situational.  The solution is to document the ones that might cause scalability issues and let the consumers decide what will work best for them.

Such things might be mitigated by using a hot observable that was going to run anyway.  Then the overhead from filtering is minimal.

I'll do some perf testing and see if there's a problem and try to optimize the implementation as much as I can.  I can't promise the approach taken will work for Lists, but Caches always have the advantage with regard to performance.  Even if you have to deprecate the operator for lists, I would still like to see it for Cache because I use it A LOT in my code base (and having a native operator would be a huge improvement).

RolandPheasant:
Looks good to me. Is this your last PR for now?  

If so, would you mind bumping the version to 8.1 so we can release.  See [version.json#L2](https://github.com/reactivemarbles/DynamicData/blob/0be0d951a50112114ac82e7671de800b3a2e89cc/version.json#L2 ).

dwcullop:
I was working one more, but I'm not sure when I'll have time to finish it (still needs unit tests).

It adds operators for `IObservable<Optional<T>>` that are available for `Optional<T>` ... `Convert`, `ValueOr`, `SelectValues`, etc. so you can write code for Optional and Observable Optional that's essentially the same and is easier to read without all the extra `Select` and `Filter` calls in your Rx statement.  Syntactic Sugar, really.  I'm not even sure you'll want it, but I use them in my own projects, so I figured you and/or others might appreciate them as well.

```C#
IObservable<string>  FileNameObservable();
Optional<FileInfo> LoadFile(string file);
Optional<Image> LoadImage(FileInfo fi);

IObservable<Image> imageObservable = FileNameObservable().Convert(LoadFile).Convert(LoadImage).SelectValues();
                       
```

I got distracted by trying to optimize this PR (and failing miserably).  I thought I had a way to eliminate the two Transforms but wasn't able to make it work (yet).

I'll rev the version in this PR and save those for another day.

Also, if there are any open issues that you think are more urgent than others, feel free to tag me and I'll see if I can help out with them.  I was thinking about trying to address the `TransformManyAsync` request (#461) by cloning `TransformMany` and then adapting it to be async.

RolandPheasant (approved):
Yet another excellent addition to the library. Thanks

RolandPheasant:
@dwcullop all your changes have been deployed.  Thanks for an excellent contribution to Dynamic Data.

Are you on the slack channel?  It's very useful for discussions.  See https://reactiveui.net/slack, and look for the #dynamicdata sub channel.

image

dwcullop:
> @dwcullop all your changes have been deployed. Thanks for an excellent contribution to Dynamic Data.

Thanks.  I have a few more ideas and will contribute more as I am able to find the time.

> Are you on the slack channel? It's very useful for discussions. See https://reactiveui.net/slack, and look for the #dynamicdata sub channel.

Never used Slack before but I'll check it out.

PR #739, opened 2023-10-14
Feature:  EditDiff extension method for IObservable<Optional<T>>

dwcullop:
### Description
Similar to #738, this is an overload that operates on `IObservable<Optional<T>>` instead of `IObeservable<IEnumerable<T>>`.  It is useful because it provides yet another way to have child-change sets (and also works nicely with `MergeManyChangeSets` as seen below).

When the first Optional with a value is observed, an Add change is emitted.  When an empty Optional is observed, a Remove change is emitted for the previously Added value.  If a non-empty Optional is observed, the Keys are computed and compared to determine if either a single Update (keys are the same) or a Remove/Add pair of Updates (keys are different) should be emitted.

### Rationale
Since `Optional` is essentially a container with at most one value, the same behavior could be easily done with #738 and a helper function such as:

```C#
public static IEnumerable<T> AsEnumerable(this Optional<T> optional)
{
    if (optional.HasValue) yield return optional.Value;
}

public static IObservable<IChangeSet<TObject, TKey>> EditDiff<TObject, TKey>(this IObservable<Optional<TObject>> source, Func<TObject, TKey> keySelector, IEqualityComparer<TObject>? equalityComparer = null)
        where TObject : notnull
        where TKey : notnull =>
    source.Select(opt => opt.AsEnumerable()).EditDiff(keySelector, equalityComparer);
```

However, using such an approach will result in using an entire SourceCache that is essentially empty.  This specialized version is optimized around only having to deal with a single value at a time and is extremely light weight.

### Example
```C#
interface IPerson : INotifyPropertyChange
{
  int Id { get; }
  string Name { get; }
  Optional<IPerson> Spouse { get; }
  IReadOnlyList<IPerson> Children { get; }
}

IObservableCache<IPerson, int> peopleCache = GetPeopleCache();

// Treat all changes to all spouse properties as one observable changeset
// Creates "Add" events when a Spouse property is set with a value
// Creates "Remove" events when it is set without a value
IObservable<IChangeSet<IPerson, int>> spouseChangeSet = 
    peopleCache.Connect()
               .MergeManyChangeSets(person => person.WhenAnyValue(p => p.Spouse).EditDiff(p => p.Id));
```

### DIFF
--- a/src/DynamicData.Tests/Cache/EditDiffChangeSetOptionalFixture.cs
+++ b/src/DynamicData.Tests/Cache/EditDiffChangeSetOptionalFixture.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reactive.Linq;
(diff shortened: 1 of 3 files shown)


dwcullop:
I also have an implementation for the inverse of this operation that I am considering submitting because I find it to be very useful.

Something like:
`IObservable<Optional<T>> ToObservableOptional<T, TKey>(this IObservable<IChangeSet<T, TKey>> source, TKey key);`

It's similar to `Watch` except it converts to Optional (e.g., Remove events yield `Optional.None`).

RolandPheasant (approved):
Makes sense. Thanks

RolandPheasant:
@dwcullop annoyingly there's a merge conflict here.

dwcullop:
> @dwcullop annoyingly there's a merge conflict here.

That's on me for submitting so many (similar) PRs at once, but I had to get them out while I had the inspiration.  I'll resolve the conflicts, no worries.

dwcullop:
It worked out for the best anyway because the implementation was too complicated.  The whole point of this specialization is that it can be lighter weight (compared to using the IEnumerable version), but it still ended up using a bunch of Rx operators that weren't strictly needed.  

I updated the implementation to make it as thin as possible (only operator is Synchronize).  It's a lot less code and is easier to read as well.

PR #741, opened 2023-10-22
Feature:  Operators for IObservable<Optional<T>>

dwcullop:
## Description

This PR adds some extension methods to `IObservable<Optional<T>>` to enable the same fluent syntax for `IObservable<Optional<T>>` that exists for `Optional<T>`.  The operations are the identical to the single `Optional` versions except they apply to every value that comes through the observable event stream.

## Examples Usage

Some basic usage shown below:

```csharp
// Assume These
IObservable<Optional<string>> GetOptionalStringObservable();
Optional<int> ParseInt(string input) =>
    int.TryParse(input, out var result) ? Optional.Some(result) : Optional.None<int>();
Optional<int> GetFallbackValue();

IObservable<int> intObservable = GetOptionalStringObservable()
                                       .Convert(ParseInt)
                                       .OnHasValue(i => Console.WriteLine("Parsed: " + i.ToString()))
                                       .OrElse(GetFallbackValue);
```

Combine them with other Observable Optional functionality, such as that from #740.

```csharp
const int MagicId = 37;
var sourceCache = new SourceCache<Person, int>(p => p.Id);

using var cleanup1 = sourceCache.ToOptionalObservable(MagicId , initialOptionalWhenMissing: true)
                        .OnHasNoValue(() => Console.WriteLine("Person is not present"))
                        .Convert(p => p.FullName)
                        .SelectValues()
                        .Subscribe(name => Console.WriteLine("Person's Full Name: " + name));

using var cleanup2 = sourceCache.ToOptionalObservable(MagicId , initialOptionalWhenMissing: false)
                        .OnHasValue(p => Console.WriteLine("Person Added: " + p.FullName))
                        .ValueOrThrow(() => new Exception("Person was removed!"))
                        .Subscribe();
```

It can also be used with the extension methods added in #739 to create this advanced helper functions that syncs a value between an observable cache and a source cache:

(There are better ways to implement such a thing, this is just an example.)

```csharp
IDisposable SyncValue<TSource, TKey, TDest, TDestKey>(IObservableCache<TSource, TKey> source, ISourceCache<TDest, TDestKey> destination, TKey syncValueKey, Func<TSource, TDest> converter) =>
    source.Connect()
        .ToOptionalObservable(syncValueKey)  // Create Observable Optional
        .Convert(converter)                  // Convert to Destination Type
        .EditDiff(destination.KeySelector)   // Convert to Change Set
        .PopulateInto(destination);          // Merge into Destination
```

### DIFF
--- a/src/DynamicData.Tests/Kernal/OptionObservableFixture.cs
+++ b/src/DynamicData.Tests/Kernal/OptionObservableFixture.cs
@@ -1,7 +1,8 @@
 ﻿using System;
-using System.Globalization;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reactive.Linq;
 using DynamicData.Kernel;
-using DynamicData.Tests.Domain;
 
 using FluentAssertions;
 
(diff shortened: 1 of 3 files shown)


JakenVeina:
All tests are passing on the branch you're requesting to be bulled, but you're five commits out-of-date from the branch you're requesting a pull to.  Either run a merge from to `dwcullop:feature/observable-optional-extensions', or re-base `dwcullop:feature/observable-optional-extensions' into `reactivemarbles:main` to resolve whatever conflicts are causing the failures.

dwcullop:
Thanks. I will try that. I didn't realize. That's one problem with having some many PRs in flight.

glennawatson:
Codecov recently had an upgrade to its algorithm so suspect it is that. If you check it in it'll just be at that level.

RolandPheasant (approved):

JakenVeina:
> Thanks for looking at this.
> 
> I think it is up to date. 2 days ago I updated it. I tried to sync again just now and it already has all the commits. This branch is 5 commits ahead, but that's what I'm trying to merge.
> 
> Screenshot_20231025_003942_Chrome
> 
> The issue isn't a merge conflict. It is saying the coverage level of the unit tests drops below an acceptable level, which is odd because the one file I'm adding has 100% coverage.

Ah, my mistake, I misread it. Test coverage, not test results.

PR #743, opened 2023-10-23
Feature:  MergeChangeSets for Observable Cache ChangeSets

dwcullop:
## Description

This PR creates `MergeChangeSets` operators for Cache Observable ChangeSets.  This operator is very similar to `Merge` except it is ChangeSet aware and allows the consumer to specify how to handle duplicate keys.

It leverages the same changeset merging logic that is used by `MergeManyChangeSets` (except it doesn't require the changesets to be children of some other changeset).  This same logic will also be needed for the List -> Cache version of `MergeManyChangeSets`.

## Summary of Changes

1) Moves ChangeSet Merge logic out of `MergeManyChangeSets` class and into a pair of internal helper classes so it can be re-used by other operators:
      - `ChangeSetCache<TObject, TKey>`
      - `ChangeSetMergeTracker<TObject, TKey>`
1) Adds tracking for `OnComplete` to ChangeSetAggregator
1) Adds `Spy` operators to the test project (very helpful in debugging anything Rx related)
Includes special overloads for debugging observable changesets.
1) Lots of unit tests

## Examples

You can use it with a collection of changesets:
```cs
interface IItem
{
    public IObservable<IChangeSet<ISubItem, int>>  SubItemChanges { get; }
}

IEnumerable<IItem> itemCollection = GetTheItems();

// Merge all the change sets
using var mergedSubItems = itemCollection.Select(i => i.SubItemChanges).MergeChangeSets().AsObservableCache();
```

You can use it with Observables with ChangeSets:
```cs
IObservable<IItem> itemObservable = GetItemObservable();

// Merge all the change sets as Items come in
using var mergedSubItems = itemObservable.Select(i => i.SubItemChanges).MergeChangeSets().AsObservableCache();
```

You can call it with just two changesets:
```cs
IItem item1 = GetItem(1);
IItem item2 = GetItem(2);

using var bothSubItems = item1.SubItemChanges.MergeChangeSets(item2.SubItemChanges).AsObservableCache();
```

You can call it as an operator on ChangeSets with a collection of changesets:
```cs
IItem mainItem = GetItem(1);
IEnumerable<IItem> otherItems = GetOthers();

using var allSubItems = mainItem.SubItemChanges.MergeChangeSets(otherItems.Select(i => i.SubItemChanges)).AsObservableCache();
```

### An Advanced Example

The problem with merging multiple changesets is what to do if two changesets add a value with the same key.  All of the overloads `MergeChangeSets`, like `MergeManyChangeSets`, allow the consumer to provide a `IEqualityComparer` and/or a `IComparer` to control which of the values is emitted in the resulting merged changeset.

Different `IComparer` instances can be used to create different resulting ChangeSets based on the criteria.  An example of this is shown below.

```cs
interface IMarketPrice
{
    Guid MarketId { get; }
    string ItemId { get; }
    decimal Price { get; } 
    DateTimeOffset TimeStamp { get; }
}

class LowestPriceComparer : IComparer<MarketPrice>
{
    public int Compare(MarketPrice x, MarketPrice y) => x.Price.CompareTo(y);
}

class LatestPriceComparer : IComparer<MarketPrice>
{
    public int Compare(MarketPrice x, MarketPrice y) => y.TimeStamp.CompareTo(x);
}

IEnumerable<IObservableCache<MarketPrice, string>> priceCaches = GetTheMarketPrices();

// Create a cache with the lowest prices from all the markets
using var lowestPrices = priceCaches.Select(m => m.Connect()).MergeChangeSets(new LowestPriceComparer());

// Create a cache with the most recent prices across all markets
using var latestPrices = priceCaches.Select(m => m.Connect()).MergeChangeSets(new LatestPriceComparer());

```

As prices are added/removed/updated from all of the Price Cache instances, they will be reflected in the `lowestPrice` or `latestPrice` caches, but only if they're unique or they're the best choice according to the comparer.

If a "best choice" is removed from one of the sources, the result cache will see an "Update" event for the 2nd best choice (or a "Remove" event if there are no other choices).


dwcullop:
@RolandPheasant Would appreciate your feedback on this one. Functionality is very close to `Or` operator except it is optimized for just that case (not And or Xor, etc) and optionally supports extra parameters (comparer and/or equality comparer).

Do you think this should be a new operator? Or new extra overloads for Or?  Or something else?

It basically gives you the same functionality as MergeManyChangesets but doesn't require you to have child change sets. You can just have a collection of change sets (like with Or).

RolandPheasant:
>  Functionality is very close to `Or` operator except it is optimized for just that case (not And or Xor, etc) and optionally supports extra parameters (comparer and/or equality comparer).
> Do you think this should be a new operator? Or new extra overloads for Or? Or something else?

Whilst it's very close to Or, due to historic differences I am included to keep it as it is here.   We can consider later whether Or could be overloaded.

> > It basically gives you the same functionality as MergeManyChangesets but doesn't require you to have child change sets. You can just have a collection of change sets (like with Or).

That's great

dwcullop:
> > Functionality is very close to `Or` operator except it is optimized for just that case (not And or Xor, etc) and optionally supports extra parameters (comparer and/or equality comparer).
> > Do you think this should be a new operator? Or new extra overloads for Or? Or something else?
> 
> Whilst it's very close to Or, due to historic differences I am included to keep it as it is here. We can consider later whether Or could be overloaded.

That makes sense to me as well.  We don't want to change the behavior to which people are accustomed.  I just wanted to make sure you were okay with adding something that is similar to something existing. 

There is a difference between this and `Or` which might break things for some people:  `MergeChangeSets` will fire `OnCompleted` when all the source changesets have fired `OnCompleted` and fires `OnError` if any of the source changesets fire `OnError` (I wanted the same behavior as `Observable.Merge`).

dwcullop:
Hey Reactive @RolandPheasant! 

This one is close to ready to go (need to see the test coverage results).  The only open issue is that it does NOT fail if one of the child changesets fails (but the overloads that take IObservable<IObservable<...>> will fail if the outer observable fails).

I asked the question on Slack to see if what people would expect. We will see if they want it to fail on a child failure (like `Observable.Merge` does), or if they wouldn't expect it to fail (like `MergeManyChangeSets` doesn't).

RolandPheasant:
> There is a difference between this and Or which might break things for some people: MergeChangeSets fires OnCompleted when all the source changesets have fired OnCompleted and fires OnError if any of the source changesets fire OnError (I wanted the same behavior as Observable.Merge).

That's probably what I should have done with Or in the first place.

src/DynamicData.Tests/Cache/MergeChangeSetsFixture.cs
```diff
@@ -0,0 +1,1056 @@
...
+using System.Reactive.Linq;
+using DynamicData.Kernel;
+using DynamicData.Tests.Utilities;
+using FluentAssertions;
+using Microsoft.Reactive.Testing;
+using Xunit;
+
+namespace DynamicData.Tests.Cache;
+
+public sealed class MergeChangeSetsFixture : IDisposable
```
RolandPheasant, line 15: Very impressive level of testing

src/DynamicData.Tests/Utilities/ObservableSpy.cs
```diff
@@ -0,0 +1,183 @@
...
+    /// <typeparam name="T">The type of the Observable.</typeparam>
+    /// <param name="source">The source IObservable to be Spied on.</param>
+    /// <param name="logger">The logger instance to use for logging.</param>
+    /// <param name="infoText">Optional text to include with each log message.</param>
+    /// <param name="formatter">Optional Value transformer to control how the observed values are logged.</param>
+    /// <param name="showSubs">Indicates whether or not subscription related messages should be emitted.</param>
+    /// <param name="showTimestamps">Indicates whether or not timestamps should be prepended to messages.</param>
+    /// <returns>An IObservable{T} with the Spy events included.</returns>
+    /// <remarks>Adapted from https://stackoverflow.com/q/20220755/.</remarks>
+    public static IObservable<T> Spy<T>(this IObservable<T> source, string? infoText = null, Action<string>? logger = null,
```
RolandPheasant, line 27: This is super cool

RolandPheasant (approved):
This is a major contribution. Thanks

src/DynamicData.Tests/Utilities/ObservableSpy.cs
dwcullop: It is amazing. I wish I could take credit, but it's not my idea... I just fleshed it out a bit.  It is game-changing in terms of debugging Rx.  I did add the ChangeSet-specific overloads that allow it to peer into the individual changes though.  This has saved me a ton of time and hassle and helps get you right to the root of an issue.  I really needed its help for a couple of the test-cases and so I ported it over from my personal library.

I think you should consider promoting it to the main library instead of the test library, but I didn't want to be presumptuous.  Also, I don't think the interface is quite ready for mainstream.  It's fine for debugging, but for production, people will want to integrate their own loggers and not have to specify the logging facility each time.  Not exactly sure what that would look like, but it was definitely out of scope for this PR, so I included it as-is.

src/DynamicData.Tests/Cache/MergeChangeSetsFixture.cs
dwcullop: To be fair, I copied over most of the tests from the `MergeManyChangeSets` fixture and then tweaked them.  Saving all that time allowed me to just add a few more, which is how found one corner case that I had missed in `MergeManyChangeSets`, but now that the implementation is shared, fixing one fixed them both.

But it was a real corner-corner case that required using a custom EqualityComparer and custom Comparer together.

The issue was:
1) Values are added (merged results sees the add)
2) Values are updated to values that are equal according to the EqualityComparer (merge results sees nothing)
3) Values are refreshed (merged results sees refresh events for objects it has never observed) and they still are equal to the originally added (and updated) values

I changed it so that now, at 3, it won't see the refresh events either because something shouldn't be refreshed if it hasn't been seen.

I'm pretty sure that's the right behavior.  If the refresh had caused the updated value to be the new best value, then it would have observed the Update at that point.  Tell me if you think that's the wrong behavior.

PR #744, opened 2023-10-25
Feature:  List MergeManyChangeSets for Cache ChangeSets

dwcullop:
### Description

This PR adds `MergeManyChangeSets` for List ChangeSets where the child changeset is a Cache ChangeSet.  In other words, it allows you to create a Cache ChangeSet from a child cache of items in a Source List.

Functionality is identical to that in #743 except the input changeset type is list instead of cache.

### Dependencies

What was once a single PR is now three.  While the other two are independent, this PR requires them both to work (or else the tests won't pass and/or it might not build).
- #742
- #743 (Leverages helper class for the Merging logic.  That class is temporarily also included in this PR for illustrative purposes but will be removed after the other PR completes)


RolandPheasant:
@dwcullop I just pushed a commit to this PR  which adds some replace tests.   See https://github.com/reactivemarbles/DynamicData/pull/744/commits/53feeeb8c20e69af142717d5d6f2f52efe3ea04a

RolandPheasant:
@dwcullop I've work it out.

If you change the following:

```cs
var sourceListOfCaches = _source
    .WhereReasonsAreNot(ListChangeReason.Moved, ListChangeReason.Refresh)
    .Transform(obj => new ChangeSetCache<TDestination, TDestinationKey>(_changeSetSelector(obj)))
    .Synchronize(locker)
    .AsObservableList();
```

to 

```cs
var sourceListOfCaches = _source
    .Transform(obj => new ChangeSetCache<TDestination, TDestinationKey>(_changeSetSelector(obj)))
    .Synchronize(locker)
    .AsObservableList();
```

The problem goes away.  The reason is WhereReasonsAreNot removes the index from all changes because the index is unreliable when reasons are removed.  I think I used to rely on that operator internally but due to such issues I stopped using it.

That said, we should always mitigate against such things in the operators (in case someone is using WhereReasonsAreNot).  So for safety the replace method in Transformer should become:

```cs
    {
        var change = item.Item;

        if (change.CurrentIndex == -1 || change.PreviousIndex == -1)
        {
            // Find the original, with it's corresponding index
            var previous = transformed.First(x => x.Source.Equals(change.Previous.Value));
            var index = transformed.IndexOf(previous);
            if (index == -1)
            {
                throw new UnspecifiedIndexException($"Cannot find index of {change.Previous.Value}");
            }

            transformed[index] = _containerFactory(change.Current, previous.Destination, index);
        }
        else
        {
            Optional<TDestination> previous = transformed[change.PreviousIndex].Destination;
            if (change.CurrentIndex == change.PreviousIndex)
            {
                transformed[change.CurrentIndex] = _containerFactory(change.Current, previous, change.CurrentIndex);
            }
            else
            {
                transformed.RemoveAt(change.PreviousIndex);
                transformed.Insert(change.CurrentIndex, _containerFactory(change.Current, Optional<TDestination>.None, change.CurrentIndex));
            }
        }

        break;
    }
```

dwcullop:
> @dwcullop I've work it out.

Hurray!  That resolved the issue!  PR has been updated and I'll mark it as ready!

src/DynamicData.Tests/List/MergeManyCacheChangeSetsFixture.cs
```diff
@@ -0,0 +1,737 @@
...
+            DisposeMarkets();
+        }
+
+        // then
+        hasSourceSequenceCompleted.Should().Be(completeSource);
+        hasMergedSequenceCompleted.Should().Be(completeSource && completeChildren);
+    }
+
+    [Fact]
+    public void MergedObservableWillFailIfSourceFails()
```
JakenVeina, line 692: Should we also test that errors from child streams propagate out?

src/DynamicData.Tests/List/MergeManyCacheChangeSetsFixture.cs
```diff
@@ -0,0 +1,737 @@
...
+        using var cleanup = _marketList.Connect().Concat(throwObservable)
+                            .MergeManyChangeSets(m => m.LatestPrices).Subscribe(_ => { }, err => receivedError = err);
+
+        // when
+        DisposeMarkets();
+
+        // then
+        receivedError.Should().Be(expectedError);
+    }
+
```
JakenVeina, line 710: Following off of the bug I fixed yesterday, could/should we add a test to cover downstream-teardown (I.E. unsubscription) behavior? Internally, this means cleaning up all the outstanding child stream subscriptions, right?

src/DynamicData/List/Internal/MergeManyCacheChangeSets.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                var shared = sourceListOfCaches.Connect().Publish();
+
+                // this is manages all of the changes
+                var changeTracker = new ChangeSetMergeTracker<TDestination, TDestinationKey>(() => sourceListOfCaches.Items.ToArray(), _comparer, _equalityComparer);
+
+                // when a source item is removed, all of its sub-items need to be removed
+                var removedItems = shared
+                    .OnItemRemoved(mc => changeTracker.RemoveItems(mc.Cache.KeyValues, observer))
+                    .Subscribe();
```
JakenVeina, line 56: `.Subscribe()` calls here should probably all be `.SubscribeSafe()` as its documentation recommends its usage within operators. This literally just `try`/`catch`es the `.Subscribe()` call and feeds anything it catches to `onError`. This is important for operators, because it prevents things like `.OnNext()` or `.OnCompleted()` from throwing, when operators that orchestrate different streams together (like, say `.Concat()`) are involved. 

I was also going to suggest passing `observer.OnError` to this subscription, but I'm less sold on that, after thinking it through. Since the source here is shared with `.Publish()` the only errors that can occur on this subscription (that aren't already going to be propagated from the "main" subscription) are from `.OnItemRemoved()`, which doesn't involve any user code. If there's a bug on our part, that's going to throw out of `.Subscribe()` or `.OnNext()` or `.OnComplete()` or whatever the user was doing that triggered the removal operation. But maybe that's actually preferable, since it's not the user's fault?

src/DynamicData.Tests/List/MergeManyCacheChangeSetsFixture.cs
RolandPheasant: We probably should do.  I wonder whether we should raise an issue and gradually trawl through the system.

JakenVeina: I wouldn't expect that there's a systemic issue with not handling downstream-teardown, that was just a defect I introduced a few days ago. But it is a behavior worth testing, wherever we have resources that need cleanup.

Perhaps instead what we need is a guidelines document for testing operators.

RolandPheasant (approved):
Let's get this in, and @JakenVeina we should review places where we can improve usages in Subscription (safe or not) here and throughout the system in a subsequent PR

src/DynamicData.Tests/List/MergeManyCacheChangeSetsFixture.cs
dwcullop: I decidedly did not want errors from child streams to tear down the whole thing.  I could be convinced that is the wrong behavior, but my thinking was that an error on the child stream shouldn't disrupt all of the other child streams or the main stream.

src/DynamicData/List/Internal/MergeManyCacheChangeSets.cs
dwcullop: Since it's a shared Observable, I don't think they all need to be SubscribeSafe.  This one in particular is only used internally, so if it throws something is really wrong.

Maybe the one below should be SubscribeSafe so that the consumer can detect a bad subscription via their OnError.  That one also passes in `observer.OnError`.

But I do like Roland's suggestion of uniformly addressing these as a single PR.

PR #750, opened 2023-11-05
Feature:  MergeManyChangeSets with Parent Item Comparison

dwcullop:
## Description

This PR adds overloads to the Cache verion of `MergeManyChangeSets` that allow an additional `IComparer` parameter that operates on the **source** type so that priority of the resulting changeset can be based on factors in the source item, and not just child items.

Two `IComparer` instance can be provided so that child-based comparisons can be done as a fallback when the source comparison evaluates as the same.

## Examples

Say there's a mobile app that aggregates movie review web sites and has these basic interfaces:
```cs
interface IMovieReview
{
    Guid MovieId { get; }
    Guid CriticId { get; }
    double Rating { get; }
    DateTimeOffset DatePublished { get; }
    string Title { get; }
    Uri Link { get; }
}

interface IMovieCritic : INotifyPropertyChanged
{
    Guid Id { get; }
    string Name { get; }
    double Popularity { get; }
    IObservableCache<IMovieReview, Guid> Reviews { get; }
}

interface IMovieWebService
{
    IObservableCache<IMovieCritic, Guid> AllCritics { get; }
}

public static class MovieWebApi
{
    public IMovieWebService Service => GetTheWebServiceSomehow();
}

```

And the goal is to create an observable changeset that has a review for each movie, but when the movie has been reviewed by multiple critics, it should emit the review from the most popular critic.  This is problematic because the Internet is fickle and popularity can change.  How can you do such a thing?  DynamicData and Rx will come to the rescue.

With this PR, all such a changeset can be created with just a few lines of code:

```cs

// Need a comparer for Critics
class CriticComparer : IComparer<IMovieCritic>
{
    // Not a production-quality implementation, but you get the idea
    public int Compare(IMovieCritic x, IMovieCritic y) => y.Popularity.CompareTo(x.Popularity);
}

// Create the observable cache of movie reviews for every movie using the review for the most popular critic to have rated that movie
using var movieReviewCache = MovieWebApi.Service.AllCritics.Connect()
                                .AutoRefresh(critic => critic.Popularity)
                                .MergeManyChangeSets((critic, id) => critic.Reviews.Connect(), new CriticComparer())
                                .AsObservableCache();

```

That's all there is to it!  And now `movieReviewCache` will always contain the review of each movie from the most popular critic to have reviewed that movie.  For example, it will exhibit some of the following behaviors:
1) If a new critic is added, any movies they've reviewed will be added (unless they've already been reviewed by a more popular reviewer)
2) If a critic is removed, any movies only reviewed by them will be removed.  Other movies will be `Updated` to the review from next most-popular reviewer.
3) If any critic is the first to review a movie, it will be added.  When another critic reviews the same movie, their review may or may not replace the first one, depending on which reviewer is the most popular.
4) If the most popular critic to review a given movie removes their review, it will be updated to the next most popular critic to have reviewed that movie.  If no one else has a review for that movie, it will be removed from the cache instead.
5) If the Popularity of the Reviewer changes, reviews will be added/updated depending on how many more popular reviewers have reviewed the same movies such that the result cache always contains the review from the most popular reviewer.

### Another Example

At this point, you might be wondering "What if two Critics have the same popularity and rate the same movie?  I need it to always include the Best/First/Most Recent/etc review in those cases."

Well, there's an overload for that as well.  You just need a comparer for the child objects that will be used as a fallback when two parents compare to be the same.

```cs

// Need a comparer for Reviews (sorts earliest publishing dates first)
class FirstPublishedReviewComparer : IComparer<IMovieReview>
{
    // Not a production-quality implementation, but you get the idea
    public int Compare(IMovieReview x, IMovieReview y) => x.DatePublished.CompareTo(y.DatePublished);
}

// Create the observable cache of movie reviews for every movie using the review for the most popular critic to have rated that movie (going by publish date when multiple reviewers are tied)
using var movieReviewCache = MovieWebApi.Service.AllCritics.Connect()
                                .AutoRefresh(critic => critic.Popularity)
                                .MergeManyChangeSets((critic, id) => critic.Reviews.Connect(), new CriticComparer(), new FirstPublishedReviewComparer())
                                .AsObservableCache();

```

## Testing Improvements

- Moves the data classes `Market` and `MarketPrice` to be part of `Domain` namespace promote re-use
- Adds extension methods for creating tests

### DIFF
--- a/src/DynamicData/Cache/Internal/ChangeSetMergeTracker.cs
+++ b/src/DynamicData/Cache/Internal/ChangeSetMergeTracker.cs
@@ -185,33 +185,19 @@ internal class ChangeSetMergeTracker<TObject, TKey>
     {
         var cached = _resultCache.Lookup(key);
 
-        // Only proceed if the key has a current value
-        if (cached.HasValue)
+        // Received a refresh change for a key that hasn't been seen yet
+        // Nothing can be done, so ignore it
+        if (!cached.HasValue)
         {
-            // If the refreshed value is the current one
-            if (ReferenceEquals(cached.Value, item))
-            {
-                // When using a compare and the current value has changed, so do a full search for
-                // the best value to make sure the current choice is still the best choice
-                if ((_comparer is not null) && UpdateToBestValue(sources, key, cached))
-                {
-                    // A new value was choosen, so there's nothing left to do
-                    return;
-                }
+            return;
+        }
 
-                // The current one is still the best choice and it was refreshed, so
-                // emit the Refresh downstream so consumers will see it.
-                _resultCache.Refresh(key);
-            }
-            else
-            {
-                // If the current value isn't being refreshed and using a comparer,
-                // check if the refreshed item is now a better choice
-                if ((_comparer is not null) && ShouldReplace(item, cached.Value))
-                {
-                    _resultCache.AddOrUpdate(item, key);
-                }
-            }
+        // In the sorting case, a refresh requires doing a full update because any change could alter what the best value is
+        // If we don't care about sorting OR if we do care, but re-selecting the best value didn't change anything
+        // AND the current value is the exact one being refreshed, then emit the refresh downstream
+        if (((_comparer is null) || !UpdateToBestValue(sources, key, cached)) && ReferenceEquals(cached.Value, item))
+        {
+            _resultCache.Refresh(key);
         }
     }
 
--- a/src/DynamicData/Cache/Internal/ChangeSetMergeTracker.cs
+++ b/src/DynamicData/Cache/Internal/ChangeSetMergeTracker.cs
@@ -48,6 +48,31 @@ internal class ChangeSetMergeTracker<TObject, TKey>
         EmitChanges(observer);
     }
 
+    public void RefreshItems(IEnumerable<TKey> keys, IObserver<IChangeSet<TObject, TKey>> observer)
+    {
+        var sourceCaches = _selectCaches().ToArray();
+
+        // Update the Published Value for each item being removed
+        if (keys is IList<TKey> list)
+        {
+            // zero allocation enumerator
+            foreach (var item in EnumerableIList.Create(list))
+            {
+                ForceEvaluate(sourceCaches, item.Value, item.Key);
+            }
+        }
+        else
+        {
+            foreach (var item in keys)
+            {
+                ForceEvaluate(sourceCaches, item.Key);
+            }
+            UpdateToBestValue(sources, key, cached)
+        }
+
+        EmitChanges(observer);
+    }
+
     public void ProcessChangeSet(IChangeSet<TObject, TKey> changes, IObserver<IChangeSet<TObject, TKey>> observer)
     {
         var sourceCaches = _selectCaches().ToArray();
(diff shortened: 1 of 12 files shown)


dwcullop:
@RolandPheasant I think this one is now ready to go.  Please take a look at your leisure and let me know what you think.

SuperJMN:
ExplodindoBoomExplosãoCabeçaexplodindoGIF

src/DynamicData/Cache/Internal/ChangeSetMergeTracker.cs
```diff
@@ -48,6 +48,30 @@ public void RemoveItems(IEnumerable<KeyValuePair<TKey, TObject>> items, IObserve
...
     }
 
+    public void RefreshItems(IEnumerable<TKey> keys, IObserver<IChangeSet<TObject, TKey>> observer)
+    {
+        var sourceCaches = _selectCaches().ToArray();
+
+        // Update the Published Value for each item being removed
+        if (keys is IList<TKey> list)
+        {
+            // zero allocation enumerator
```
RolandPheasant, line 58: I included the zero allocation enumerator a long time ago after I realised that implementing an IChangeSet allocated when it is iterated.

Later I re-implemented the change sets to inherit List<T> rather than to delegate it.  This hacky way of doing things is a throw back to the earlier way of doing things,  Ideally we should remove IChangeSet altogether but in practice, it would break every person on this earth who has included a custom Dynamic Data extension.

At an even later point, I wondered whether there is a single person in this universe that ever implemented a custom IChangeSet and thought it to be unlikely.  To test this hypothesis I introduced and internal direct cast to List<T>, see [CacheChangeSetEx.cs](https://github.com/reactivemarbles/DynamicData/blob/main/src/DynamicData/Cache/CacheChangeSetEx.cs) and use it internally in several key places.

This was in 2020 and there has not been a single complaint about any code being broken by this cast.  I think we can safely just cast to the concrete type throughout the code base.  Just to be thorough, I think we should use the ToConcreteType() extension but change the implementation to this:

```cs
    public static ChangeSet<TObject, TKey> ToConcreteType<TObject, TKey>(this IChangeSet<TObject, TKey> changeSet)
        where TObject : notnull
        where TKey : notnull
    {
        if (changeSet is ChangeSet<TObject, TKey> concreteList)
        {
            return concreteList;
        }

        throw new NotSupportedException("Dynamic Data does not support a custom implementation of IChangeSet");
    }
```

RolandPheasant (approved):
It's optional whether to action the zero allocation thing or not.

Maybe that would be best done as a distinct task / pr/

src/DynamicData/Cache/Internal/ChangeSetMergeTracker.cs
dwcullop: I am using the `ToConcreteType` extension when enumerating the ChangeSets.  I'm not sure if I'm using the thing for `EnumerableIList.Create` correctly because I just cloned from an example.

I'm happy to change the implementation of `ToConcreteType` but it should probably be another PR because it will affect so many things not related to this PR, and if it needs to be rolled back, it's easier if it's isolated.

PR #751, opened 2023-11-07
ToConcreteType Update

dwcullop:
## Description

This PR continues to effort to replace `IChangeSet<TObject, TKey>` with `ChangeSet<TObject, TKey>` in order to avoid allocations that can take place when enumerating with the former instead of the latter.

- Updates `ToConcreteType` to throw an exception if a non-ChangeSet implementation of IChangeSet is encountered
- Changes anywhere enumeration would happen using IChangeSet to invoke `ToConcreteType` so that it is done with `ChangeSet` instead.
- When possible, `IChangeSet` was completely replaced with `ChangeSet` in order to ensure more consistent use and reduce the number of places that `ToConcreteType` needs to be invoked explicitly

### DIFF
--- a/src/DynamicData/Cache/Internal/DisposeMany.cs
+++ b/src/DynamicData/Cache/Internal/DisposeMany.cs
@@ -49,7 +49,7 @@ internal sealed class DisposeMany<TObject, TKey>
 
     private void RegisterForRemoval(IChangeSet<TObject, TKey> changes, Cache<TObject, TKey> cache)
     {
-        changes.ForEach(
+        changes.ToConcreteType().ForEach(
             change =>
             {
                 switch (change.Reason)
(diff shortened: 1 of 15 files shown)


src/DynamicData/Cache/Internal/DisposeMany.cs
```diff
@@ -49,7 +49,7 @@ public DisposeMany(IObservable<IChangeSet<TObject, TKey>> source, Action<TObject
 
     private void RegisterForRemoval(IChangeSet<TObject, TKey> changes, Cache<TObject, TKey> cache)
     {
-        changes.ForEach(
+        changes.ToConcreteType().ForEach(
```
RolandPheasant, line 52: I think this should be changed to the traditional foreach() loop instead of the ForEach extension.

I wrote the ForEach extension, but it was misguided as it also allocates.  However, looking at the usages:

image

That should be changed in a distinct PR

RolandPheasant (approved):
Thanks.

src/DynamicData/Cache/Internal/DisposeMany.cs
dwcullop: Yeah, I can do a distinct PR for that.  I'd like to understand more about why it allocates because that is news to me.  As a fan of `ForEach`, maybe it can be updated to avoid the allocation instead of needing to be removed.

PR #759, opened 2023-11-19
Fix API Checks, Add paragraphs to comments.

ChrisPulman:
API check now fails when doesn't match.


src/DynamicData.Tests/API/ApiApprovalBase.cs
```diff
@@ -15,29 +21,125 @@ namespace DynamicData.APITests
...
         /// <summary>
         /// Check Approval.
         /// </summary>
         /// <param name="assembly">The assembly.</param>
+        /// <param name="memberName">Name of the member.</param>
         /// <param name="filePath">The file path.</param>
-        /// <returns>A <see cref="Task"/> representing the result of the asynchronous operation.</returns>
-        protected static Task CheckApproval(Assembly assembly, [CallerFilePath] string? filePath = null)
+        /// <returns>
+        /// A <see cref="Task" /> representing the result of the asynchronous operation.
```
JakenVeina, line 33: This doc is no longer valid.

ChrisPulman: I have found the issue with the original code and updated, I was missing a Task and an await, many apologies.

glennawatson (approved):

src/DynamicData.Tests/API/ApiExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <summary>
+    /// Checks to make sure the API is approved.
+    /// </summary>
+    /// <param name="assembly">The assembly that is being checked.</param>
+    /// <param name="namespaces">The namespaces.</param>
+    /// <param name="filePath">The caller file path.</param>
+    /// <returns>
+    /// A Task.
+    /// </returns>
+    public static async Task CheckApproval(this Assembly assembly, string[] namespaces, [CallerFilePath] string filePath = "")
```
JakenVeina, line 30: Why an extension method, as opposed to just putting it directly in the test? `Verifier.Verify()` has its own `[CallerFilePath]` annotation, so calling directly should still work.

src/DynamicData.Tests/DynamicData.Tests.csproj
```diff
@@ -1,10 +1,10 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
   <PropertyGroup>
     <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
-    <NoWarn>$(NoWarn);CS0618;CA1801;CA1063;CS8767;CS8602; CS8618</NoWarn>
+    <NoWarn>$(NoWarn);CS0618;CA1801;CA1063;CS8767;CS8602;CS8618;IDE1006</NoWarn>
```
JakenVeina, line 4: Was this an intentional addition? Cloning the branch and removing it, I don't actually see this popping up anywhere.

src/DynamicData.Tests/API/ApiExtensions.cs
glennawatson: It was brought over from rxui where multiple projects and tests are run against different frameworks etc. less of an issue with dynamicdata tho.

PR #762, opened 2023-11-22
Index Safety for List Tranform Operator

dwcullop:
## Description

Adds the changes suggested by @RolandPheasant in #744 to avoid throwing an exception when Transform is used in conjunction with other operators such as (`WhereReasonsAre*`) that remove the Index from the Changes.

Transform will now check for missing indexes and do a lookup to find the correct index values to use for that situation.

Also adds two unit tests to the Fixture for List Transform.  Before the other changes, one test would pass and the other would fail.  With the changes, they both pass.

### DIFF
--- a/src/DynamicData/List/Internal/Transformer.cs
+++ b/src/DynamicData/List/Internal/Transformer.cs
@@ -82,14 +82,26 @@ internal sealed class Transformer<TSource, TDestination>
                 case ListChangeReason.Refresh:
                     {
                         var change = item.Item;
+                        var index = change.CurrentIndex;
+                        if (index < 0)
+                        {
+                            // Find the corresponding index
+                            var current = transformed.FirstOrDefault(x => x.Source.Equals(change.Current));
+                            index = transformed.IndexOf(current) switch
+                            {
+                                int i when i >= 0 => i,
+                                _ => throw new UnspecifiedIndexException($"Cannot find index of {change.Current}")
+                            };
+                        }
+
                         if (_transformOnRefresh)
                         {
-                            Optional<TDestination> previous = transformed[change.CurrentIndex].Destination;
-                            transformed[change.CurrentIndex] = _containerFactory(change.Current, previous, change.CurrentIndex);
+                            Optional<TDestination> previous = transformed[index].Destination;
+                            transformed[index] = _containerFactory(change.Current, previous, index);
                         }
                         else
                         {
-                            transformed.RefreshAt(change.CurrentIndex);
+                            transformed.RefreshAt(index);
                         }
 
                         break;
--- a/src/DynamicData/List/Internal/Transformer.cs
+++ b/src/DynamicData/List/Internal/Transformer.cs
@@ -87,10 +87,10 @@ internal sealed class Transformer<TSource, TDestination>
                         {
                             // Find the corresponding index
                             var current = transformed.FirstOrDefault(x => x.Source.Equals(change.Current));
-                            index = transformed.IndexOf(current) switch
+                            index = current switch
                             {
-                                int i when i >= 0 => i,
-                                _ => throw new UnspecifiedIndexException($"Cannot find index of {change.Current}")
+                                TransformedItemContainer tic when transformed.IndexOf(tic) is int i && (i >= 0) => i,
+                                _ => throw new UnspecifiedIndexException($"Cannot find index of {change.Current}"),
                             };
                         }
 
--- a/src/DynamicData/List/Internal/Transformer.cs
+++ b/src/DynamicData/List/Internal/Transformer.cs
@@ -89,7 +89,7 @@ internal sealed class Transformer<TSource, TDestination>
                             var current = transformed.FirstOrDefault(x => x.Source.Equals(change.Current));
                             index = current switch
                             {
-                                TransformedItemContainer tic when transformed.IndexOf(tic) is int i && (i >= 0) => i,
+                                TransformedItemContainer tic when transformed.IndexOf(tic) is int i and (>= 0) => i,
                                 _ => throw new UnspecifiedIndexException($"Cannot find index of {change.Current}"),
                             };
                         }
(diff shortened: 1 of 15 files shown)


JakenVeina:
This feels kinda like a band-aid, to me. If there's operators that aren't preserving indexes that should be, we should fix them. If there's operators that aren't preserving indexes because they can't, then how is `.Transform()` able to reliably re-acquire that lost info?

I missed whatever conversation spawned this, what's the usage-scenario here?

RolandPheasant:
> This feels kinda like a band-aid, to me. If there's operators that aren't preserving indexes that should be, we should fix them. If there's operators that aren't preserving indexes because they can't, then how is `.Transform()` able to reliably re-acquire that lost info?
> 
> I missed whatever conversation spawned this, what's the usage-scenario here?

Any stateless operator, or any of the operators `WhereReasonsAre*` operators remove the index, or at least cannot guarantee index integrity.  For example 

```cs
myChangeSetObservable.WhereReasonsAreNot(Remove)
```
removes the remove events, and additionally does not maintain state. Therefore when a subsequent Add arrives, the index which the Add carries will be incorrect and could cause an index out of range exception.  Thus, the WhereReasonsAre*` actually remove the index.  It's a necessary evil which unfortunately have to be handled by other operators.

JakenVeina:
Couldn't any of those filtering operators, in theory, track all the indices and issue `Move` or `Update` changes to compensate? Not that that's necessarily better.

Even so, I'm still not quite following how a downstream operator can recreate index information, if it was deemed impossible to determine by the upstream operator.

dwcullop:
> We could adjust the code to not remove the index. For context, I have used `WhereReasonsAreNot(Refresh)` and it's overload `SuppressRefresh()` quite a lot over the years.
> 
> @dwcullop would you be able to address this too? Or if you don't have the time we could to it later?

I will give it a shot.  It doesn't seem like too much work.  I'll get to it when I can.

dwcullop:
> I missed whatever conversation spawned this, what's the usage-scenario here?

When I was making the List version of `MergeManyChangeSets`, I combined the `WhenReasonAreNot` and the `Transform` operators which caused `Transform` to throw during a `Replace` test case because it assumed the indexes were valid and they weren't (because they had been removed).

If it happened to me, it could happen to anyone, and the library should not throw just by combining valid operators.

The other operators, including `Transform`, usually handle the case of the indexes being -1, but the `Replace` change reason was overlooked.  This PR just closes that gap by doing a lookup to find the index because the value in the Change is invalid.  I believe this is how all of the operators handle the same situation.

src/DynamicData/List/Internal/Transformer.cs
```diff
@@ -98,15 +98,31 @@ private void Transform(ChangeAwareList<TransformedItemContainer> transformed, IC
                 case ListChangeReason.Replace:
                     {
                         var change = item.Item;
-                        Optional<TDestination> previous = transformed[change.PreviousIndex].Destination;
-                        if (change.CurrentIndex == change.PreviousIndex)
+
+                        if (change.CurrentIndex == -1 || change.PreviousIndex == -1)
```
dwcullop, line 102: @RolandPheasant  Would it ever happen that the CurrentIndex is valid but the PreviousIndex is missing?  If so, then I might need to tweak this a bit so that it follows the code path at the bottom that allows the currentIndex to be different than the previous index.

I'm not sure that's a valid concern though.

RolandPheasant: It's possible, but if they are out of sync then probably best to place it safe and assume their both in a bad state.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -2254,6 +2254,13 @@ public static IObservable<IChangeSet<T>> WhereReasonsAreNot<T>(this IObservable<
             throw new ArgumentException("Must enter at least 1 reason", nameof(reasons));
         }
 
+        // Optimization that avoids removing the indexes if only refreshes are removed
+        if (reasons.Length == 1 && reasons[0] == ListChangeReason.Refresh)
+        {
+            // Caller probably meant to call this instead, so help them out.
+            return source.SuppressRefresh().NotEmpty();
```
RolandPheasant, line 2261: Actually, I just look up SuppressRefresh() and it overloads WhereReasonsAreNot().  This means we'll have to use some specialized handling in this method.

RolandPheasant (changes requested):
Alas, using SupressRefresh will not work.  See my comment.

src/DynamicData/List/ObservableListEx.cs
dwcullop: The Cache version does that, but not the List version which explicitly strips out the Refresh events.

If you'd prefer, I can change the SuppressRefresh to match the Cache version and move the old code to be inside of `WhereReasonsAreNot`.  Either way, the logic will appear in only one place.

I'm not sure which approach is better, but maybe it would be good to align the List and Cache versions.

dwcullop: I like having the implementation in `SuppressRefresh` better because then it avoids the allocation of the ListChangeReason[] and then checking it for the one single value.  It's like the more optimized version.  Doing it the other way means there is no optimized version, just an easier way to call it.

But it does mean not bringing the List/Cache versions back together...  But I think that's okay because they don't need to be aligned.

dwcullop: Okay, I changed my mind.  I put the logic for both cases (Refresh only and everything else) in the `WhereReasonsAreNot` operator.  Yes, there's no way to call it without the array, but all of the logic is in one place instead of two.

Plus, that enabled me to change the `SuppressRefresh` operator to invoke `WhereReasonsAreNot` so that it is the same as the Cache version.

Both of these simplify the implementation which is a good thing for maintainers.

RolandPheasant:
The Codecov things makes it super annoying to actually follow the changes.

RolandPheasant:
> The Codecov things makes it super annoying to actually follow the changes.

@glennawatson is there any way to make it less aggressive?. There's a huge number of false messages.

RolandPheasant (approved):

RolandPheasant:
@ChrisPulman thanks for the commit. Much appreciated

dwcullop:
> @ChrisPulman thanks for the commit. Much appreciated

Yeah, that was a fast turn around.

PR #776, opened 2023-12-03
Binding options - allow system wide options

RolandPheasant:
Allow system wide binding option defaults + add capability to suppress reset for first time load,

Currently binding options like reset threshold can be specified at the bind operator.  For example, 

```cs
myObservableChangeSet.Bind(var out myCollection, resetThreshold:100);
```
The limitation here is you may want to specify system wide defaults, which will eliminate the need to specify the threshold on every bind instance. To solve this problem, static system wide options have been introduced.

```cs
DynamicDataOptions.Binding = new BindingOptions(100)
```
which means

```cs
myObservableChangeSet.Bind(var out myCollection);
```
will have a default threshold of 100, without the need to specify it.

There are additional options on BindingObject:

**UseReplaceForUpdates**  - some control suites / platforms cannot handle replace events on the observable collection. If not true, replace and add new is used, 
**ResetOnFirstTimeLoad**  - similar to above, some control suites cannot handle a reset event so there's an option to prevent reset on first time load.

These options can be applied globally or at an individual bind level:

```cs
myObservableChangeSet.Bind(var out myCollection, new BindingOptions(100, ResetOnFirstTimeLoad : false);

// or to prevent reset completely and not just for the first time load
myObservableChangeSet.Bind(var out myCollection, new BindingOptions(100, BindingOptions.NeverFireReset());
```
Tests to follow, and there should be no breaking changes.

### DIFF
--- a/src/DynamicData/Cache/ObservableCacheEx.cs
+++ b/src/DynamicData/Cache/ObservableCacheEx.cs
@@ -596,14 +596,42 @@ public static class ObservableCacheEx
     /// <param name="refreshThreshold">The number of changes before a reset notification is triggered.</param>
     /// <returns>An observable which will emit change sets.</returns>
     /// <exception cref="System.ArgumentNullException">source.</exception>
-    public static IObservable<IChangeSet<TObject, TKey>> Bind<TObject, TKey>(this IObservable<IChangeSet<TObject, TKey>> source, IObservableCollection<TObject> destination, int refreshThreshold = 25)
+    public static IObservable<IChangeSet<TObject, TKey>> Bind<TObject, TKey>(this IObservable<IChangeSet<TObject, TKey>> source, IObservableCollection<TObject> destination, int refreshThreshold = BindingOptions.DefaultResetThreshold)
         where TObject : notnull
         where TKey : notnull
     {
         if (source is null) throw new ArgumentNullException(nameof(source));
         if (destination is null) throw new ArgumentNullException(nameof(destination));
 
-        return source.Bind(destination, new ObservableCollectionAdaptor<TObject, TKey>(refreshThreshold));
+        // if user has not specified different defaults, use system wide defaults instead.
+        // This is a hack to retro fit system wide defaults which override the hard coded defaults above
+        var defaults = DynamicDataOptions.Binding;
+
+        var options = refreshThreshold == BindingOptions.DefaultResetThreshold
+            ? defaults
+            : defaults with { ResetThreshold = refreshThreshold };
+
+        return source.Bind(destination, new ObservableCollectionAdaptor<TObject, TKey>(options));
+    }
+
+    /// <summary>
+    ///  Binds the results to the specified observable collection using the default update algorithm.
+    /// </summary>
+    /// <typeparam name="TObject">The type of the object.</typeparam>
+    /// <typeparam name="TKey">The type of the key.</typeparam>
+    /// <param name="source">The source.</param>
+    /// <param name="destination">The destination.</param>
+    /// <param name="options"> The binding options.</param>
+    /// <returns>An observable which will emit change sets.</returns>
+    /// <exception cref="System.ArgumentNullException">source.</exception>
+    public static IObservable<IChangeSet<TObject, TKey>> Bind<TObject, TKey>(this IObservable<IChangeSet<TObject, TKey>> source, IObservableCollection<TObject> destination, BindingOptions options)
+        where TObject : notnull
+        where TKey : notnull
+    {
+        if (source is null) throw new ArgumentNullException(nameof(source));
+        if (destination is null) throw new ArgumentNullException(nameof(destination));
+
+        return source.Bind(destination, new ObservableCollectionAdaptor<TObject, TKey>(options));
     }
 
     /// <summary>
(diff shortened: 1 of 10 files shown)


src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -731,12 +743,10 @@ public static class ObservableCacheEx
...
-    /// <param name="resetThreshold">The number of changes before a reset event is called on the observable collection.</param>
-    /// <param name="useReplaceForUpdates"> Use replace instead of remove / add for updates.  NB: Some platforms to not support replace notifications for binding.</param>
-    /// <param name="adaptor">Specify an adaptor to change the algorithm to update the target collection.</param>
+    /// <param name="options"> The binding options.</param>
     /// <returns>An observable which will emit change sets.</returns>
     /// <exception cref="System.ArgumentNullException">source.</exception>
-    public static IObservable<IChangeSet<TObject, TKey>> Bind<TObject, TKey>(this IObservable<ISortedChangeSet<TObject, TKey>> source, out ReadOnlyObservableCollection<TObject> readOnlyObservableCollection, int resetThreshold = 25, bool useReplaceForUpdates = true, ISortedObservableCollectionAdaptor<TObject, TKey>? adaptor = null)
+    public static IObservable<IChangeSet<TObject, TKey>> Bind<TObject, TKey>(this IObservable<ISortedChangeSet<TObject, TKey>> source, out ReadOnlyObservableCollection<TObject> readOnlyObservableCollection, BindingOptions options)
```
dwcullop, line 749: I wasn't sure about "Global" options but as long as the can be overwritten on a case-by-case basis, I'm good with it.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -298,7 +298,44 @@ public static IObservable<IChangeSet<T>> Bind<T>(this IObservable<IChangeSet<T>>
...
+    /// <returns>An observable which emits the change set.</returns>
+    /// <exception cref="System.ArgumentNullException">
+    /// source
+    /// or
+    /// targetCollection.
+    /// </exception>
+    public static IObservable<IChangeSet<T>> Bind<T>(this IObservable<IChangeSet<T>> source, IObservableCollection<T> targetCollection, BindingOptions options)
+        where T : notnull
+    {
+        if (source is null)
```
dwcullop, line 328: For consistency with other changes, consider not using the braces here.

RolandPheasant:
> Other than using the braces in the null checks, I like everything about this.
> 
> Null checks can also be succinctly like:
> 
> ```cs
>     _ = parameter ?? throw new ArgumentNullException(nameof(parameter));
> ```

I would like an options which says to the style police to use either this format or a single line conditional without the braces.  Both are fine with me, as I find using up the view port for null checks is needless.  Originally the entire of DD was that way then I accepted a PR which introduced style cop which  reformatted the entire thing to add all the braces,  I did not realise this at the time and have been fighting with it ever since,

@ChrisPulman is there a way to tell style cop that single line conditionals are OK without braces?  If so can we reformat all with the new rule?

src/DynamicData/List/ObservableListEx.cs
RolandPheasant: See comment below

src/DynamicData/Cache/ObservableCacheEx.cs
RolandPheasant: The global part sort of future proofs binding, Especially as machines are becoming more powerful it's good to be able to  in one swoop increase the thresholds.

It's a shame all those other overloads have to be maintained but it would cause too much pain to remove them all.  We should consider doing so on the next big version change.

dwcullop:
> @ChrisPulman is there a way to tell style cop that single line conditionals are OK without braces? If so can we reformat all with the new rule?

Actually, I just noticed that, after recent changes, StyleCop is suggesting the use of `ArgumentNullException.ThrowIfNull(parameter);` instead.

To me, it always makes sense to have the same pattern hidden behind an abstraction, even if it is very simple.

So, if there's going to be a mass replacement, my suggestion would be to use this instead.

ChrisPulman:
> ArgumentNullException.ThrowIfNull(parameter);

In order to do this, we need a net4.6.2 / netstandard2.0 version of the function. I understand we just need a couple of attributes and a static method for ArgumentNullException.ThrowIfNull, but I can't be certain that these two frameworks have the ability to replicate the functionality.

ChrisPulman:
> > ArgumentNullException.ThrowIfNull(parameter);
> 
> In order to do this, we need a net4.6.2 / netstandard2.0 version of the function. I understand we just need a couple of attributes and a static method for ArgumentNullException.ThrowIfNull, but I can't be certain that these two frameworks have the ability to replicate the functionality.

With some research only netstandard2.1 supports the NotNull attribute so unfortunately we are stuck with the dual code block. We may be able to build an extension method that shortcuts the issue in a better way though, I will see if I can come up with something.

RolandPheasant:
@dwcullop / @JakenVeina I've merged this, but will do some additional testing before release - I'll do a local package build and try it against some of my test / demo products,

JakenVeina:
> However, I wonder whether it's possible to run a formatter to change all the null reference exception checks with our code. Doing so would probably be difficult. Any ideas?

You kids and your aversions to just writing code. :P

I'm there are styling tools that'd do it, but I'm not familiar with any of them. I'll wager I can get nearly all of them with a good Regex Find/Replace. Make an issue so we don't forget, and I'd be happy to tackle it.

glennawatson:
Resharper and rider have tools for selecting one style

dwcullop:
If we can't use the standard static method, we should write our own (which could just invoke the standard one when building for a framework that supports it, maybe... Not sure what else the standard one does).  Either way, it will be good to have all of the code using the same abstraction for this pattern.

PR #780, opened 2023-12-04
List-to-List MergeManyChangeSets

dwcullop:
## Description

Implements the `MergeManyChangeSets` operator for cases when both the Parent and Child ChangeSets are List Changes.  It requires only a single prototype:

```csharp
IObservable<IChangeSet<TDestination>> MergeManyChangeSets<TObject, TDestination>(this IObservable<IChangeSet<TObject>> source, Func<TObject, IObservable<IChangeSet<TDestination>>> observableSelector);
```

The implementation is very similar to the Cache version except the Change Tracker is much simpler because it doesn't have to deal with duplicate keys.

## Testing

In addition to verifying Null handling and OnComplete/OnError behavior, the Unit Tests cover the following operations for both the Source and Child lists:
1) Add
1) AddRange
1) Insert
1) Remove
1) RemoveAt
1) RemoveMany
1) RemoveRange
1) Replace
1) Clear

## Other Changes

Other changes include:
1) Adding use of `Bogus` library to generate test data
1) Various code modernization changes consistent with similar recent efforts
1) Minor comment improvements

### DIFF
--- a/src/DynamicData.Tests/Domain/Animal.cs
+++ b/src/DynamicData.Tests/Domain/Animal.cs
@@ -15,18 +15,11 @@ public enum AnimalFamily
     Bird
 }
 
-public class Animal : AbstractNotifyPropertyChanged
+public class Animal(string name, string type, AnimalFamily family, bool include = true) : AbstractNotifyPropertyChanged
 {
-    private bool _includeInResults;
+    private bool _includeInResults = include;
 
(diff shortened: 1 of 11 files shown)


RolandPheasant (approved):
Looks great to me. Low risk, and I'm sure you'll test them all out on your daytime projects.

PR #781, opened 2023-12-05
Remove some index -1 exceptions

### LINKED ISSUE #683: [Bug]: System.ArgumentException: index cannot be greater than the size of the collection
### Describe the bug 🐞

Regulary when we start our application we get the following exception and the application crashes. Could you maybe tell what is the best way to catch the exception, so I can be more specific about the situation?

`System.ArgumentException: index cannot be greater than the size of the collection
   at DynamicData.ChangeAwareList`1.Insert(Int32 index, T item) in /_/src/DynamicData/List/ChangeAwareList.cs:line 223
   at DynamicData.Binding.ObservableCollectionEx.<>c__DisplayClass5_0`2.<ToObservableChangeSet>b__1(ChangeAwareList`1 list, EventPattern`1 args) in /_/src/DynamicData/Binding/ObservableCollectionEx.cs:line 159
   at System.Reactive.Linq.ObservableImpl.Scan`2._.OnNext(TSource value) in /_/Rx.NET/Source/src/System.Reactive/Linq/Observable/Scan.cs:line 40
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Reactive.Stubs.<>c.<.cctor>b__2_1(Exception ex) in /_/Rx.NET/Source/src/System.Reactive/Internal/Stubs.cs:line 16
   at System.Reactive.ObserverBase`1.OnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/ObserverBase.cs:line 61
   at System.Reactive.AutoDetachObserver`1.OnErrorCore(Exception exception) in /_/Rx.NET/Source/src/System.Reactive/Internal/AutoDetachObserver.cs:line 77
   at System.Reactive.ObserverBase`1.OnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/ObserverBase.cs:line 61
   at System.Reactive.Sink`1.ForwardOnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/Internal/Sink.cs:line 61
   at System.Reactive.Sink`1.ForwardOnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/Internal/Sink.cs:line 61
   at System.Reactive.Concurrency.Synchronize`1._.OnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/Concurrency/Synchronization.Synchronize.cs:line 50
   at System.Reactive.AutoDetachObserver`1.OnErrorCore(Exception exception) in /_/Rx.NET/Source/src/System.Reactive/Internal/AutoDetachObserver.cs:line 77
   at System.Reactive.ObserverBase`1.OnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/ObserverBase.cs:line 61
   at System.Reactive.Sink`1.ForwardOnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/Internal/Sink.cs:line 61
   at System.Reactive.ObserveOnObserverNew`1.DrainStep(ConcurrentQueue`1 q) in /_/Rx.NET/Source/src/System.Reactive/Internal/ScheduledObserver.cs:line 573
   at System.Reactive.ObserveOnObserverNew`1.DrainShortRunning(IScheduler recursiveScheduler) in /_/Rx.NET/Source/src/System.Reactive/Internal/ScheduledObserver.cs:line 509
   at System.Reactive.Concurrency.EventLoopScheduler.Run() in /_/Rx.NET/Source/src/System.Reactive/Concurrency/EventLoopScheduler.cs:line 356
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()`

### Step to reproduce

Currently I can not reproduce it yet. 

### Reproduction repository

None at the moment

### Expected behavior

No exception

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

_No response_

### Version

_No response_

### Device

_No response_

### DynamicData Version

7.12.11

### Additional information ℹ️

_No response_

### LINKED ISSUE #710: [Bug]: System.ArgumentException: index cannot be negative
### Describe the bug 🐞

I'm trying to transform the values in an AvaloniaDictionary (a dictionary implementing INotifyCollectionChanged) into a ReadOnlyObservableCollection for use in an application using the following:

`_hotkeyMap.ToObservableChangeSet<AvaloniaDictionary<Domain, Char?>, KeyValuePair<Domain, Char?>>()
                .Transform((x) => x.Value)
                .ObserveOn(RxApp.MainThreadScheduler)
                .Bind(out _hotkeysInUse)
                .Subscribe();`

Upon adding the first item to the dictionary an ArgumentException is being thrown, with the message "index cannot be negative", stack trace from exception as follows.
 	DynamicData.dll!DynamicData.ChangeAwareList<System.Collections.Generic.KeyValuePair<silmetra.plotsafe2.data.Model.Domain, char?>>.Insert(int index, System.Collections.Generic.KeyValuePair<silmetra.plotsafe2.data.Model.Domain, char?> item) Line 219	C#
 	DynamicData.dll!DynamicData.Binding.ObservableCollectionEx.ToObservableChangeSet.AnonymousMethod__1(DynamicData.ChangeAwareList<System.Collections.Generic.KeyValuePair<silmetra.plotsafe2.data.Model.Domain, char?>> list, System.Reactive.EventPattern<System.Collections.Specialized.NotifyCollectionChangedEventArgs> args) Line 161	C#
 	System.Reactive.dll!System.Reactive.Linq.ObservableImpl.Scan<System.Reactive.EventPattern<System.Collections.Specialized.NotifyCollectionChangedEventArgs>, DynamicData.ChangeAwareList<System.Collections.Generic.KeyValuePair<silmetra.plotsafe2.data.Model.Domain, char?>>>._.OnNext(System.Reactive.EventPattern<System.Collections.Specialized.NotifyCollectionChangedEventArgs> value) Line 40	C#
 	System.Reactive.dll!System.Reactive.Subjects.Subject<System.Reactive.EventPattern<System.Collections.Specialized.NotifyCollectionChangedEventArgs>>.OnNext(System.Reactive.EventPattern<System.Collections.Specialized.NotifyCollectionChangedEventArgs> value) Line 147	C#
Avalonia.Base.dll!Avalonia.Collections.AvaloniaDictionary<silmetra.plotsafe2.data.Model.Domain, char?>.NotifyAdd(silmetra.plotsafe2.data.Model.Domain key, char? value) Line 230	C#

Tracking it down, it seems to be coming from a similar place to #683 in ObservableCollectionEx

`case NotifyCollectionChangedAction.Add when changes.NewItems is not null:
                                {
                                    if (changes.NewItems.Count == 1 && changes.NewItems[0] is T item)
                                    {
                                        list.Insert(changes.NewStartingIndex, item);
                                    }
                                    else
                                    {
                                        list.InsertRange(changes.NewItems.Cast<T>(), changes.NewStartingIndex);
                                    }

                                    break;
                                }`

where NewStartingIndex is -1.  The AvaloniaDictionary is doing the following:

`if (CollectionChanged != null)
            {
                var e = new NotifyCollectionChangedEventArgs(
                    NotifyCollectionChangedAction.Add,
                    new[] { new KeyValuePair<TKey, TValue>(key, value) },
                    -1);
                CollectionChanged(this, e);
            }`

which is where I believe the -1 is coming from, but that also seems to be valid according to the documentation for NotifyCollectionChangedEventArgs.  Is this something I'm doing wrong, or is this a bug in ObservableCollectionEx as it appears to be?

### Reproduction repository

None as yet.

### Expected behavior

No exceptions should be thrown.

### Screenshots 🖼️

_No response_

### IDE

Visual Studio 2022

### Operating system

Windows

### Version

_No response_

### Device

_No response_

### DynamicData Version

_No response_

### Additional information ℹ️

_No response_

### LINKED ISSUE #723: [Bug]: `AddRange` `ChangeSet` has unexpected `Index` of -1
### Describe the bug 🐞

While writing a new operator for our internal use, I unexpectedly encountered an `AddRange` `ChangeSet` with an `Index` of -1. I traced it back to [line 88 in SourceList.Connect()](https://github.com/reactivemarbles/DynamicData/blob/main/src/DynamicData/List/SourceList.cs#L88), which leaves the index at its default of -1.

One of two things is true:
- It's legal for an `AddRange` `ChangeSet` to have an index of -1, and it presumably means something like "this is the initial state of the collection". If that's the case, it ought to be documented somewhere, and I've found nothing.
- This is a bug and 0 should be passed as the index.

### Steps to reproduce

        var source = new SourceList<string>();
        source.Add("A");
        var changeSets = new List<IChangeSet<string>>();
        source.Connect().Subscribe(changeSets.Add).Dispose();
        Trace.Assert(changeSets[0].First().Type == ChangeType.Range);

        // This assertion fails
        Trace.Assert(changeSets[0].First().Range.Index >= 0);

### Expected behavior

I was expecting Index to equal zero for an `AddRange` that initialises the collection.

### DynamicData Version

7.13.1

RolandPheasant:
1.  Prevent index -1 exception when connecting to source list. Fixes #723 
2.  Prevent further  index=-1 exceptions when converting an observable collection (including AvaloniaDictionary). Fixes #683 and fixes #710


src/DynamicData.Tests/Binding/AvaloniaDictionaryFixture.cs
```diff
@@ -0,0 +1,303 @@
...
+
+public interface IAvaloniaReadOnlyDictionary<TKey, TValue>
+    : IReadOnlyDictionary<TKey, TValue>,
+        INotifyCollectionChanged,
+        INotifyPropertyChanged
+    where TKey : notnull
+{
+}
+
+public class AvaloniaDictionary<TKey, TValue> : IAvaloniaDictionary<TKey, TValue> where TKey : notnull
```
RolandPheasant, line 85: I copied this from the AvaloniaProject so we can test it,

src/DynamicData/Kernel/IndexMinusOneException.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    private const string DefaultMessage = "Index of minus one detected. Please raise an issue for DynamicData with an exact re-production";
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="IndexMinusOneException"/> class.
+    /// </summary>
+    /// <param name="context"> Some context.</param>
+    /// <param name="member"> The calling member.</param>
+    /// <param name="line"> The calling line number..</param>
+    public IndexMinusOneException(string context, [CallerMemberName] string? member = null, [CallerLineNumber] int line = 0)
```
RolandPheasant, line 20: We should throw this beat whenever we detect a -1 index (unless of course we can handle it)

src/DynamicData.Tests/Binding/AvaloniaDictionaryFixture.cs
JakenVeina: A little weird, but okay. Me,

I think what I would have done is just improve coverage, in general, on the fixture for this `ToObservableChangeSet()` operator. To me, the core issue is that the operator wasn't compatible with the full functionality of `INotifyCollectionChanged`, `AvaloniaDictionary` just happens to be one of the implementations that leverages functionality we don't support. I.E. 

I will say you should add a comment here, mentioning that this is copied directly form Avalonia, and why.

src/DynamicData.Tests/Binding/AvaloniaDictionaryFixture.cs
```diff
@@ -0,0 +1,303 @@
...
+
+        _collection.Add("Someone", person);
+
+        _results.Messages.Count.Should().Be(1);
+        _results.Data.Count.Should().Be(1);
+        _results.Data.Items.First().Should().Be(person);
+    }
+
+    [Fact]
+    public void Replace()
```
JakenVeina, line 42: This is the particular scenario that was causing exceptions, right?

src/DynamicData/Kernel/IndexMinusOneException.cs
JakenVeina: `MissingIndexException`? `UnspecifiedIndexException`?

.editorconfig
```diff
@@ -699,6 +699,7 @@ csharp_style_namespace_declarations = block_scoped:silent
 csharp_style_prefer_method_group_conversion = true:silent
 csharp_style_prefer_top_level_statements = true:silent
 csharp_style_prefer_primary_constructors = true:suggestion
+dotnet_diagnostic.RCS1194.severity = none
```
JakenVeina, line 703: What's the wisdom on not doing the full constructor set for custom exceptions? Is that not recommended practice, anymore? In my mind, I always thought the biggest reason wasn't just for consistency, but that it has something to do with serialization of exceptions, which is of course pretty important for systems that, say, like building crash reports.

JakenVeina (changes requested):

.editorconfig
RolandPheasant: It is saying that I should implement several different constructors because it's "standard", when the reality is it is created for internal use by dynamic data, in which case I am in total control of which constructors are required, The recommendation is dumb because it is not intended for public use, and if some decides to throw it outside the lib, that is not my concern.

src/DynamicData.Tests/Binding/AvaloniaDictionaryFixture.cs
RolandPheasant: In fact each of the methods in this test causes an exception.

RolandPheasant:
@JakenVeina 

> I will say you should add a comment here, mentioning that this is copied directly form Avalonia, and why.

I actually thought that but didn't bother. You're right.  Comment added and pushed.

RolandPheasant:
@JakenVeina 

> UnspecifiedIndexException?

Ouch, Good call. I'd forgotten this was even a thing,  It's been so many years.  I'll consolidate

src/DynamicData.Tests/Binding/AvaloniaDictionaryFixture.cs
RolandPheasant: With the standard observable collection these scenarios cannot be reproduced - period.  I had covered everything on earth that MSFT produced but specifically the Avalonia Dictionary give alternative results.

The only other viable reproduction would be to re-create the Avalonia dictionary in all but name, so I may as well use the real thing.

src/DynamicData/Kernel/IndexMinusOneException.cs
RolandPheasant: Now uses UnspecifiedIndexException

JakenVeina:
@RolandPheasant 

> Ouch, Good call. I'd forgotten this was even a thing,  It's been so many years.  I'll consolidate

Yeah, that's definitely what I was referring to. Not just attempting to suggest a different name that I didn't think already existed.

JakenVeina (approved):

PR #760, opened 2023-11-20
Testing rework: `.ToObservableChangeSet()`

JakenVeina:
Rebuilt test fixtures for both `.ToObservableChangeSet()` operators, increasting code coverage from 69% to 100% across the internal implementation classes.

Identified existing defects within the `.ToObservableChangeSet()` operators, mostly related to error and completion propagation, and implemented testing for this behavior, with the tests being marked as "Skip" until the defects are fixed. This also includes the defect referred in #635.

Added an `IsCompleted` flag to `ChangeSetAggregator<T>`, to allow for testing of completion propagation.

PR #782, opened 2023-12-05
Removed dead testing code

JakenVeina:
Removed dead testing code that was disabled previously, and is now superseded by fd9083f90eef3301cb78961886054598ab2d4931 from #760.


dwcullop (approved):
This should be fine.

PR #783, opened 2023-12-06
Added missing API verification changes.

JakenVeina:
It seems there's something wrong with the PR build process, cause it should have started failing this 8 commits ago.


RolandPheasant:
Ooh. That useReplaceForUpdates default  change from false to true was unintentional. I'll double check and my revert that

RolandPheasant (approved):

PR #785, opened 2023-12-06
Benchmarks Enhancements and Organization

JakenVeina:
Fixed broken setup for displaying Benchmark artifacts in Solution Explorer.

Moved performance tests in the .Tests project over to the .Benchmarks project.

Total test runtime (on my machine) is now roughly 18.5 seconds, down from 1.1 minutes.


RolandPheasant:
@JakenVeina I've merged main into this and the builds fails.

JakenVeina:
That's very strange. I only turned NRT on specifically because it didn't produce any errors on my end.

RolandPheasant (approved):

PR #788, opened 2023-12-08
Apply locking to fix  join race conditions. Fixes #787

### LINKED ISSUE #787: [Bug]: "InvalidOperationException: Collection was modified" When using joins
### Describe the bug 🐞

This is the come-back of #364  
Subscribing to a join is still not thread-safe, as GetInitialUpdates will enumerate the underlying Dictionary, which could be updated at the same time. 

### Step to reproduce

```
[Fact]
    public void LetsSeeWhetherWeCanRandomlyHitADifferentRaceCondition()
    {
        var leftCache = new SourceCache<Thing, long>(x => x.Id);
        var rightCache = new SourceCache<Thing, long>(x => x.Id);
        var joined = leftCache.Connect().InnerJoin(rightCache.Connect(), x => x.Id,
            (keys, leftThing, rightThing) =>
                new Thing { Id = keys.leftKey, Name = $"{leftThing.Name} x {rightThing.Name}" });

        IDisposable StartUpdating(ISourceCache<Thing, long> sourceCache) =>
            Observable.Range(1, 1000000, Scheduler.Default)
                .Subscribe(x => sourceCache.AddOrUpdate(new Thing { Id = x, Name = $"{x}" }));

        using var leftUpdater = StartUpdating(leftCache);
        using var rightUpdater = StartUpdating(rightCache);
        using var subscription = joined.Subscribe();
    }
```

### Reproduction repository

https://github.com/rcm-adrien-gervaise/DynamicData

### Expected behavior

No exception should be thrown

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

_No response_

### Version

_No response_

### Device

_No response_

### DynamicData Version

_No response_

### Additional information ℹ️

Test reproducing the issue and suggested fix on https://github.com/rcm-adrien-gervaise/DynamicData

RolandPheasant:
Fixes race condition for cache joins.  See #787

### DIFF
--- a/src/DynamicData.Tests/Cache/InnerJoinFixtureRaceCondition.cs
+++ b/src/DynamicData.Tests/Cache/InnerJoinFixtureRaceCondition.cs
@@ -32,6 +32,25 @@ public class InnerJoinFixtureRaceCondition
         ids.InnerJoin(itemsCache.Connect(), x => x.Id, (_, thing) => thing).Subscribe((z) => { }, ex => { }, () => { });
     }
 
+    // See https://github.com/reactivemarbles/DynamicData/issues/787 
+    [Fact(Skip = "Heavyweight test so probably best run manually if this issue comes up again")]
+    public void LetsSeeWhetherWeCanRandomlyHitADifferentRaceCondition()
+    {
+        var leftCache = new SourceCache<Thing, long>(x => x.Id);
+        var rightCache = new SourceCache<Thing, long>(x => x.Id);
+        var joined = leftCache.Connect().InnerJoin(rightCache.Connect(), x => x.Id,
+            (keys, leftThing, rightThing) =>
+                new Thing { Id = keys.leftKey, Name = $"{leftThing.Name} x {rightThing.Name}" });
+
+        IDisposable StartUpdating(ISourceCache<Thing, long> sourceCache) =>
+            Observable.Range(1, 1_000_000, Scheduler.Default)
+                .Subscribe(x => sourceCache.AddOrUpdate(new Thing { Id = x, Name = $"{x}" }));
+
+        using var leftUpdater = StartUpdating(leftCache);
+        using var rightUpdater = StartUpdating(rightCache);
+        using var subscription = joined.Subscribe();
+    }
+
     public class Thing
     {
         public long Id { get; set; }
--- a/src/DynamicData.Tests/Cache/InnerJoinFixtureRaceCondition.cs
+++ b/src/DynamicData.Tests/Cache/InnerJoinFixtureRaceCondition.cs
@@ -33,22 +33,43 @@ public class InnerJoinFixtureRaceCondition
     }
 
     // See https://github.com/reactivemarbles/DynamicData/issues/787 
-    [Fact(Skip = "Heavyweight test so probably best run manually if this issue comes up again")]
+    [Fact]
     public void LetsSeeWhetherWeCanRandomlyHitADifferentRaceCondition()
     {
-        var leftCache = new SourceCache<Thing, long>(x => x.Id);
-        var rightCache = new SourceCache<Thing, long>(x => x.Id);
-        var joined = leftCache.Connect().InnerJoin(rightCache.Connect(), x => x.Id,
-            (keys, leftThing, rightThing) =>
-                new Thing { Id = keys.leftKey, Name = $"{leftThing.Name} x {rightThing.Name}" });
+        using var leftSource = new SourceCache<Thing, long>(thing => thing.Id);
+        using var rightSource = new SourceCache<Thing, long>(thing => thing.Id);
 
-        IDisposable StartUpdating(ISourceCache<Thing, long> sourceCache) =>
-            Observable.Range(1, 1_000_000, Scheduler.Default)
-                .Subscribe(x => sourceCache.AddOrUpdate(new Thing { Id = x, Name = $"{x}" }));
+        var resultStream = ObservableCacheEx.InnerJoin(
+            left: leftSource.Connect(),
+            right: rightSource.Connect(),
+            rightKeySelector: rightThing => rightThing.Id,
+            (keys, leftThing, rightThing) => new Thing()
+            {
+                Id = keys.leftKey,
+                Name = $"{leftThing.Name} x {rightThing.Name}"
+            });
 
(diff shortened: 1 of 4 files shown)


src/DynamicData.Tests/Cache/InnerJoinFixtureRaceCondition.cs
```diff
@@ -32,6 +32,25 @@ public void LetsSeeWhetherWeCanRandomlyHitARaceCondition()
...
+    public void LetsSeeWhetherWeCanRandomlyHitADifferentRaceCondition()
+    {
+        var leftCache = new SourceCache<Thing, long>(x => x.Id);
+        var rightCache = new SourceCache<Thing, long>(x => x.Id);
+        var joined = leftCache.Connect().InnerJoin(rightCache.Connect(), x => x.Id,
+            (keys, leftThing, rightThing) =>
+                new Thing { Id = keys.leftKey, Name = $"{leftThing.Name} x {rightThing.Name}" });
+
+        IDisposable StartUpdating(ISourceCache<Thing, long> sourceCache) =>
+            Observable.Range(1, 1_000_000, Scheduler.Default)
```
JakenVeina, line 46: Surely we could replicate the issue with fewer than 1,000,000 items. Then maybe the test would be feasible to include for normal runs? Is the issue really that rare?

src/DynamicData/Cache/Internal/FullJoin.cs
```diff
@@ -118,6 +118,9 @@ internal class FullJoin<TLeft, TLeftKey, TRight, TRightKey, TDestination>(IObser
                     return joinedCache.CaptureChanges();
                 });
 
-                return new CompositeDisposable(leftLoader.Merge(rightLoader).SubscribeSafe(observer), leftCache, rightCache);
+                lock (locker)
```
JakenVeina, line 121: This seems bizarre to me. The resource we need to lock around is the two caches, right? Those aren't actually being accessed here, just passed around by reference. By my reckoning, the only way a read or write occurs on one of the caches, from inside this lock, is if the `.Merge()` stream emits unsynchronized notifications *during* subscription, and that would be a defect inside the `.Merge()` operator.

src/DynamicData.Tests/Cache/InnerJoinFixtureRaceCondition.cs
RolandPheasant: Perhaps we could make it 1000 times.  However, it may potentially become a traffic light test whereas with these numbers I was able to replicate it every time.

src/DynamicData/Cache/Internal/FullJoin.cs
RolandPheasant: That was my initial thought, but then again this operator by nature produces asynchrony ands the rx guideline is to make sure your custom operators are synchronised.

JakenVeina: Right, which is why the other recommendation is to rely on operators you use to only emit synchronized notifications.

Basically, this suggests to me that there's a bug inside `.Merge()` which this doesn't actually fix. Unless I'm missing something, at the moment I'm only looking at all this on my phone.

RolandPheasant: I agree, but what is certain is this change fixes the issue.

src/DynamicData.Tests/Cache/InnerJoinFixtureRaceCondition.cs
JakenVeina: Dropping to `100_000` gave me 10 failures in a row without the fix, and an average runtime of 500ms with the fix.

JakenVeina: I got it down to about 150ms on average with the fix, and 100 failures in a row without it.

I realized that there are 3 things racing against each other here, but the test was only repeating 2 of those things. The `.Subscribe()` call is the third thing, and it was only running once, right at the end of the test.

I also setup the item generator "threads" to run infinitely, but start removing items after the first 100, so that calls to `.Subscribe()` don't take progressively more and more time.

```cs
using var leftSource = new SourceCache<Thing, long>(thing => thing.Id);
using var rightSource = new SourceCache<Thing, long>(thing => thing.Id);

var resultStream = ObservableCacheEx.InnerJoin(
    left: leftSource.Connect(),
    right: rightSource.Connect(),
    rightKeySelector: rightThing => rightThing.Id,
    (keys, leftThing, rightThing) => new Thing()
    {
        Id = keys.leftKey,
        Name = $"{leftThing.Name} x {rightThing.Name}"
    });

using var leftThingGenerator = BeginGeneratingThings(leftSource, "Left");
using var rightThingGenerator = BeginGeneratingThings(rightSource, "Left");

for (var i = 0; i < 100; ++i)
{
    using var subscription = resultStream.Subscribe();
}

IDisposable BeginGeneratingThings(SourceCache<Thing, long> source, string namePrefix)
    => Observable.Range(1, int.MaxValue, ThreadPoolScheduler.Instance)
        .Subscribe(id =>
        {
            source.AddOrUpdate(new Thing()
            {
                Id = id,
                Name = $"{namePrefix}Thing #{id}"
            });
            source.RemoveKey(id - 100);
        });
```

RolandPheasant: @JakenVeina you test is included in the PR now -> nice work.

JakenVeina (approved):

src/DynamicData/Cache/Internal/FullJoin.cs
JakenVeina: I played around with it last night, and yeah, there really is only two effective fixes, without significantly redesigning the operator: this, or swapping the `.AsObservableCache(false)` calls to `.AsObservableCache(true)`. The issue isn't that notifications aren't being synchronized, it's that there's code running _before_ the first notification, to build up the initial changeset, which accesses the caches shared by all subscribers, and because it's buried inside `.Connect().Subscribe()` we can't lock around it directly.

RolandPheasant: Yeah, originally it used the locked version of the cache but I optimised later but changing that to use unlocked with more operator based locking.

PR #798, opened 2023-12-16
Feature: Support DebugSpy in Release Builds

dwcullop:
## Description

Improvements to `Spy` / `DebugSpy` operators:
- Adds support for the `DebugSpy` operator to work even in release builds (requires `DEBUG_SPY_ALWAYS` build flag to be set)
- Will now include a subcription Id with each output so that when an Observable has multiple simultaneous subscribers, it is easy to identify which messages go with which subscriptions
- Improves output of `Spy` / `DebugSpy` for List ChangeSets

### DIFF
--- a/src/DynamicData.Tests/Utilities/ObservableSpy.cs
+++ b/src/DynamicData.Tests/Utilities/ObservableSpy.cs
@@ -3,13 +3,22 @@ using System.Diagnostics;
 using System.Linq;
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
+using System.Runtime.InteropServices;
 using System.Threading;
+using VerifyTests;
 
 namespace DynamicData.Tests.Utilities;
 
 internal static class ObservableSpy
 {
     private static readonly string ChangeSetEntrySpacing = Environment.NewLine + "\t";
+
+    private static class NativeMethods
+    {
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode, EntryPoint = "OutputDebugStringW")]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Interoperability", "SYSLIB1054:Use 'LibraryImportAttribute' instead of 'DllImportAttribute' to generate P/Invoke marshalling code at compile time", Justification = "Affects operation")]
+        public static extern void OutputDebugString(string lpOutputString);
+    }
 
     /// <summary>
     /// Spys on the given IObservable{T} by emitting logging information that is tagged with the current ThreadId for all related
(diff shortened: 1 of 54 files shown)


src/DynamicData.Tests/Utilities/ObservableSpy.cs
```diff
@@ -71,110 +81,131 @@ internal static class ObservableSpy
...
+                { HasValue: true, Value: T val } => $" [Previous: {formatter(val)}]",
+                _ => string.Empty,
+            });
 
     private static Action<string> CreateLogger(Action<string> baseLogger, Func<string> timeStamper, string opName) =>
             msg => baseLogger($"{timeStamper()}[{Environment.CurrentManagedThreadId:X2}] |{opName}| {msg}");
 
-#if DEBUG
-    static void DebugLogger(string str) => Debug.WriteLine(str);
+#if DEBUG || DEBUG_SPY_ALWAYS
```
JakenVeina, line 203: I feel like this should be...

```cs
#if DEBUG
    private static void DebugLogger(string str) => Debug.WriteLine(str); 
#elif DEBUG_SPY_ALWAYS
    private static void DebugLogger(string str) => NativeMethods.OutputDebugString(str); 
#endif
```

I.E. the native call only gets used for Release builds.

RolandPheasant: I agree.

src/DynamicData.Tests/Utilities/ObservableSpy.cs
```diff
@@ -11,6 +12,13 @@ internal static class ObservableSpy
 {
     private static readonly string ChangeSetEntrySpacing = Environment.NewLine + "\t";
 
+    private static class NativeMethods
+    {
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode, EntryPoint = "OutputDebugStringW")]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Interoperability", "SYSLIB1054:Use 'LibraryImportAttribute' instead of 'DllImportAttribute' to generate P/Invoke marshalling code at compile time", Justification = "Affects operation")]
+        public static extern void OutputDebugString(string lpOutputString);
```
RolandPheasant, line 19: One issue with using native methods is it would prevent us from promoting this to the main lib.

dwcullop: Per the discussion on Slack, I'm okay with that.  This is something to only be used with the test project.  If we move another to the main lib, it would be the base functions, and not any overloads that use this.

I wouldn't even suggest checking in code that enables this, it is just a temporary for situations where an issue won't repro in debug builds.

JakenVeina (approved):

PR #800, opened 2023-12-18
Feature:  Stress Tests for List-to-List MergeManyChangeSets

dwcullop:
## Description
- Migrates the stress tests from the Cache-List version of MergeManyChangeSets test fixture to the List-List version, but consolidates them so they don't take so long to run
- Applies improvements made to the Cache-List version as well to keep them in sync
- Aligns implementation of List-List with the Cache-List version (removes use of intermediate SourceList)
- Fixes minor bug in the List-List version where it wasn't passing the `IEqualityComparer` implementation to the internal class

### DIFF
--- a/src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
+++ b/src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reactive;
 using System.Reactive.Concurrency;
@@ -25,8 +26,6 @@ public sealed class MergeManyChangeSetsListFixture : IDisposable
     const int AddRangeSize = 53;
     const int RemoveRangeSize = 37;
 #endif
-    private static readonly TimeSpan s_MaxAddTime = TimeSpan.FromSeconds(0.1);
-    private static readonly TimeSpan s_MaxRemoveTime = TimeSpan.FromSeconds(1.0);
 
     private readonly ISourceList<AnimalOwner> _animalOwners = new SourceList<AnimalOwner>();
     private readonly ChangeSetAggregator<AnimalOwner> _animalOwnerResults;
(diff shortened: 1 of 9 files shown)


src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
```diff
@@ -39,6 +46,71 @@ public MergeManyChangeSetsListFixture()
...
+                        mergedSub.Dispose();
+                    }
+                }
+                while (addingAnimals);
+            };
+
+        // Act
+
+        // Assert
+        CreateTest(TaskPoolScheduler.Default, ownerCount, animalCount).Should().NotThrowAsync();
```
JakenVeina, line 111: Are we just going for, like, clarification of intent here? Like, the reason there's no other assertions is that we're specifically testing to look for exceptions?

src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
```diff
@@ -39,6 +46,71 @@ public MergeManyChangeSetsListFixture()
         _animalResults = _animalOwners.Connect().MergeManyChangeSets(owner => owner.Animals.Connect()).AsAggregator();
     }
 
+    [Theory]
+    [InlineData(5, 7)]
+    [InlineData(10, 50)]
+    [InlineData(10, 1_000)]
+    [InlineData(200, 500)]
+    [InlineData(1_000, 10)]
+    public async Task MultiThreadedStressTest(int ownerCount, int animalCount) =>
```
JakenVeina, line 55: The runtime here is still roundabout 20 seconds worst-case, yeah? I'd really rather not have tests running that long that run, by default, for local devs.

I've had two thoughts for this, although neither is really a part of this PR.

A) Add a .runsettings file to the solution, which I think should allow us to configure default exclusions for all test runs. I worry, though, that this isn't applicable to devs that don't use VS (like, say, VS Code instead).

B) Keep stress tests to a minimum of 1s of run time (you can even bound the test by a timer, rather than a number of iterations), and then adjust our PR build scripts to take all stress tests and repeat them for X iterations.

src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
dwcullop: Exceptions or deadlocks.  Testing the final output is in the other test.
This one is just sub/unsub repeatedly while other threads do add/remove to both the parent and child containers.

I included this test because of the exception that was reported the other day that occurred when subscribing while something was modifying the container.

src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
dwcullop: Would local devs run them?  I don't run all the tests (except maybe at the end to confirm nothing else is broken).

I'm not familiar with a .runsettings file, but they don't need to be run by local devs.  At least not the ones that take a while.

JakenVeina: I definitely run all, at least before pushing,

When I was working on the `DisposeMany()` changes, and there were tests across the whole library that broke, I was running the full set of tests pretty regularly, and it was really cumbersome at the time, a full test run was several minutes, due to that set of performance tests, I moved over to the benchmarks project. There was no way for me to really tell that those tests weren't "supposed" to be run regularly, traits don't readily show in the Test Explorer.

Maybe I'm being overly-picky, but if the intention of a test is that you don't normally run it, I feel that it should be opt-in, not opt-out.

src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
JakenVeina: I meant specifically wrapping the test in a delegate, and then invoking it. Does that account for deadlocks? The test itself is fine, the delegate abstraction just seems pointless.

dwcullop: `.Should().NotThrowAsync()` only works on `Func<Task>`.  Is there a better way to explicitly test for the lack of an exception?  I'm certainly open to suggestions.

src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
dwcullop: I know it takes a bit to run, but I do think it should be run as part of the regular suite.  I am trying to avoid regressions that would cause the operator to stop returning the correct values in a stress condition.

I could reduce the values so they run faster, but that may not be stressful enough.

What we could do is reduce the number of tests.  We can just run the one really stressful case.

Do you run in the debugger?  Release mode?  I can have the other test cases be avoided in those circumstances.  Or figure out how to make them on demand.

src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
JakenVeina: No, my point is that explicitly testing for the lack of an exception seems pointless to me. ALL tests test for the lack of exceptions, implicitly.

src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
JakenVeina: I generally just run from Test Explorer, in Debug config.

I definitely don't want to remove stress testing entirely, I'm just trying to consider how to keep local dev runs reasonable. It's an inevitable fact that the more cumbersome tests are to run, the less devs will actually run them.

To me, option B is reasonable, we get to keep all tests, and we just do extra stressing for CI builds.

RolandPheasant: or 

C) Use a dedicate project for stress tests?  These could run on the build server by default.

src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
dwcullop: Okay, I'll fix it.

src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
dwcullop: So I'm fine with creating a dedicated project, but I'm not sure it should be part of this PR.  But after looking more carefully, I realized that both the stress test and the "no throw" test were doing the exact same work.

There's no need to do that twice, so I combined the tests.  It will do all of the async add/removes while doing sub/unsub ... And then at the end, it checks to make sure the results are fine.

I also added a parallelize option so that more of the add/removes happen on top of each other and I reduced the add/remove delay times.

All of these together should reduce the amount of time needed to wait for this test.  In another PR, we can move them to their own project.

dwcullop: I put all of the code needed for the Stress Test in one function.  None of the helper functions were used outside of that function anymore anyway.  I updated Cache-List version to match.

dwcullop:
@RolandPheasant  This one is ready I think.

RolandPheasant (approved):
Again, I am happy to leave this to your discretion.

Also this is all test code, and besides we can release when we are ready so zero risk.

src/DynamicData.Tests/List/MergeManyChangeSetsListFixture.cs
JakenVeina: > but I'm not sure it should be part of this PR

For sure, it's a totally separate discussion.

PR #801, opened 2023-12-18
TestSourceCache and TestSourceList

JakenVeina:
Added new implementations of `ISourceCache<>` and `ISourceList<>`, to allow manual injection of error and completion notifications, during tests for operators that work directly upon these interfaces, rather than `IObservable<IChangeSet>>`

Again, I needed a little more testing capability than we currently have, and I figured these will be useful for a handful of operators.

### DIFF
--- a/src/DynamicData.Tests/Utilities/TestSourceCache.cs
+++ b/src/DynamicData.Tests/Utilities/TestSourceCache.cs
@@ -1,0 +1,109 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
+
+using DynamicData.Kernel;
+
+namespace DynamicData.Tests.Utilities;
+
+public sealed class TestSourceCache<TObject, TKey>
+        : ISourceCache<TObject, TKey>
+    where TObject : notnull
+    where TKey : notnull
+{
+    private readonly IObservable<int> _countChanged;
+    private readonly BehaviorSubject<Exception?> _error;
+    private readonly BehaviorSubject<bool> _hasCompleted;
+    private readonly SourceCache<TObject, TKey> _source;
+
+    public TestSourceCache(Func<TObject, TKey> keySelector)
+    {
+        _error = new(null);
+        _hasCompleted = new(false);
+        _source = new(keySelector);
+
+        _countChanged = WrapStream(_source.CountChanged);
+    }
+
+    public int Count
+        => _source.Count;
+
+    public IObservable<int> CountChanged
+        => _countChanged;
+
+    public IEnumerable<TObject> Items
+        => _source.Items;
+    
+    public IEnumerable<TKey> Keys
+        => _source.Keys;
+
+    public Func<TObject, TKey> KeySelector
+        => _source.KeySelector;
+
+    public IEnumerable<KeyValuePair<TKey, TObject>> KeyValues
+        => KeyValues;
+
+    public void Complete()
+    {
+        AssertCanMutate();
+
+        _hasCompleted.OnNext(true);
+    }
+
+    public IObservable<IChangeSet<TObject, TKey>> Connect(
+            Func<TObject, bool>? predicate = null,
+            bool suppressEmptyChangeSets = true)
+        => WrapStream(_source.Connect(predicate, suppressEmptyChangeSets));
+
+    public void Dispose()
+    {
+        _error.Dispose();
+        _hasCompleted.Dispose();
+        _source.Dispose();
+    }
+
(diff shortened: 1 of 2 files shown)


src/DynamicData.Tests/Utilities/TestSourceCache.cs
```diff
@@ -0,0 +1,109 @@
...
+using DynamicData.Kernel;
+
+namespace DynamicData.Tests.Utilities;
+
+public sealed class TestSourceCache<TObject, TKey>
+        : ISourceCache<TObject, TKey>
+    where TObject : notnull
+    where TKey : notnull
+{
+    private readonly IObservable<int> _countChanged;
```
dwcullop, line 15: Could you just expose a single `Subject<IChangeSet<TObject, TKey>>` and then Connect could merge that subject with the actual Connect?

Then you could even add a method to inject ChangeSets into the stream.

I don't think you need two subjects because each Subject already handles OnComplete and OnError, I think.

src/DynamicData.Tests/Utilities/TestSourceCache.cs
```diff
@@ -0,0 +1,109 @@
...
+    private void AssertCanMutate()
+    {
+        if (_error.Value is not null)
+            throw new InvalidOperationException("The source collection is in an error state and cannot be mutated.");
+
+        if (_hasCompleted.Value)
+            throw new InvalidOperationException("The source collection is in a completed state and cannot be mutated.");
+    }
+
+    private IObservable<T> WrapStream<T>(IObservable<T> sourceStream)
```
dwcullop, line 98: If you go with the single Subject, then this can just Merge the two together, except Merge won't complete until both child streams complete, so you'd need to have some logic that checks a flag that is set if completion is forced.

dwcullop:
> Added new implementations of `ISourceCache<>` and `ISourceList<>`, to allow manual injection of error and completion notifications, during tests for operators that work directly upon these interfaces, rather than `IObservable<IChangeSet>>`

I love this.  This could be very useful.  My suggestion would be to use a single `Subject<IChangeSet<TObject,TKey>>` and merge a subscription to that with the `Connect` subscription and then the consumer could inject other changesets in addition to force complete and force an error.

JakenVeina:
> then the consumer could inject other changesets in addition to force complete and force an error.

I actually considered this a bit, but I didn't think it made sense to inject ACTUAL changes, directly, cause all that would do is guarantee that the downstream will be out-of-sync with the source, and I don't see how that could ever be a scenario we'd actually want to TEST for.

Using a single subject also wouldn't really cover the ability to inject errors and completions into `.CountChanged`, as the `<T>` wouldn't match. At least, not simply, it would mean I can't use a shared `WrapStream()` method for building `.CountChanged`. The only real advantage is to reduce memory footprint a little bit, and I'm really not worried about performance for a rarely-used test implement.

src/DynamicData.Tests/Utilities/TestSourceCache.cs
JakenVeina: If swapping to a single subject means I get rid of one clause for handling stream completion, only to replace it with a different one, that doesn't seem like a simplification to me.

dwcullop: Okay, I didn't see how many different things were being wrapped with this method.  This is covering ALL the observables, not just Connect.  No, I like this much better now.  It allows you to fail individual subscriptions without having to fail / complete the Subject (which probably isn't recoverable by design).

dwcullop:
> Using a single subject also wouldn't really cover the ability to inject errors and completions into `.CountChanged`, as the `<T>` wouldn't match. At least, not simply, it would mean I can't use a shared `WrapStream()` method for building `.CountChanged`.

Yeah, I didn't realize it was wrapping ALL of the exposed observables...  I just saw it being used for Connect.  What you say makes sense and this is really cool, but is the use-case for this to afford the ability to force errors/completions to operators in order to verify that they do what they're supposed to do?  

For tests I've written that needed that, I just Dispose the container to force OnComplete.  To force OnError, use `Concat` with `Observable.Throw` and then Dispose the container.

Is this just an easier way of doing that?  Does it have other benefits?

JakenVeina:
This is the ONLY way of doing it for operators that operate on `ISourceCache<>`/`ISourceList<>` instead of `IObservable<>`. You can't `.Concat()` onto `.Connect()` if you're not the one calling `.Connect()`.

Otherwise, yeah, `.Concat()` and `.Throw()` works great.

dwcullop (approved):
Thanks for answering my questions.  Looks good to me.

Issue #716, opened 2023-09-05
[Bug]: ExpireAfter might not expire the most recent item.


alex-borzenko:
### Describe the bug 🐞

When using ExpireAfter without polling but scheduling expiry action for each item, the most recent item might get stuck, if the scheduler fires a microsecond earlier then that item's due time. I had to work around it by wrapping the scheduler and adding a millisecond to the dueTime (works for my "slow" trading app :) ).

### Step to reproduce

    /// <summary>
    /// This test demonstrates the problem with ExpireAfter
    /// </summary>
    /// <param name="count"></param>
    /// <param name="delay"></param>
    [TestCase(10,10)]
    public async Task TestExpireAfter(int count, int delay)
    {
        var source = new SourceCache<int, int>(x => x);

        source
            .ExpireAfter(_ => TimeSpan.FromMilliseconds(delay), Scheduler.Default)
            .Subscribe();

        for (var i = 0; i < count; ++i)
        {
            source.AddOrUpdate(i);
            //wait 10 times longer then the expiration time for the item
            await Task.Delay(TimeSpan.FromMilliseconds(10*delay));
            if(source.Count>0)
                Console.WriteLine($"Missed: {i} Count: {source.Count}");
        }
        
        source.Dispose();
    }

### Reproduction repository

https://github.com/reactivemarbles/DynamicData/

### Expected behavior

There should be no output

### Screenshots 🖼️

Sample output:
Missed: 2 Count: 1
Missed: 6 Count: 1
Missed: 7 Count: 1

This proves that adding new item will expire the previous and sometimes! the new one.

### Operating system

Happened on Mac and Linux container

### DynamicData Version

7.14.2

JakenVeina:
It looks like a more direct workaround would be to supply the `pollingInterval` argument to `.ExpireAfter()`. That triggers single recurring background action to check all items for expiration, rather than scheduling a unique removal action per item.

JakenVeina:
Actually, ignore me. You specifically referenced the `SourceCache` version of the operator, and I went and dug into the `SourceList` version.

I'm gonna go ahead and take the same approach here as I did for `.ToObservableChangeSet()`, I.E. beef up the testing and benchmarking on the existing versions, then rebuild the operators from scratch, based on that. Especially, considering time-based expiration is fresh on my mind after that effort.

PR #809, opened 2023-12-23
Testing Utility: FakeScheduler

JakenVeina:
Added a new implementation of `IScheduler` to allow full control over execution of scheduled actions, by test orchestration code.

Needed this to be able to simulate inaccuracies in real-world schedulers, for intervals around 1ms, in particular with reference to #716.


dwcullop:
I don't have an issue with this, but I'd like to understand the need better.  The regular test scheduler allows for control down to the Tick (100ns).  What does this give us that the other doesn't?

A replacement Test Scheduler should also have the time control methods the other one does, like `AdvanceTo` and `AdvanceBy` (not sure if those are exactly the names).

JakenVeina:
The test secheduler does not allow simulation of inaccuracies. It will always invoke scheduled operations in exactly the at exactly the (virtual) time they were scheduled for.

At least, I don't see any APIs to control invocation and time virtualization separately.

JakenVeina:
I've identified the build failure as an un-related faulty test, which intermittently fails, and documented it as #817.

RolandPheasant (approved):
I won't know the difference this makes until you use it. However it's for testing, so let's get it in.

PR #819, opened 2024-01-04
TransformAsync enhancements

RolandPheasant:
1. Improve implementation of TransformAsync 
2. Add max concurrency param to prevent unbounded parallelisation.
3. Add transform on refresh overloads
4. Improve slow running concurrency test
5. Use TransformAsyncOptions for primitive objects - also include for Safe variants.

TODO:

1. ~Deal with large number of overloads required to include the extra optional params~
2. ~Maybe scrap 1 and use a TransformOptions object (similar to BindingOptions) and move away from the bewildering number of overloads.~

@dwcullop this is what I meant regarding using pure rx to handle the async.

### DIFF
--- a/src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
+++ b/src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
@@ -9,7 +9,6 @@ using DynamicData.Binding;
 using DynamicData.Tests.Domain;
 
 using FluentAssertions;
-
 using Xunit;
 
 namespace DynamicData.Tests.Cache;
@@ -204,6 +203,84 @@ public class TransformAsyncFixture
         stub.Results.Messages.Count.Should().Be(2, "Should be 2 updates");
         stub.Results.Messages[0].Adds.Should().Be(1, "Should be 1 adds");
         stub.Results.Messages[1].Updates.Should().Be(1, "Should be 1 update");
+    }
+
+
+    private record PersonWithAgeGroup(Person Person, string AgeGroup);
+
+
+
+    [Fact]
+    public void TransformWithoutRefresh()
+    {
+        using var source = new SourceCache<Person, string>(p => p.Name);
+        using var results = source.Connect()
+            .AutoRefresh()
+            .TransformAsync((p, key) => Task.FromResult(new PersonWithAgeGroup(p, p.Age < 18 ? "Child" : "Adult"))).AsAggregator();
+
+        var person = new Person("SomeOne", 16);
+        source.AddOrUpdate(person);
+
+        results.Data.Count.Should().Be(1);
+        results.Data.Lookup("SomeOne").Value.AgeGroup.Should().Be("Child");
+
+        person.Age = 21;
+
+
+        results.Data.Count.Should().Be(1);
+        results.Data.Lookup("SomeOne").Value.AgeGroup.Should().Be("Child");
+
+    }
+
+    [Fact]
+    public void TransformOnRefresh()
+    {
+        using var source = new SourceCache<Person, string>(p => p.Name);
+        using var results = source.Connect()
+            .AutoRefresh()
+            .TransformAsync((p, key) => Task.FromResult(new PersonWithAgeGroup(p, p.Age < 18  ? "Child" : "Adult")), TransformAsyncOptions.WithTransformOnRefresh).AsAggregator();
+
+        var person = new Person("SomeOne", 16);
+        source.AddOrUpdate(person);
+
+        results.Data.Count.Should().Be(1);
+        results.Data.Lookup("SomeOne").Value.AgeGroup.Should().Be("Child");
+        
+        person.Age = 21;
+
+
+        results.Data.Count.Should().Be(1);
+        results.Data.Lookup("SomeOne").Value.AgeGroup.Should().Be("Adult");
+
+    }
+
+    [Fact]
+    public async Task WithMaxConcurrency()
+    {
+        /* We need to test whether the max concurrency has any effect.
+
+             If  maxConcurrency == 100, this test takes a little more than 100 ms
+             If maxConcurrency = 10, this test takes a little more than 1s 
+
+            So it works, but how can it be tested in a scientific way ??
+        */
+
+        const int transformCount = 100;
+        const int maxConcurrency = 10;
+
+
+        using var source = new SourceCache<Person, string>(p => p.Name);
+        using var results = source.Connect()
+            .TransformAsync(async (p, key) =>
+            {
+                await Task.Delay(100);
+
+                return Task.FromResult(new PersonWithAgeGroup(p, p.Age < 18 ? "Child" : "Adult"));
+            }, TransformAsyncOptions.WithMaximumConcurrency(maxConcurrency)).AsAggregator();
+
+        source.AddOrUpdate(Enumerable.Range(1, transformCount).Select(l => new Person("Person" + l, l)));
+
+       await results.Data.CountChanged.Where(c=>c == transformCount).Take(1);
     }
 
     private class TransformStub : IDisposable
--- a/src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
+++ b/src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
@@ -5,7 +5,7 @@ using System.Reactive;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Threading.Tasks;
-
+using DynamicData.Binding;
 using DynamicData.Tests.Domain;
 
 using FluentAssertions;
@@ -93,7 +93,7 @@ public class TransformAsyncFixture
     public async Task RemoveFlowsToTheEnd()
     {
         var transform = 0;
-        var count = 500;
+        var count = 100;
         ReadOnlyObservableCollection<Person> collection;
 
         var cache = new SourceCache<Person, string>(p => p.Name);
@@ -121,9 +121,8 @@ public class TransformAsyncFixture
             cache.RemoveKey(p.Name);
         }
 
-        while (transform != count)
-            await Task.Delay(100);
-        await Task.Delay(3000);
+        await collection.ToObservableChangeSet().Take(count * 2);
+
         collection.Count.Should().Be(0);
     }
 
--- a/src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
+++ b/src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
@@ -254,8 +254,10 @@ public class TransformAsyncFixture
 
     }
 
-    [Fact]
-    public async Task WithMaxConcurrency()
+    
+    [Theory, InlineData(10), InlineData(100)]
+
+    public async Task WithMaxConcurrency(int maxConcurrency)
     {
         /* We need to test whether the max concurrency has any effect.
 
@@ -266,8 +268,6 @@ public class TransformAsyncFixture
         */
 
         const int transformCount = 100;
-        const int maxConcurrency = 10;
-
 
         using var source = new SourceCache<Person, string>(p => p.Name);
         using var results = source.Connect()
--- a/src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
+++ b/src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
@@ -258,10 +258,8 @@ public class TransformAsyncFixture
 
         source.AddOrUpdate(Enumerable.Range(1, transformCount).Select(l => new Person("Person" + l, l)));
 
-       // var xxx = Observable.Timer(TimeSpan.FromSeconds(1)).Select(_=> Observable.Thr);
-
-       await results.Data.CountChanged.Where(c=>c == transformCount).Take(1)
-           .Timeout(TimeSpan.FromSeconds(2));
+        await results.Data.CountChanged.Where(c => c == transformCount).Take(1)
+            .Timeout(TimeSpan.FromSeconds(2));
     }
 
     private class TransformStub : IDisposable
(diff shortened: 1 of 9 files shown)


src/DynamicData/Cache/Internal/TransformAsync.cs
```diff
@@ -3,71 +3,60 @@
...
 
 namespace DynamicData.Cache.Internal;
 
-internal sealed class TransformAsync<TDestination, TSource, TKey>(IObservable<IChangeSet<TSource, TKey>> source, Func<TSource, Optional<TSource>, TKey, Task<TDestination>> transformFactory, Action<Error<TSource, TKey>>? exceptionCallback, IObservable<Func<TSource, TKey, bool>>? forceTransform = null)
+internal class TransformAsync<TDestination, TSource, TKey>(
+    IObservable<IChangeSet<TSource, TKey>> source,
+    Func<TSource, Optional<TSource>, TKey, Task<TDestination>> transformFactory,
+    Action<Error<TSource, TKey>>? exceptionCallback,
+    IObservable<Func<TSource, TKey, bool>>? forceTransform = null,
+    int maximumConcurrency = 4,
```
JakenVeina, line 16: The introduction of this parameter is the main performance improvement here, yeah?

I would have the default be `null`, for a couple of reasons:

1. You're really just passing this parameter directly to `.Merge()` and `.Merge()` doesn't enforce a default, you either supply a limit, or there isn't one.
2. 4 is the reasonable default that .NET usually uses for CPU parallelism, but this isn't about limiting CPU parallelism. The practical scenario that this supports is about I/O concurrency, E.G. database queries.

Otherwise, this functionally looks good to me.

dwcullop: Yeah, I had no idea there was an overload of Merge that took a max concurrency parameter.  But having looked it up, I agree it should be null, and if it is, then invoke the overload that doesn't use one.

RolandPheasant: I only found out about the overload to limit concurrency recently myself. Rx is the gift that keeps giving.

You both make fair points, and the same also occurred to me. So it will be null.

What do you two think about point 2? Complex overloads are a cognitive pain which optional params tried to resolve. That's why I M thinking about moving towards options objects for primitive optional values

JakenVeina: Options objects wouldn't be TOO crazy, but I'm not sure where the line is for how many parameters it takes to merit a full object. Is there any wisdom we can glean from Microsoft's usages of that pattern? If we did do this, we would definitely want to implement Options objects as `readonly record struct` I think.

In my personal opinion, I don't consider optional parameters as being cognitive pain at *all*, because when I invoke almost any method where the parameter purpose isn't self-evident at a glance, I do...

```cs
var result = SomeMethod(
    parameter1: value1,
    parameter2: value2,
    parameter3: value3);
```

This plays seamlessly with optional parameters, cause all I have to do is omit the ones I don't want.

This is also pretty much functionally and semantically identical to using an options struct.

```cs
var result = SomeMethod(new()
{
    Option1 = value1,
    Option2 = value2,
    Option3 = value3
});

Both of these offer essentially the exact same cost for the method call, as all the parameters/options (used or not) are allocated on the stack. The difference really is just syntax, and how much code we have to write in the library.

**Comment by dwcullop** [comment]
> @dwcullop this is what I meant regarding using pure rx to handle the async.

Yeah, this is good stuff.  My work on `TransformManyAsync` does fall back onto `TransformAsync`, so this will improve that as well.  I'm going to try and update it to follow the `Rx-ness` you've demonstrated here.

File: src/DynamicData/Cache/Internal/TransformAsync.cs (hunk 1 above)
Comment (dwcullop) [reply src/DynamicData/Cache/Internal/TransformAsync.cs:16]: `record struct` is fine, but not all versions of C# support it.  Either way, make sure to pass it using `in` so that it isn't copied on the stack.

An options object is not a bad idea because it enables changes to be made without breaking things (adding new options, changing defaults) but we should also have overloads that allow a few commonly used parameters to be set without having to create the object.

File: src/DynamicData/Cache/Internal/TransformAsync.cs [hunk 2]
```diff
@@ -3,71 +3,60 @@
...
 
-    private async Task<IChangeSet<TDestination, TKey>> DoTransform(ChangeAwareCache<TransformedItemContainer, TKey> cache, IChangeSet<TSource, TKey> changes)
+    private IObservable<IChangeSet<TDestination, TKey>> DoTransform(
+        ChangeAwareCache<TransformedItemContainer, TKey> cache, Func<TSource, TKey, bool> shouldTransform)
     {
-        var transformed = await Task.WhenAll(changes.Select(Transform)).ConfigureAwait(false);
+        var toTransform = cache.KeyValues.Where(kvp => shouldTransform(kvp.Value.Source, kvp.Key)).Select(kvp =>
+            new Change<TSource, TKey>(ChangeReason.Update, kvp.Key, kvp.Value.Source, kvp.Value.Source)).ToArray();
+
+        return toTransform.Select(change => Observable.Defer(() => Transform(change).ToObservable()))
```
dwcullop, line 47: I think `Observable.FromAsync(() => Transform(change))` gives the same semantics as `Defer` plus `ToObservable` and more clearly expresses your intent.

dwcullop: So, you could do:

```csharp
return toTransform.Select(change => Transform(change))
        .Merge()
        .ToArray()
        .Select(transformed => ProcessUpdates(cache, transformed));
```

Or, if you want to provide the maximum concurrency parameter:

```csharp
return toTransform.Select(change => Observable.FromAsync(() => Transform(change)))
        .Merge(maximumConcurrency)
        .ToArray()
        .Select(transformed => ProcessUpdates(cache, transformed));
```

src/DynamicData/Cache/Internal/TransformAsync.cs
```diff
@@ -3,71 +3,60 @@
...
+            .Merge(maximumConcurrency < 1 ? int.MaxValue : maximumConcurrency)
+            .ToArray()
+            .Select(transformed => ProcessUpdates(cache, transformed));
+    }
 
-        return ProcessUpdates(cache, transformed);
+    private IObservable<IChangeSet<TDestination, TKey>> DoTransform(
+        ChangeAwareCache<TransformedItemContainer, TKey> cache, IChangeSet<TSource, TKey> changes)
+    {
```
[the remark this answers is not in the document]
JakenVeina: I rather agree about this not being DD's responsibility, but from the perspective of " we're calling `.Merge()` so, let's expose the optimization parameter that `.Merge()` has, I think it makes sense.

src/DynamicData/Cache/Internal/TransformAsync.cs
```diff
@@ -3,71 +3,60 @@
...
+        return changes.Select(change => Observable.Defer(() => Transform(change).ToObservable()))
```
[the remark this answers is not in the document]
JakenVeina: Good eye on this.

To me, this seems like a big indicator that something nonsensical is being done here. It seems like the only point of converting each change to its own observable is to be able to `.Merge()` them together, with the maxConcurrent parameter, yes? But that accomplishes nothing if we then do `.ToArray()` on the whole thing. That's gonna block the whole thread, is it not?

src/DynamicData/Cache/Internal/TransformAsync.cs
JakenVeina: Agreed, I believe `Defer()` is redundant here.

RolandPheasant: Defer is necessary, otherwise the task is invoked immediately, which would result in the merge limiter having no effect as it would be the results being limited not the actual tasks.

JakenVeina: No, `Observable.FromAsync()` already works that way. The async function returning `Task` (I.E. `Transform`) is not invoked until the observable is subscribed to.

Observe:

```cs
var whenCompletedSource = new TaskCompletionSource<int>();

var whenCompleted = Observable.FromAsync(() =>
{
    Console.WriteLine("Observable.FromAsync()");
    return whenCompletedSource.Task;
});

Console.WriteLine("whenCompleted.Subscribe()");
whenCompleted.Subscribe(x => Console.WriteLine($"whenCompleted.OnNext({x})"));

Console.WriteLine("whenCompletedSource.SetResult(7)}");
whenCompletedSource.SetResult(7);
```

src/DynamicData/Cache/Internal/TransformAsync.cs
JakenVeina: > it enables changes to be made without breaking things

Oooh, an excellent point.

src/DynamicData/Cache/Internal/TransformAsync.cs
RolandPheasant: Your example is for a single invocation. What I'd be interested to see is to try and limit say a 1000 tasks in a way that that only 4 of them would run concurrently

dwcullop: #### TL;DR

The function passed to `Observable.FromAsync` gets run once per subscription but not until the subscription happens (like `Defer`) and `Merge(x)` does not subscribe to the X+1-th item until one of the previous streams completes.

You need `Defer` if you invoke a method to get a task and then convert to an observable, but I think `Defer` plus `ToObservable` is semantically the same as `Observable.FromAsync`, so coder's choice, but I do feel like `Observable.FromAsync` is clearer.

#### Code
I used this test... It's only 20, not 1000, but I don't think that changes anything...

```csharp
async Task Main()
{
    // These two seem to be the same
    var obsAsync = Observable.FromAsync(AsyncThing);
    // var obsAsync = Observable.Defer(() => AsyncThing().ToObservable());

    var obs = Observable.Range(0, 20).Select(n => obsAsync.Spy($"Async #{n}"));
    
    // Only 3 AsyncThings at a time
    await obs.Merge(3);
    
    // Let them all fly
    //await obs.Merge();
}

static int current = 0;
static int counter = 0;

async Task<int> AsyncThing()
{
    int n = Interlocked.Increment(ref current);
    Console.WriteLine($"Starting AsyncThing #{n} ({Interlocked.Increment(ref counter)})");
    
    try
    {
        await Task.Delay(100);
        return n;
    }
    finally
    {
        Console.WriteLine($"Ending AsyncThing #{n} ({Interlocked.Decrement(ref counter)})");
    }
}
```

`AsyncThing` is invoked once per subscription.  The Observable is created, but the async method is not invoked until the subscribe.

With `Merge()`, `counter` gets up to 20 because they're all running at once.  With `Merge(3)`, then `counter` never gets bigger than 3.

src/DynamicData/Cache/Internal/TransformAsync.cs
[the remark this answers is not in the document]
dwcullop: > I rather agree about this not being DD's responsibility, but from the perspective of " we're calling `.Merge()` so, let's expose the optimization parameter that `.Merge()` has", I think it makes sense.

I guess if the logic is in the middle of the operator, there aren't many good options for allowing consumers to customize it.

src/DynamicData/Cache/Internal/TransformAsync.cs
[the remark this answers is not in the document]
dwcullop: No, it won't block the thread. `ToArray` returns an `IObservable<T[]>` that fires once (when the input sequence has completed).  There's no need to block, you just don't get events downstream until the all the `Merge`ed streams have completed.

I think it does make sense.  It's just a new technique to me.

src/DynamicData/Cache/Internal/TransformAsync.cs
RolandPheasant: Thanks for looking into this. I add the overload and some testing soon.

RolandPheasant: @JakenVeina I can confirm that Observable,FromAsync() works like the Defer as you suggested.  I was not willing to change what I had until I had a test.  I get similar run times as previously.

image

I have a question for you against the test.

src/DynamicData.Tests/Cache/TransformAsyncFixture.cs
```diff
@@ -207,6 +205,84 @@ public void Update()
...
+        results.Data.Lookup("SomeOne").Value.AgeGroup.Should().Be("Adult");
+
+    }
+
+    
+    [Theory, InlineData(10), InlineData(100)]
+
+    public async Task WithMaxConcurrency(int maxConcurrency)
+    {
+        /* We need to test whether the max concurrency has any effect.
```
RolandPheasant, line 262: @JakenVeina any ideas how I can test concurrency apart from looking at the timings?

src/DynamicData/Binding/TransformAsyncOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// Options with WithTransformOnRefresh = true.
+    /// </summary>
+    public static readonly TransformAsyncOptions WithTransformOnRefresh = new(TransformOnRefresh: true);
+
+    /// <summary>
+    /// Specify maximum concurrency only.
+    /// </summary>
+    /// <param name="maximumConcurrency">The maximum concurrency.</param>
+    /// <returns>A TransformAsyncOptions object.</returns>
+    public static TransformAsyncOptions WithMaximumConcurrency(int maximumConcurrency) => new(maximumConcurrency);
```
dwcullop, line 23: Consider just making a "Default" static instance and then let people use the `with` syntax instead of having a factory function for each property, which might get messy when there are a bunch of them.

```csharp
var options = TransformAsyncOptions.Default with { TransformOnRefresh = true };
```

PR #822, opened 2024-01-05
Fixed ambiguous overloads for `ObservableCacheEx.ExpireAfter()` and `ObservableListEx.ExpireAfter()`.

JakenVeina:
Also moved `EnsureUniqueKeys()` into proper order.

Currently, the following code produces an ambigous method error at compile-time, for the call to `.ExpireAfter()`.

```cs
var source = new SourceCache<string, int>(x => x.GetHashCode());

var result = source.ExpireAfter(_ => TimeSpan.FromSeconds(1));
```

Since any fix for this would be a breaking change, I've implemented it by just consolidating all of the overloads into one method, for simplicity.

The issue itself is relatively minor, as it can be easily worked-around by doing...

```cs
var source = new SourceCache<string, int>(x => x.GetHashCode());

var result = source.ExpireAfter(
    timeSelector: _ => TimeSpan.FromSeconds(1),
    interval: null);
```

or

```cs
var source = new SourceCache<string, int>(x => x.GetHashCode());

var result = source.ExpireAfter(
    timeSelector: _ => TimeSpan.FromSeconds(1),
    scheduler: null);
```

Since this is a breaking change with low impact, I would propose NOT merging this until we already have a major-version change queued up, due to other issues or enhancements.

### DIFF
--- a/src/DynamicData.Benchmarks/Cache/ExpireAfter_Cache_ForSource.cs
+++ b/src/DynamicData.Benchmarks/Cache/ExpireAfter_Cache_ForSource.cs
@@ -67,8 +67,8 @@ public class ExpireAfter_Cache_ForSource
 
         using var subscription = source
             .ExpireAfter(
-                timeSelector:   static item => item.Lifetime,
-                interval:       null)
+                timeSelector:       static item => item.Lifetime,
+                pollingInterval:    null)
             .Subscribe();
 
         PerformRandomizedEdits(source);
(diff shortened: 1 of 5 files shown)


dwcullop:
Even though it's a breaking change, I think this is a good thing.  Waiting until a major release mitigates the breaking change somewhat...  Thanks for this.

JakenVeina:
Turns out the list version of `ExpireAfter` has the exact same issue, so I added the fix for that as well.

RolandPheasant:
@JakenVeina apart form merge conflicts, is this ready?

RolandPheasant (approved):

Issue #758, opened 2023-11-18
[Feature]: New higher-performance Filter() operators


JakenVeina:
### Describe the functionality desired 🐞

Consider the following new filtering operators, which should provide performance improvements over existing operators:

```cs
public static IObservable<IChangeSet<TObject, TKey>> Filter<TObject, TKey, TState>(
    this    IObservable<IChangeSet<TObject, TKey>>  source,
            IObservable<TState>                     state,
            Func<TState, TObject, bool>             predicate,
            bool                                    suppressEmptyChangeSets = true);

public static IObservable<IChangeSet<TItem>> Filter<TItem, TState>(
    this    IObservable<IChangeSet<TItem>>  source,
            IObservable<TState>             state,
            Func<TState, TItem, bool>       predicate,
            bool                            suppressEmptyChangeSets = true);

public static IObservable<IChangeSet<TObject, TKey>> FilterValues<TObject, TKey>(
    this    IObservable<IChangeSet<TObject, TKey>>  source,
            Func<TObject, bool>                     predicate,
            bool                                    suppressEmptyChangeSets = true);
```

Not set on the name for the third one, but it stems from a conversation with @RolandPheasant and @dwcullop in Slack. Alternatives for this would maybe be `.FilterSlim()` or `.FilterDeterministic()` or `.FilterWithoutCaching()`. Whatever the name, the goal would be to keep it consistent with other operators (like `.Transform()`) that normally keep an internal items cache, but could have versions that don't.

### The steps the functionality will provide

For the first two operators, utilizing an `IObservable<TState>` parameter, the primary improvement operator is to reduce memory allocations required by existing operators that support dynamic filtering.

The third operator would improve performance for static-filtering scenarios, where the filter predicate is relatively fast, and is known to be deterministic, I.E. collection items are immutable, at least from the perspective of filtering. Under this scenario, `.Filter()` can be implemented without the need for an internal cache of items, becoming allocation-free for all single-change ChangeSets.

This operator would not be applicable as an `ObservableList` operator, as using deterministic-filtering does not eliminate the need for an internal items cache. An internal cache is still needed to track item indexes, so downstream indexes can be adjusted for items filtered out of the source.

### Considerations

[...]

PR #823, opened 2024-01-06
Feature/stateless filtering and transforming

JakenVeina:
Implemented one of the operators described in #758 as well as an equivalent transform operator.

Also made a few housekeeping changes, but I've split those out into separate commits, so they can more-easily be reviewed and/or excluded if others don't agree.

Benchmark results, highlighting the performance of the new operators, versus existing ones:

| Method         | Mean     | Error   | StdDev  | Ratio | RatioSD | Gen0    | Allocated | Alloc Ratio |
|--------------- |---------:|--------:|--------:|------:|--------:|--------:|----------:|------------:|
| Filter         | 293.3 μs | 5.82 μs | 7.78 μs |  1.00 |    0.00 | 76.1719 | 311.27 KB |        1.00 |
| FilterByValues | 229.8 μs | 3.16 μs | 2.80 μs |  0.78 |    0.02 | 57.3730 | 235.05 KB |        0.76 |

| Method            | Mean     | Error   | StdDev   | Median   | Ratio | RatioSD | Gen0     | Gen1    | Allocated | Alloc Ratio |
|------------------ |---------:|--------:|---------:|---------:|------:|--------:|---------:|--------:|----------:|------------:|
| Transform         | 457.1 μs | 8.79 μs | 18.15 μs | 448.8 μs |  1.00 |    0.00 | 153.3203 | 25.3906 | 627.89 KB |        1.00 |
| TransformByValues | 261.1 μs | 5.13 μs |  4.28 μs | 260.8 μs |  0.55 |    0.03 |  57.1289 |       - | 235.04 KB |        0.37 |


dwcullop:
I'm in favor of these additions, but I do have a small issue with the names.  `FilterByValues` is confusing.  `Value**s**` makes it seem like each one is going to be filtered by multiple values.  At the very least, it should be singular `FilterByValue` but I think that `ValueFilter` would be an even better choice.

JakenVeina:
I'm not sold on the names, either. Other options include `FilterDeterministic`,  `FilterSlim`, `FilterValues`. I'm gonna say `ValueFilter` is a no-go, for me, because there's a lot of discoverability value in having it start with `Filter`, so it shows up right alongside the other `Filter` operators.

Regarding documentation, is the documentslation I already added insufficient for you?

dwcullop:
> I'm not sold on the names, either. Other options include `FilterDeterministic`, `FilterSlim`, `FilterValues`. I'm gonna say `ValueFilter` is a no-go, for me, because there's a lot of discoverability value in having it start with `Filter`, so it shows up right alongside the other `Filter` operators.

Agreed... Although Intellisense finds anything with "Filter" in the name, the things starting with "Filter" are shown first.  So... No `ValueFilter` and probably not anything with "Value" because that word has special meaning in DotNet and, to some, it may imply it somehow uses a stack-based value type (ala `ValueTask`), which is not necessarily the case at all.

Of the remaining choices, I like `FilterSlim`.  `FilterDeterministic` is okay, I guess, but it's rather long and it isn't immediately clear to what the "Deterministic" refers.  Other ideas:
- `FilterInline`
- `FilterChangeSet` (because it just filters the changeset without the intermediate cache)
- `FilterFilterFilter` (okay, I guess I'm out of ideas for now)

I feel like it might be better to get the opinions of someone who isn't familiar with DD to find a name that people can know what it does intuitively.  My intuition is ruined by familiarity.  Maybe we could take a survey in the Rx slack?

> Regarding documentation, is the documentslation I already added insufficient for you?

Yes, your documentslation is fine, but the XML comments might not be enough for everyone to absorb it.  So, a Wiki article would be helpful so that more people will see it (different people discover things differently).

JakenVeina:
In a perfect world, I'd say it should be `Filter()` with all the OTHER overloads being `FilterWithCaching()`, so maybe `FilterWithoutCaching()` as another option?

dwcullop:
> In a perfect world, I'd say it should be `Filter()` with all the OTHER overloads being `FilterWithCaching()`, so maybe `FilterWithoutCaching()` as another option?

Of the names we've examined so far, my favorite is `FilterSlim`.  It isn't long too long and it's clear it does something less than the other `Filter`.

dwcullop:
Maybe your PR title has the right of it... `FilterStateless`?  I dunno.

src/DynamicData/Cache/Internal/TransformByValues.cs
```diff
@@ -0,0 +1,59 @@
+﻿// Copyright (c) 2011-2023 Roland Pheasant. All rights reserved.
+// Roland Pheasant licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using System.Reactive;
+using System.Reactive.Linq;
+
+namespace DynamicData.Cache.Internal;
+
+internal sealed class TransformByValues<TDestination, TSource, TKey>
```
RolandPheasant, line 10: Is this the same a Convert?

See https://github.com/reactivemarbles/DynamicData/blob/main/src/DynamicData/Cache/ObservableCacheEx.cs#L1014.


Maybe that could be resurrected?

src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -4764,6 +4791,34 @@ sortOrder switch
...
+    /// If either the predicate or the items itemselves have any mutable state, mutating it will result in undefined and potentially incorrect behavior of the result stream.
+    /// This operator also assumes that the transformation factory itself is relatively high-performance, as it will be invoked repeatedly, and its results are not internally cached.
+    /// </summary>
+    /// <typeparam name="TDestination">The type of collection items produced by the transformation.</typeparam>
+    /// <typeparam name="TSource">The type of collection items to be transformed.</typeparam>
+    /// <typeparam name="TKey">The type of the key values of each collection item.</typeparam>
+    /// <param name="source">The source stream of collection items to be transformed.</param>
+    /// <param name="transformFactory">The transformation to be applied to each item.</param>
+    /// <returns>A stream of collection changesets where upstream collection items are transformed by the given factory function.</returns>
+    public static IObservable<IChangeSet<TDestination, TKey>> TransformByValues<TDestination, TSource, TKey>(
```
RolandPheasant, line 4806: I wonder whether we can achieve this overload using an optimizer.   For example Sort accepts SortOptimisations which amongst other options allows the consumer to specify SortOptimisations.ComparesImmutableValuesOnly.

If something similar was used for Transform, or for Filter we could use different code paths in the background and eliminate the need to new named overloads.

RolandPheasant (changes requested):
I've been sat on this for a while so I could think long and hard about it and I think using optimiser options (see my comments against the code) could be a good solution. 

The only draw back would be yet more optional overloads, unless of course we introduce an overload which accepts the transform / filter functions + a record containing all the additional options.  This would be similar to my suggestion on https://github.com/reactivemarbles/DynamicData/pull/819 and also similar to the new binding options.

src/DynamicData/Cache/ObservableCacheEx.cs
JakenVeina: I could probably get behind that, in theory, but part of the problem is that this optimization is ONLY applicable for certain scenarios of filtering and transforming. Like, any of the overloads that take `IObservable<>` of something to control filtering flat-out cannot use this optimization.

So, as an example, if we were to define `FilterOptimizations.DontCacheItems`, we'd have to do one of a few things:

1. Only define one overload to take a `FilterOptimizations` paramter, the one where the filter predicate is just `Func<T, bool>`. This would kinda make the enum pointless, cause then we wouldn't be able to add a `FilterOptimizations` parameter to any other overloads, so any optimizations we might want for those would have to be in a separate enum.
2. Throw an exception if you try and specify `FilterOptimizations.DontCacheItems` with an overload that supports dynamic filtering, E.G. `IObservable<Func<T, bool>>`.
3. Just silently have `FilterOptimizations.DontCacheItems` do nothing for overloads that support dynamic filtering, leading users to think they're getting a performance boost when they're not.

On a separate note (and this comes with a grain of salt, because I have not done a DEEP dive into it, just seen it on the surface) my current impression of how sorting works in DynamicData is that I don't really like it. I don't like how much seemingly-unnecessary stuff there is inside `ISortedChangeSet<>`, and `SortOptimizations` is part of that. I have found that the "extra" stuff in `ISortedChangeSet<>` makes it impossible to implement some of the most common querying scenarios supported by LINQ. All that being said, I'm initially very wary about the idea of adding `FilterOptimizations` in the same manner as `SortOptimizations`.

src/DynamicData/Cache/Internal/TransformByValues.cs
JakenVeina: Uhhh, yep, that looks pretty much the same, save for a few technical optimizations.

I think it's worth considering that `.Convert()` has a bit of a discoverability problem, then. So, we could consider options for addressing that:

1. Add `TransformByValues()` (or whatever name we agree on) anyway, and have it and `.Convert()` both call the same implementation behind the scenes.
2. #1 but also mark `.Convert()` as Obsolete, to be removed some day.
3. Just try and address it with some new Wiki content, as Darrin already suggested. In which case, I would still swap `.Convert()` over to the implementation I wrote here.

RolandPheasant:
@JakenVeina regarding naming just go with your gut - dealers choice  

One more suggestion however is maybe FilterImmutable / TransformImmutable after all there is a GroupByImmutable which although has a slightly different behaviour it may provide consistency.

Let me know when this is ready and let's merge.

JakenVeina:
If `.GroupByImmutable()` already exists, then that sounds like a winner.

And my gut says `.TransformImmutable()` fits better than `.Conver()` for the same reasons, so I think I'l go with the "alias for now, deprecate for later" approach for that one.

dwcullop:
I can deal with the `Immutable` names.  `TransformImmutable` / `FilterImmutable`, etc.

JakenVeina:
@RolandPheasant I've updated namings, as discussed, and re-wrote the documentation, after some feedback from @Insire.

Since the older `.Convert()` operator is already marked as obsolete, I left it as-is, to be removed eventually.

Insire:
I'm quite happy with the new xml comments, which was a major point of discussion i had with @JakenVeina

RolandPheasant (approved):

PR #824, opened 2024-01-06
Feature:  TransformMany Async

dwcullop:
## Description

Implements `TransformManyAsync` and `TransformManySafeAsync` for #461.
Creates a new operator implementation that is based on the implementation of `MergeManyChangeSets` but allows the child selector to be Async, allows for an optional error handler, and does "best effort" to try to minimize the number of changesets.

`TransformMany` will emit a single changeset with all of the added child values for each parent changeset.  This is not ideal for the async version because each parent value added creates an async call.  This would mean that no changeset could be emitted until all of the async operations have completed.

As a compromise, the new operator will use an atomic counter to determine how many async operations have completed (but not been processed) and will not emit a changeset until all completed operations have been processed.  This changeset will be emitted even if there are async operations pending because there is no way to tell how they will take.

This should be a good balance between timely changesets and fewer changesets.


src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -4832,6 +4834,313 @@ sortOrder switch
         where TSource : notnull
         where TSourceKey : notnull => new TransformMany<TDestination, TDestinationKey, TSource, TSourceKey>(source, manySelector, keySelector).Run();
 
+    /// <summary>
```
RolandPheasant, line 4837: I wonder whether we should break this class into a series of partial classes as it's becoming huge.

RolandPheasant (approved):

src/DynamicData/Cache/ObservableCacheEx.cs
dwcullop: That's not a bad idea... But if we do it, it should be in dedicated PR.

PR #829, opened 2024-01-08
Feature:  Improve Performance of MergeManyChangeSets with fewer changesets

dwcullop:
## Description

This PR improves the performance of the various `MergeManyChangeSet` operators by reducing the number of changeset that are emitted.  Instead of one changeset per change in child changesets, the changes are held until they've all been processed, and the result is then emitted as a single changeset.

### Impact

All changes from a parent changeset will result in a single downstream changeset.  If it adds 2 items and removes 3 others, then the downstream will see a single changeset that adds of the child items from the added parent items and removes all of the child items from the 3 removed parent items.

Updates will result in a single changeset with removal of old child items/adding of new child items.

## Logic Overview

By setting a flag before handling any changes to the parent changeset, we can detect if changes to the child changeset are due to parent changes (add/remove) we can avoid emitting them.  At the end of handling a parent changeset, all of the changes can be emitted at once.

When the flag is not set, then the child changeset are not the result of parental changes, and they can be emitted immediately.

## Other Changes

Creates an overload for `SubscribeSafe` extension method that enables it to be invoked with Action objects instead of just `IObserver<T>`.

## Testing

- Updated unit tests to expect fewer changesets as appropriate
- Added a new test to confirm clearing the parent cache results in a single changeset removing all the children

### DIFF
--- a/src/DynamicData.Tests/Cache/MergeManyChangeSetsCacheFixture.cs
+++ b/src/DynamicData.Tests/Cache/MergeManyChangeSetsCacheFixture.cs
@@ -230,7 +230,7 @@ public sealed class MergeManyChangeSetsCacheFixture : IDisposable
         _marketCacheResults.Data.Count.Should().Be(MarketCount);
         markets.Sum(m => m.PricesCache.Count).Should().Be(MarketCount * PricesPerMarket);
         results.Data.Count.Should().Be(MarketCount * PricesPerMarket);
-        results.Messages.Count.Should().Be(MarketCount);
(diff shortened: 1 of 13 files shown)


RolandPheasant (approved):
Smart optimisation

dwcullop:
> Smart optimisation

I had wanted this behavior before, but it wasn't obvious how...   When I started digging into `TransformMany`, I realized someone (you, presumably) went to a lot of trouble to minimize the number of changesets and that I should really go back and figure it out so the behavior is consistent.

Issue #672, opened 2022-12-22
[Feature]: New Transform-like Query operator that is working with an IObservable-transformer


KeilFelix:
### Background

This is a follow up feature request on #409 with a **(hopefully) new and powerful idea**. 🤞 
Every time I use DynamicData I am missing query operators that are working on IObservable like the FilterOnObservable. My TSource-objects normally have IObservable properties, and I think that this is a natural reactive design pattern.

### Idea

Basically a new query operator with the following signature _(exemplary only for ObservableList)_

`IObservable<IChangeSet<TDestination>> TransformOnObservable<TSource, TDestination>(this IObservable<IChangeSet<TSource>> source, Func<TSource, IObservable<TDestination>> transformFactory)`

#### Desired behavior
The operator should project a list where each item is the latest item from each sources' IObservable<TDestination> and should be automatically replaced if a new item is received. As long as an Observable does not provide any value there should be no entry in the list. If the Observable completes the item should be removed.

#### Possible other names

- TransformMany-overload
- TransformLatest
- TransformSwitch

#### Alternative signature

`IObservable<IChangeSet<TDestination>> TransformObservable<TDestination>(this IObservable<IChangeSet<IObservable<TDestination>>> source)`

This signature would be sufficient because a normal Transform could be done beforehand.

### Exemplary Usage
With this operator it would be easy to reimplement FilterOnObservable or a Sort/Group:

```
public static IObservable<IChangeSet<TObject>> FilterOnObservableVariant<TObject>(this IObservable<IChangeSet<TObject>> source, Func<TObject, IObservable<bool>> objectFilterObservable)
{
    return source
        .TransformOnObservable(obj => objectFilterObservable(obj).Select(filter => (Filter: filter, Object: obj)))
        .Filter(t => t.Filter)
        .Transform(t => t.Object);
}
```
_Note to above sample: Maybe a StartValue should be prepended and  a never ending Observable appended.
This would match the original behavior. But on the other side it is more natural without it and it can be easily done by the user when creating the objectFilterObservable-Func_

### Prototype Implementation
I came up with a draft and the FilterOnObservableFixture passes with this!

```
public static IObservable<IChangeSet<TDestination>> TransformOnObservable<TSource, TDestination>(
    this IObservable<IChangeSet<TSource>> source, Func<TSource, IObservable<TDestination>> transformFactory)
{
    return source.Transform(transformFactory).TransformObservable();
}

private class ObservableObjectSelector<TObject>
{
[...]

PR #841, opened 2024-01-29
Feature:  TransformOnObservable Operator for SourceCache

dwcullop:
## Description

As requested by #672, this creates a new operator `TransformOnObservable` that will transform each item in a Cache ChangeSet into an Observable and use the latest value from that Observable as the value the corresponding key in the resulting change set.

Uses an update counter to minimize the number of changesets emitted downstream.  If the source emits a changeset with 5 add events, and each of the sub-observables emits a value upon subscription, then the downstream should see a single changeset with 5 add events.

### Notes
- The Key will not be observed until the new Observable emits at least one value.
- Subsequent values emitted from the Observable will show up as `Update` changes for the same Key value.
- If the Key is removed from the source, the removal will be reflected downstream
- If the source completes and all of the sub-observables completes, then the result observable changeset will also complete.

## Bonus Changes
- Improvements to `OnItem*` operators:
  - Added overloads that use an Action that will take the key as a parameter
  - Updated all to a common implementation (optimized)
  - Changed `OnItemRemoved` to not use the special class unless the "invokeOnUnsub" flag is set because it uses an extra Cache and lock that isn't needed

### DIFF
--- a/src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet8_0.verified.txt
+++ b/src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet8_0.verified.txt
@@ -1585,13 +1585,25 @@ namespace DynamicData
         public static System.IObservable<DynamicData.IChangeSet<TObject, TKey>> OnItemAdded<TObject, TKey>(this System.IObservable<DynamicData.IChangeSet<TObject, TKey>> source, System.Action<TObject> addAction)
             where TObject :  notnull
             where TKey :  notnull { }
+        public static System.IObservable<DynamicData.IChangeSet<TObject, TKey>> OnItemAdded<TObject, TKey>(this System.IObservable<DynamicData.IChangeSet<TObject, TKey>> source, System.Action<TObject, TKey> addAction)
+            where TObject :  notnull
+            where TKey :  notnull { }
         public static System.IObservable<DynamicData.IChangeSet<TObject, TKey>> OnItemRefreshed<TObject, TKey>(this System.IObservable<DynamicData.IChangeSet<TObject, TKey>> source, System.Action<TObject> refreshAction)
+            where TObject :  notnull
+            where TKey :  notnull { }
+        public static System.IObservable<DynamicData.IChangeSet<TObject, TKey>> OnItemRefreshed<TObject, TKey>(this System.IObservable<DynamicData.IChangeSet<TObject, TKey>> source, System.Action<TObject, TKey> refreshAction)
             where TObject :  notnull
             where TKey :  notnull { }
         public static System.IObservable<DynamicData.IChangeSet<TObject, TKey>> OnItemRemoved<TObject, TKey>(this System.IObservable<DynamicData.IChangeSet<TObject, TKey>> source, System.Action<TObject> removeAction, bool invokeOnUnsubscribe = true)
             where TObject :  notnull
             where TKey :  notnull { }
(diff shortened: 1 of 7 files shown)


src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet8_0.verified.txt
```diff
@@ -1931,6 +1931,14 @@ namespace DynamicData
             where TSource :  notnull
             where TSourceKey :  notnull
             where TCollection : System.Collections.Specialized.INotifyCollectionChanged, System.Collections.Generic.IEnumerable<TDestination> { }
+        public static System.IObservable<DynamicData.IChangeSet<TDestination, TKey>> TransformOnObservable<TSource, TKey, TDestination>(this System.IObservable<DynamicData.IChangeSet<TSource, TKey>> source, System.Func<TSource, System.IObservable<TDestination>> transformFactory)
```
RolandPheasant, line 1934: On the latest main branch,  I have got rid of testing for Net6 and Net7 so we'll be able to get rid of a couple of the verification results.

src/DynamicData/Cache/Internal/TransformOnObservable.cs
```diff
@@ -0,0 +1,61 @@
...
+    public IObservable<IChangeSet<TDestination, TKey>> Run() => Observable.Create<IChangeSet<TDestination, TKey>>(observer =>
+    {
+        var locker = new object();
+        var cache = new ChangeAwareCache<TDestination, TKey>();
+        var updateCounter = 0;
+
+        // Helper to emit any pending changes when all the updates have been handled
+        void EmitChanges()
+        {
+            if (Interlocked.Decrement(ref updateCounter) == 0)
```
RolandPheasant, line 26: I think the concept of an update counter to batch results is inspired.  The same concept would probably benefit several other operations.

RolandPheasant (approved):

RolandPheasant:
Doh, I've approved but I just noticed a test is failing,.

src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet8_0.verified.txt
dwcullop: No wonder I couldn't find them and had to edit those files by hand!  😆  I can roll them back.

src/DynamicData/Cache/Internal/TransformOnObservable.cs
dwcullop: > I think the concept of an update counter to batch results is inspired. The same concept would probably benefit several other operations.

Thank you!  :blush:
I used it for `TransformManyAsync` to get behavior similar to `TransformMany` after you stressed the importance of minimizing the number of ChangeSets.  I was also thinking that it might be beneficial to use in other places, so I'll keep a lookout for those.

There is some nuance to using it.  For example, you have to increment the counter before the `Synchronize` so that updates coming in on other threads are counted and the emission downstream is deferred until it can be handled.

There are also some downsides:
- More difficult (sometimes impossible) to deterministically know how many changesets you'll get
- Possible to never get a downstream changeset (if updates happen constantly)

dwcullop: Thank you!  :blush:
I used it for `TransformManyAsync` to get behavior similar to `TransformMany` whenever possible.  I was also thinking that it might be beneficial to other places.  Perhaps it warrants an abstraction that makes it easier to integrate...  But then again, perhaps not.

dwcullop:
> Doh, I've approved but I just noticed a test is failing,.

Yeah, I changed something to use `Lookup` which is returning an `Optional<string>` and they don't automatically compare to a `string`... Easy fix.

Issue #848, opened 2024-02-10
[Feature]: Group Operator with a Dynamic Grouping Key Selector


dwcullop:
### Describe the functionality desired

A Grouping operator that supports the Group Key Selector changing dynamically, such as:

```csharp
IObservable<IGroupChangeSet<TObject, TKey, TGroupKey>> Group(
    this IObservable<IChangeSet<TObject, TKey>> source,
    IObservable<Func<TObject, TKey, TGroupKey>> grouperObservable);

// Also maybe
IObservable<IGroupChangeSet<TObject, TKey, TGroupKey>> Group(
    this IObservable<IChangeSet<TObject, TKey>> source,
    IObservable<Func<TObject, TKey, TGroupKey>> grouperObservable,
    IObservable<Unit> regrouper);
```

### The steps the functionality will provide

When `groupObservable` fires, it will regroup all the items according to the new Grouping Key selector function.

If `regrouper` is provided, it will re-evaluate the grouping using the current Grouping Key.

### Considerations

Current Group operators don't support changing of the Group Key selector.  The Group Key selector can be dynamically re-applied, but it can't be changed.
```csharp
IObservable<IGroupChangeSet<TObject, TKey, TGroupKey>> Group<TObject, TKey, TGroupKey>(
this IObservable<IChangeSet<TObject, TKey>> source, Func<TObject, TGroupKey> groupSelectorKey, IObservable<Unit> regrouper);
```

`GroupOnObservable` allows for dynamic changing of the Group Key Value but on a per item basis.

PR #849, opened 2024-02-11
Feature:  Group by Dynamic Selector Function

dwcullop:
## Description

Addresses #848 by providing new overloads for the `Group` operator that allow the user to supply a `IObservable<Func<TObject, TKey, TGroupKey>>` so that the function that is used to obtain the GroupKey can change dynamically.  When a new value is supplied, a new GroupKey will be obtained for each item, and if the GroupKey has changed, the item will be moved to the new Group.

Also allows for an optional `IObservable<Unit>` parameter that will force the current GroupKey selector to be re-applied to all the items.

## Issues
This operator also suffers from #851.  I will submit another PR that fixes both operators at the same time.

### DIFF
--- a/src/DynamicData.Tests/Cache/GroupOnDynamicFixture.cs
+++ b/src/DynamicData.Tests/Cache/GroupOnDynamicFixture.cs
@@ -1,18 +1,18 @@
 ﻿using System;
 using System.Linq;
+using System.Reactive;
 using System.Reactive.Linq;
+using System.Reactive.Subjects;
 using System.Threading.Tasks;
 
 using Bogus;
-using DynamicData.Tests.Domain;
-using DynamicData.Binding;
 using DynamicData.Kernel;
+using DynamicData.Tests.Domain;
+using DynamicData.Tests.Utilities;
 using FluentAssertions;
 using Xunit;
 
 using Person = DynamicData.Tests.Domain.Person;
-using System.Reactive.Subjects;
-using System.Reactive;
 
 namespace DynamicData.Tests.Cache;
 
@@ -36,7 +36,6 @@ public class GroupOnDynamicFixture : IDisposable
     private readonly Randomizer _randomizer;
     private readonly Subject<Func<Person, string, string>> _keySelectionSubject = new ();
     private readonly Subject<Unit> _regroupSubject = new ();
-    private readonly IDisposable _cleanup;
     private Func<Person, string, string>? _groupKeySelector;
 
     public GroupOnDynamicFixture()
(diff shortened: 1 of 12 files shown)


src/DynamicData.Tests/Cache/GroupOnDynamicFixture.cs
```diff
@@ -0,0 +1,313 @@
...
+using FluentAssertions;
+using Xunit;
+
+using Person = DynamicData.Tests.Domain.Person;
+using System.Reactive.Subjects;
+using System.Reactive;
+
+namespace DynamicData.Tests.Cache;
+
+public class GroupOnDynamicFixture : IDisposable
```
JakenVeina, line 19: Should probably add tests for Refresh and Move changesets, for completeness, whatever the expected behavior is.

src/DynamicData.Tests/Cache/MergeManyChangeSetsCacheSourceCompareFixture.cs
```diff
@@ -52,7 +52,7 @@ public MergeManyChangeSetsCacheSourceCompareFixture()
     [Theory]
     [InlineData(5, 7)]
     [InlineData(10, 50)]
-#if !DEBUG
+#if false && !DEBUG
```
JakenVeina, line 55: This seems like it was a workaround for debugging?

src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -6457,6 +6513,11 @@ void UpdateAction(IChangeSet<TObject, TKey> updates)
         });
     }
 
+    private static Func<TObject, TKey, TResult> AdaptSelector<TObject, TKey, TResult>(Func<TObject, TResult> other)
```
JakenVeina, line 6516: This seems a little silly as a dedicated helper function, if it's only being used once.

JakenVeina (changes requested):

src/DynamicData.Tests/Cache/GroupOnDynamicFixture.cs
JakenVeina: Also, since this is an operating receiving multiple observable inputs, there should be thread-safety testing.

src/DynamicData.Tests/Cache/MergeManyChangeSetsCacheSourceCompareFixture.cs
dwcullop: There is some threading issue in the tests... I thought it was resolved, but I guess not because it fails intermittently, so I'm just disabling for now because it's out of scope for this PR.

I think it is related to the threading issue we dealt recently that turned out to be in the test.

src/DynamicData.Tests/Cache/GroupOnDynamicFixture.cs
```diff
@@ -0,0 +1,313 @@
...
+
+        // Check each group
+        expectedGroupings.ForEach(grouping => grouping.Should().BeEquivalentTo(groupResults.Groups.Lookup(grouping.Key).Value.Data.Items));
+    }
+
+    private void ForceRegroup() => _regroupSubject.OnNext(Unit.Default);
+
+    private void GroupByFavColor() => _keySelectionSubject.OnNext(FavColor);
+
+    private void GroupByParentName() => _keySelectionSubject.OnNext(ParentName);
```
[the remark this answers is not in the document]
dwcullop: It might be useful.  I'll use it in the stress test.

src/DynamicData/Cache/ObservableCacheEx.cs
dwcullop: I was thinking we could use it everywhere we have to adapt `(obj, key) => func(obj)`.  It would make the overloads that just exist to adapt the selector easier to see, but might be overkill.

src/DynamicData.Tests/Cache/GroupOnDynamicFixture.cs
dwcullop: > Should probably add tests for Refresh and Move changesets, for completeness, whatever the expected behavior is.

Move operations are not supported.  I guess they could just be forwarded.

Refresh testing is a good idea.

dwcullop: I added tests for Refresh events and some multithreaded stress tests that will fire all 3 observables independently from different threads a bunch of times and then verifies the results at the end.

src/DynamicData/Cache/ObservableCacheEx.cs
JakenVeina: Something like this, I'd leave a comment or a TODO specifying why the "normal" code was bypassed.

src/DynamicData.Tests/Cache/GroupOnDynamicFixture.cs
JakenVeina: > Move operations are not supported.

Right, so the expected behavior should be that the operator ignores `Move` changes, for completeness' sake we should test that explicitly.

> fire all 3 observables independently from different threads a bunch of times and then verifies the results at the end

Sounds perfect.

JakenVeina (approved):

src/DynamicData/Cache/ObservableCacheEx.cs
dwcullop: Done.

PR #852, opened 2024-02-14
Feature: SuspendNotifications Methods for SourceCache / IntermediateCache

dwcullop:
## Description

Adds the `INotifyCollectionChangedSuspender` interface to `IntermediateCache` and `SourceCache` which adds `SuspendNotifications` and `SuspendCount` methods which will block either the ChangeSet observable or the CountChanged observable (respectively) until the IDisposable that the method returns is Disposed.

`CountChanged` will always emit the current count on subscription, but no further changes will be emitted until the suspension is complete, when it will emit the updated count (if it has changed).  `Connect` will not emit any ChangeSets upon subscription when suspended.  When the suspension is complete, it will emit a single changeset with the most recent contents and all a changeset for all future changes.

For subscriptions created BEFORE the suspension, no more changesets will be emitted until the suspension is complete, and then a single changeset, with ALL the changes that have been made will be emitted at once.

Both suspensions can be invoked multiple times and all IDisposables have to be Disposed before messages will resume.


src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -240,4 +322,103 @@ private void InvokePreview(ChangeSet<TObject, TKey> changes)
...
+        lock (_locker)
+        {
+            Debug.Assert(_suspensionTracker.IsValueCreated, "Should not be Resuming Notifications without Suspend Count instance");
+            _suspensionTracker.Value.ResumeNotifications();
+        }
+    }
+
+    private class SuspensionTracker(Action<ChangeSet<TObject, TKey>> onResumeNotifications, Action onResumeCount) : IDisposable
+    {
+        private readonly BehaviorSubject<bool> _notifySuspendSubject = new(false);
```
JakenVeina, line 357: Poor name for this field, IMO. This is just the flag indicating whether notifications are suspended? Except it's a `BehaviorSubject<T>` so it can be subscribed to? It should be `_areNotificationsSuspended` then. Better yet, it could be combined with the actual count variable, `_notifySuspendCount` and then you've reduced the size of the whole `SuspensionTracker` object.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -240,4 +322,103 @@ private void InvokePreview(ChangeSet<TObject, TKey> changes)
...
+
+        private readonly Action _onResumeCount = onResumeCount;
+
+        private IEnumerable<Change<TObject, TKey>> _pending = Enumerable.Empty<Change<TObject, TKey>>();
+
+        private int _countSuspendCount;
+
+        private int _notifySuspendCount;
+
+        public bool CountSuspended => _countSuspendCount > 0;
```
JakenVeina, line 369: Suggest `IsCountSuspended` for the name.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -240,4 +322,103 @@ private void InvokePreview(ChangeSet<TObject, TKey> changes)
...
+
+        public bool NotificationsSuspended => _notifySuspendCount > 0;
```
JakenVeina, line 371: Suggest `AreNotificationsSuspended` for the name.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -240,4 +322,103 @@ private void InvokePreview(ChangeSet<TObject, TKey> changes)
...
+
+        public IObservable<bool> NotificationsSuspendedStatus => _notifySuspendSubject;
```
JakenVeina, line 373: Suggest `AreNotificationsSuspended` for the name, which by no coincidence collides with the name I suggested for the plain flag above. That suggests that the two ought to be combined as a `BehaviorSubject<T>` or a read-only equivalent.

Unfortunately, I don't think such a thing (a read-only equivalent of a `BehaviorSubject<T>`) exists, but since this is a private class only, I don't think there's any real danger in just exposing the subject directly.

Hit me up in Slack if all this isn't coming across clearly.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -240,4 +322,103 @@ private void InvokePreview(ChangeSet<TObject, TKey> changes)
...
+
+        public void AddPending(IEnumerable<Change<TObject, TKey>> changes)
```
JakenVeina, line 375: VERY minor nitpick, but I would suggest `AddPendingChanges` or `EnqueueChanges` or something. Just `AddPending` immediately begs the question "pending what?" which I know is obvious when looking at the method signature, but not necessarily so much when looking at calling-code.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -240,4 +322,103 @@ private void InvokePreview(ChangeSet<TObject, TKey> changes)
...
+        {
+            Debug.Assert(changes is not null, "Don't pass in a null Enumerable");
+            Debug.Assert(NotificationsSuspended, "Shouldn't be adding pending values if notifications aren't suspended");
+            _pending = _pending.Concat(changes);
```
JakenVeina, line 379: Surely a `List<T>` that gets reused would be better than chained enumerables?

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -240,4 +322,103 @@ private void InvokePreview(ChangeSet<TObject, TKey> changes)
...
+        }
+
+        public void SuspendCount() => ++_countSuspendCount;
+
+        public void ResumeNotifications()
+        {
+            if (--_notifySuspendCount == 0 && !_notifySuspendSubject.IsDisposed)
+            {
+                // Fire pending changes to existing subscribers
+                if (_pending.Any())
```
JakenVeina, line 399: Swap to just `.Count` if you swap `_pending` to a `List<T>` as suggested above.

JakenVeina (changes requested):
Functionally, I don't see anything missing or broken here.

src/DynamicData/Cache/Internal/ReaderWriter.cs
```diff
@@ -33,9 +33,7 @@ public TObject[] Items
         {
             lock (_locker)
             {
-                var result = new TObject[_data.Count];
-                _data.Values.CopyTo(result, 0);
-                return result;
+                return [.. _data.Values];
```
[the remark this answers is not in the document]
dwcullop: The IDE flagged it as an improvement so I made it.  I can move it to another PR, but seems like more hassle than it's worth for such a small change.

src/DynamicData/Cache/ObservableCache.cs
dwcullop: I don't want to expose the Subject directly.  If I change it to Subject<int>, then I can expose `_subject.Select(x => x > 0)`.  I guess that doesn't have to allocate if I label it static.

src/DynamicData/Cache/ObservableCache.cs
dwcullop: EnqueueChanges is better.

src/DynamicData/Cache/ObservableCache.cs
dwcullop: I'm not sure.  I thought this would be better by avoiding the copy to the list.  The ChangeSet is already made and allocated.  Putting them in a list would definitely be two copies.  This is just one (when it makes the final ChangeSet).

src/DynamicData/Cache/ObservableCache.cs
dwcullop: Yeah, I don't like `Any` but there's little choice with enumerables.  Let me know what you think about avoiding the copy.

src/DynamicData/Cache/ObservableCache.cs
dwcullop: Naming is hard.  I like your suggestions though.

I originally had it as `Subject<int>`.  The reason I went with a bool Subject and an int instead of just an int subject is because then to get to the bool, you have to have a more observables in subscription chain (Filter, Select, etc.), which I thought would make it more heavy weight with their closures and allocations.

dwcullop: I also didn't want to fire the Observable every time the count changed.  I just need to fire when it goes from to 0 to 1 or 1 to 0.  I thought this was a slight optimization, instead of firing and then filtering.

I'm not too concerned the size of the object because it's own initiated if it is used.  Having the counter outside the Subject is 4 bytes.

src/DynamicData/Cache/ObservableCache.cs
dwcullop: I took another stab at the name.  I'd rather not expose the Subject directly, even in an internal class, I don't like exposing implementation details if it can be avoided.

src/DynamicData/Cache/ObservableCache.cs
dwcullop: I just went with List.  It's clearer.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: So, ultimately we're trading some extra allocations on each suspended set of changes, in favor of skipping an iteration over them, to copy them into a buffer. I'm not sure which would be more "efficient" overall. Sounds like it's worth a benchmark.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: "Size of the object" is a VERY minor concern, to be sure. Source collections are not constructed often, they're expected to have a relatively-long lifetime, compared to the items and changes they orchestrate.

Since this flag stream is subscribed to by all calls to `.Connect()` and `.Watch()`, you're probably right that it's better to have fewer notifications being published.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: > I don't want to expose the Subject directly. If I change it to Subject, then I can expose `_subject.Select(x => x > 0)`. I guess that doesn't have to allocate if I label it static.

If you were gonna do this, the best way would be to have the transformed observable itself stored and reused on the object instance, rather than having every call to the property invoke `.Select()` again, and create a new `IObservable<T>` instance.

I.E.

```cs
public SuspensionTracker()
{
    ...
    _areNotificationsSuspended = _subject.Select(...);
   ...
}

...

public IObservable<bool> AreNotificationsSuspended
    => _areNotificationsSuspended;
```

JakenVeina (approved):

Issue #854, opened 2024-02-16
[Performance]:  Don't use Scheduler.Default as the default scheduler


dwcullop:
### Performance Improvement

Define some static read-only value that represents the `IScheduler` instance to use when one isn't provided.  We can then determine the best way to that value should be initialized.

![image](https://github.com/reactivemarbles/DynamicData/assets/11876551/4d74bca0-4608-4c97-bdcf-de005dca469f)

JakenVeina:
So, we're talking about something like a `DynamicData.DefaultScheduler` property somewhere that is actually `public`, for consumers to configure however they like. And we can default it to `ThreadPoolScheduler` or whatever. That sounds reasonable to me.

One thing we'd want to consider is where we're pulling the field value dynamically, versus pulling it and caching it forever, say, within an operator. Practically speaking, I think the intent of this static field is that consumers would set it once, upon startup, but we do want to make sure that if the user changes it, we haven't already pulled and captured the value anywhere, and thus wouldn't recognize the change. Maybe we could do something fancy that would actually *detect* if you're trying to change the field after it's already started being used, and throw an exception.

dwcullop:
I wasn't thinking it would be exposed publicly.  Just something internal that we can have some logic to set to the best value and have all the internal operators use that instead of some other way of getting the default.

PR #862, opened 2024-02-20
Maintenance: Centralized Value for Default IScheduler instance

dwcullop:
## Description

Addresses #854 by defining a constant value for a default IScheduler instance and changes all instances of `Scheduler.Default` to use that value instead.

### DIFF
--- a/src/DynamicData/GlobalConfig.cs
+++ b/src/DynamicData/GlobalConfig.cs
@@ -8,5 +8,5 @@ namespace DynamicData;
 
 internal static class GlobalConfig
 {
-    public static IScheduler DefaultScheduler { get; } = TaskPoolScheduler.Default;
+    public static IScheduler DefaultScheduler => TaskPoolScheduler.Default;
 }
--- a/src/DynamicData/GlobalConfig.cs
+++ b/src/DynamicData/GlobalConfig.cs
@@ -4,9 +4,9 @@
 
 using System.Reactive.Concurrency;
 
-namespace DynamicData.Internal;
+namespace DynamicData;
 
-internal static class Defaults
+internal static class GlobalConfig
 {
-    public static IScheduler Scheduler { get; } = TaskPoolScheduler.Default;
+    public static IScheduler DefaultScheduler { get; } = TaskPoolScheduler.Default;
 }
--- a/src/DynamicData/Internal/Defaults.cs
+++ b/src/DynamicData/Internal/Defaults.cs
@@ -1,0 +1,12 @@
+﻿// Copyright (c) 2011-2023 Roland Pheasant. All rights reserved.
+// Roland Pheasant licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using System.Reactive.Concurrency;
+
+namespace DynamicData.Internal;
+
+internal static class Defaults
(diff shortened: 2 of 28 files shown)


src/DynamicData/Internal/Defaults.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) 2011-2023 Roland Pheasant. All rights reserved.
+// Roland Pheasant licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using System.Reactive.Concurrency;
+
+namespace DynamicData.Internal;
+
+internal static class Defaults
```
dwcullop, line 9: @RolandPheasant Do you think this should be a public thing so that consumers can change the value if they want to?  We could call it `DynamicData.GlobalConfig` with this as a property or something.

It does sound like a foot-gun though.  Maybe it should just be some enum that allows us to ensure it only gets set to schedulers that will work.  Then again, maybe they have a reason for doing everything on the Intermediate Scheduler...  Who can say?

src/DynamicData/GlobalConfig.cs
```diff
@@ -0,0 +1,12 @@
...
+// Roland Pheasant licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using System.Reactive.Concurrency;
+
+namespace DynamicData;
+
+internal static class GlobalConfig
+{
+    public static IScheduler DefaultScheduler { get; } = TaskPoolScheduler.Default;
```
dwcullop, line 11: @RolandPheasant Do you think this should be a public thing so that consumers can change the value if they want to?

It does sound like a foot-gun, but maybe they have a reason for doing everything on the Intermediate Scheduler...  Who can say?

src/DynamicData/Internal/Defaults.cs
RolandPheasant: I suspect in this case, making it internal would probably be sufficient and it allows up to change our mind and give the consumers the best choice out of the box.

Also when I added the binding options I accidentally added the global options class in the binding namespace. That should be moved to the root namespace. It would technically be a breaking change, but I'd suggest we should not increase the bug version as it would be correcting a mistake for a bit of code for which there's probably only one or two users.

RolandPheasant: See DynamicDataOptions here https://github.com/reactivemarbles/DynamicData/blob/main/src/DynamicData/Binding/BindingOptions.cs

dwcullop: Funny you should ask... I just moved it to the main namespace.

Are you asking me to move the BindingOptions as part of this PR?  They can't go in the same class if one is public and the other is internal.

dwcullop: I moved it to the main namespace and gave it a better name but kept it as internal.  If you want, I can move the `DynamicDataOptions` as well (but it might be better as another PR).

Then later we can discuss adding a configurable Default Scheduler value to it...

dwcullop: @RolandPheasant See #864 for those changes.

JakenVeina (approved):

PR #864, opened 2024-02-20
Maintenance:  Move DynamicDataOptions

dwcullop:
## Description
- Moved `DynamicDataOptions` to its own file
- Moved `DynamicDataOptions` to the top-level namespace


dwcullop:
@RolandPheasant If this is what you had in mind, then accept it at your leisure.  Or save it for later if you're not ready to make the breaking change.

JakenVeina (approved):

RolandPheasant:
@glennawatson  / @ChrisPulman  this is a minor break which changes the namespace of an object that was introduced in the last release.  It is correcting an error, so although technically a breaking change I do not intend to bump the major version

glennawatson:
Likely doesn't impact us then. The rxui integration uses methods that have existed for quite some time.

PR #865, opened 2024-02-22
Feature:  OfType Operator

dwcullop:
## Description

A new operator `OfType` that is similar to the LINQ operator with the same name.  It converts a changeset of some base type (or interface) to a changeset of a different type by automatically excluding values that can't be cast as the new type.  It effectively combines `ImmutableFilter` and `ImmutableTransform` into a single operator for maximum efficiency for the case when both operations are trivial.

### Bonus Changes

Creates Test-compatible version of the `Spy` (named `TestSpy`) that emits output to `ITestOutputHelper` so that the contents of the Observables can be seen in the test results.


src/DynamicData.Tests/Utilities/ObservableSpy.cs
```diff
@@ -125,6 +126,28 @@ private static class NativeMethods
     private static Func<IChangeSet<T>, string> CreateListChangeSetFormatter<T>(Func<T, string> formatter) where T : notnull =>
         cs => "[List Change Set]" + ChangeSetEntrySpacing + string.Join(ChangeSetEntrySpacing, cs.Select((change, n) => $"#{n} {FormatChange(formatter, change)}"));
 
+    public static IObservable<T> TestSpy<T>(this IObservable<T> source,
+                                                                ITestOutputHelper testOutputHelper, string? opName = null,
```
JakenVeina, line 130: This indentation makes my eyes twitch. :P

src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -3316,6 +3316,26 @@ public static IObservable<T> FinallySafe<T>(this IObservable<T> source, Action f
         return source.Where(changes => changes.Count != 0);
     }
 
+    /// <summary>
+    /// Converts the changesets so that it only contains items of the given type.
+    /// </summary>
+    /// <typeparam name="TObject">The type of the object.</typeparam>
```
JakenVeina, line 3322: Suggest "The type of items in the collection".

src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -3316,6 +3316,26 @@ public static IObservable<T> FinallySafe<T>(this IObservable<T> source, Action f
         return source.Where(changes => changes.Count != 0);
     }
 
+    /// <summary>
+    /// Converts the changesets so that it only contains items of the given type.
```
JakenVeina, line 3320: Suggest "Filters an observable collection to only include items of a specific type.", or something similar, as the description here.

src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -3316,6 +3316,26 @@ public static IObservable<T> FinallySafe<T>(this IObservable<T> source, Action f
...
+    /// </summary>
+    /// <typeparam name="TObject">The type of the object.</typeparam>
+    /// <typeparam name="TKey">The type of the key.</typeparam>
+    /// <typeparam name="TDestination">The type of the objects for the result changeset.</typeparam>
```
JakenVeina, line 3324: Suggest "The type of items to be allowed through the filter."

JakenVeina (approved):

RolandPheasant (approved):

src/DynamicData.Tests/Utilities/ObservableSpy.cs
dwcullop: I'll flatten them out like all the others.

PR #874, opened 2024-03-06
[DO NOT MERGE]: IKeyValueCollection as IReadOnlyList

kronic:
Implement IKeyValueCollection as IReadOnlyList


src/DynamicData/Cache/IKeyValueCollection.cs
```diff
@@ -26,7 +26,7 @@ public interface IKeyValueCollection<TObject, TKey> : IEnumerable<KeyValuePair<T
     /// <value>
     /// The count.
     /// </value>
-    int Count { get; }
+    new int Count { get; }
```
JakenVeina, line 29: Is this an attempt to avoid a breaking change? Should we not just remove these two properties entirely? Or get rid of `IKeyValueCollection<TObject, TKey>` entirely, and replace it with `IReadOnlyList<KeyValuePair<TKey, TObject>>` everywhere it's used?

src/DynamicData/Cache/IKey.cs
```diff
@@ -2,6 +2,7 @@
 // Roland Pheasant licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for full license information.
 
+// ReSharper disable once CheckNamespace
```
JakenVeina, line 5: Would definitely want to see these annotations dropped. Dynamic data uses StyleCop and Roslynator for styling analysis, not ReSharper.

On a separate note, I'd prooobably be open to adding support for ReSharper workflows, if there's some way to do it without our team having to have ReSharper. Especially if ReSharper supports a rules config file like `.editorconfig`, so we don't have to annotate code everywhere there's rules that we regularly break.

src/DynamicData/Cache/IKeyValueCollection.cs
dwcullop: I don't think we want to mask the base members.  I think we can just define it like this and be good:

```csharp
public interface IKeyValueCollection<TObject, TKey> : IReadOnlyList<KeyValuePair<TKey, TObject>>
{
}
```

But it is a *little* confusing having it be called `Collection` when it really is a `List`.  `Collection` implies no indexer, which it absolutely has, but it's too late to fix that, I think.

src/DynamicData/Cache/IKeyValue.cs
```diff
@@ -2,6 +2,7 @@
 // Roland Pheasant licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for full license information.
 
+// ReSharper disable once CheckNamespace
```
dwcullop, line 5: Not everyone uses ReSharper.  Can we remove this?

src/DynamicData/Cache/IKeyValueCollection.cs
dwcullop: > Is this an attempt to avoid a breaking change? Should we not just remove these two properties entirely? Or get rid of `IKeyValueCollection<TObject, TKey>` entirely, and replace it with `IReadOnlyList<KeyValuePair<TKey, TObject>>` everywhere it's used?

I don't think we can get rid of `IKeyValueCollection` because it would be a breaking change, but should be derived from a List interface instead of IEnumerable because it adds Count and an indexer.

dwcullop:
Please add some description to your PR (describe what you are changing and why).  It doesn't have to be long, but it is for posterity.

JakenVeina:
Looks like the build failures are complaining about calling LINQ methods on objects that now support direct calls (E.G. `.Count()` versus `.Count`), and only in `Tests` code. Another case of the analyzers being overly-aggressive, probably, but should be an easy fix.

dwcullop:
> Looks like the build failures are complaining about calling LINQ methods on objects that now support direct calls (E.G. `.Count()` versus `.Count`), and only in `Tests` code. Another case of the analyzers being overly-aggressive, probably, but should be an easy fix. Do those errors not show up on your end?

I'm guessing he might not if he's using ReSharper and not VS.  The build failures on the server can show what changes need to be made.

glennawatson:
Rider will have the same build errors as VS. The analysers work on both. Use rider all the time.

kronic:
@dwcullop @glennawatson @JakenVeina ready for review

src/DynamicData/Cache/IKeyValueCollection.cs
JakenVeina: Yeah, I missed the fact that there were members on the interface, other than just these two, cause github collapses most of the file.

I did a little more research into it as well, and yes, this is a binary-breaking change, regardless of whether we shadow the members or not, so there's really no point in keeping them.

JakenVeina (changes requested):
Still need to drop the two members on `IKeyValueCollection` instead of having them shadow the inherited ones.

Other than that, I think this is good to merge, since we already have some breaking changes queued up. @RolandPheasant agree?

RolandPheasant:
@JakenVeina let's hold back  a couple weeks on merging this, so we can apply other breaking changes as part of a big version release.

kronic:
@RolandPheasant ping

PR #875, opened 2024-03-06
TransformOnRefresh option for TransformWithInlineUpdate.

RolandPheasant:
Transform allows the option to apply the a transform when a refresh even is received. 

This PR is to add this feature to TransformWithInlineUpdate


JakenVeina:
Needs an update to the API file.

src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -5899,13 +5899,14 @@ sortOrder switch
...
     /// <param name="updateAction">Apply changes to the original. Example (previousTransformedItem, newOriginalItem) => previousTransformedItem.Value = newOriginalItem.</param>
+    /// <param name="transformOnRefresh">Should a new transform be applied when a refresh event is received.</param>
     /// <returns>
     /// A transformed update collection.
     /// </returns>
     /// <exception cref="ArgumentNullException">source
     /// or
     /// transformFactory.</exception>
-    public static IObservable<IChangeSet<TDestination, TKey>> TransformWithInlineUpdate<TDestination, TSource, TKey>(this IObservable<IChangeSet<TSource, TKey>> source, Func<TSource, TDestination> transformFactory, Action<TDestination, TSource> updateAction)
+    public static IObservable<IChangeSet<TDestination, TKey>> TransformWithInlineUpdate<TDestination, TSource, TKey>(this IObservable<IChangeSet<TSource, TKey>> source, Func<TSource, TDestination> transformFactory, Action<TDestination, TSource> updateAction, bool transformOnRefresh = false)
```
JakenVeina, line 5909: Should we not add these as new overloads, to avoid the breaking change?

JakenVeina (changes requested):

src/DynamicData/Cache/ObservableCacheEx.cs
RolandPheasant: There's already a breaking change due to https://github.com/reactivemarbles/DynamicData/pull/864
so probably no need to worry.

See my comment to Glenn on the linked PR

dwcullop: It isn't the same breaking change as the other one.  The other one doesn't impact RxUI but this one might?  An overload might be safer.

Either way, you've got to update the API text file for the build to start passing.

dwcullop (approved):
The changes look fine to me, but you should decide on the breaking change vs. additional overload issue.  Either way, the API text file needs updating.

JakenVeina (approved):

RolandPheasant:
> The changes look fine to me, but you should decide on the breaking change vs. additional overload issue. Either way, the API text file needs updating.

@dwcullop I added the additional overloads instead of using optional params

RolandPheasant:
> Needs an update to the API file.

@JakenVeina done

PR #876, opened 2024-03-12
Fix nuget.config link

kronic:
Fix nuget.config link


JakenVeina (approved):

JakenVeina:
Crap, I missed that there was no commit message. Please make sure to include that in the future.

Issue #668, opened 2022-11-30
[Bug]: DisposeMany() suppresses exceptions


seanreid1201:
### Describe the bug 🐞

Calling DisposeMany() on an IObservable<IChangeSet<T>> prevents exceptions propagating out to the caller.
See the test case provided below.
If the call to DisposeMany is omitted, the test passes.

### Step to reproduce

    [Fact]
    public void ExceptionsArePropagated()
    {
        var source = new SourceList<int>();
        source.Add(1);

        var transformed = source
            .Connect()
            .Transform<int, int>(_ => throw new InvalidCastException())
            .DisposeMany();

        FluentActions.Invoking(() => transformed.Subscribe()).Should().Throw<InvalidCastException>();
    }

### Reproduction repository

My work VPN prevents access to github via git

### Expected behavior

When an exception is thrown from within the observable, and not handled by the caller, it should propagate out.

### IDE

Visual Studio 2022

### DynamicData Version

7.12.1

RolandPheasant:
Thanks for reporting this. I'll take a look soon

seanreid1201:
I've taken some time to look into this.
In `DynamicData.List.Internal.OnBeingRemoved<T>`, line 35, it does this:
```
var subscriber = _source.Synchronize(locker).Do(changes => RegisterForRemoval(items, changes), observer.OnError).SubscribeSafe(observer);
```
If this is changed to:
```
var subscriber = _source.Synchronize(locker).Do(changes => RegisterForRemoval(items, changes), ).Subscribe(observer);
```
then the exception bubbles out.

However, from a quick search of the code, it appears that SubscribeSafe is used in many places so other methods will also exhibit this behaviour (for example, I found that RefCount also suppresses errors).

I imagine it is likely then that you will say that this behaviour will not change.

If that's the case, is there something I can do on the outside to make the exceptions on the OnError channel throw?
I've tried various things, but they all seem to just get suppressed.
For example, in the test case I provided, if you subscribe like:
```
transformed.Subscribe(x => { }, ex => throw new Exception(), () => { });
```
the exception is still swallowed.

RolandPheasant:
Exceptions should propagate. Let's not rush into changing everything, but it probably should be done.

JakenVeina:
Exceptions aren't not-propagating, they're just being emitted via `OnError()` within the stream, instead of throwing out of `.Subscribe()`. I think this is definitely the correct approach.

[...]

PR #884, opened 2024-03-23
Observable comparers for SortAndBind

RolandPheasant:
Add  IObservable<IComparer<TObject>> option to SortAndBind to enable changing the sort order.

```cs
var comparer = SortExpressionComparer<Person>.Ascending(p => p.Name).ThenByAscending(p => p.Age);
var comparerObservable = new BehaviorSubject<IComparer<Person>>(comparer );

var sortAndBind = _cache.Connect()
            .SortAndBind(out var list, comparerObservable)
            .Subscribe()
```

This represents the completion of SortAndBind.

@dwcullop / @JakenVeina 

I know there were some previous conversations about  `SubscribeSafe`  but I skipped over that.  My question is why has the pattern  of `.SubscribeSafe(Observer.Create...` been introduced and when should it be applied?  I ask because I used `SubscribeSafe(observer)` in this PR


JakenVeina:
> My question is why has the pattern of .SubscribeSafe(Observer.Create... been introduced and when should it be applied?

Bottom line is it's recommended when doing subscriptions within operators.

> Subscribes to the specified source, re-routing synchronous exceptions during invocation of the System.IObservable`1.Subscribe(System.IObserver{`0}) method to the observer's System.IObserver`1.OnError(System.Exception) channel. This method is typically used when writing query operators.

In a practical sense, I tried to write up a small test scenario that shows this in action, and after an hour, I gave up. There's a LOT of baked-in safeguards within RX that normally make `.SubscribeSafe()` redundant. I do remember this was part of the solution for #668. I also suspect that using `.SubscribeSafe()` over `.Subscribe()` has no observable effect on performance, because the meaningful difference between the two is that `.SubscribeSafe()` does 1 or 2 extra `if` checks, on top of the many that `.Subscribe()` already does. Good news is we can easily benchmark this, if we want to be sure, now that we have a few operators that use `.SubscribeSafe()` and also have benchmarks written for them.

src/DynamicData/Binding/SortAndBind.cs
```diff
@@ -17,173 +16,233 @@ namespace DynamicData.Binding;
...
-internal sealed class SortAndBind<TObject, TKey>(
-    IObservable<IChangeSet<TObject, TKey>> source,
-    IComparer<TObject> comparer,
-    SortAndBindOptions options,
-    IList<TObject> target)
+internal sealed class SortAndBind<TObject, TKey>
     where TObject : notnull
     where TKey : notnull
 {
+    // NB: Either comparer or comparerChanged will be used, but not both.
```
JakenVeina, line 23: NB?

src/DynamicData/Binding/SortAndBind.cs
```diff
@@ -17,173 +16,233 @@ namespace DynamicData.Binding;
...
 
-    private int GetInsertPositionBinary(TObject item)
-    {
-        var index = target.BinarySearch(item, comparer);
-        var insertIndex = ~index;
+        private int GetInsertPosition(TObject item) => options.UseBinarySearch ? GetInsertPositionBinary(item) : GetInsertPositionLinear(item);
 
-        // sort is not returning uniqueness
-        if (insertIndex < 0)
+        private int GetInsertPositionBinary(TObject item)
```
JakenVeina, line 221: Wow, this diff is obnoxiously terrible in GitHub, BTW. The syntax-aware diff in VS is so much better. Never seen a case this bad before.

JakenVeina (approved):

src/DynamicData/Binding/SortAndBind.cs
RolandPheasant: https://www.britannica.com/dictionary/NB

JakenVeina: TIL

PR #888, opened 2024-03-28
New Virtualization operator

RolandPheasant:
Introduce `SortAndVirtualize` which replaces the old `Vitualise` mechanism.  This is part of the drive to eliminate the Sort operator altogether.

**Old**

```cs
var virtualRequests =new BehaviorSubject<IVirtualRequest>(new VirtualRequest(0, 25))

myCache 
  .Sort(myComparer)
  .Virtualise(virtualRequests)
```
**New**

```cs
var virtualRequests =new BehaviorSubject<IVirtualRequest>(new VirtualRequest(0, 25))

myCache 
  .SortAndVirtualize(myComparer)
```

Additionally I also included a variant of `SortAndBind` for binding to the virtual output.  This works on `IList<T>` and `ReadOnlyObservableCollection<T>`

```cs
var virtualRequests =new BehaviorSubject<IVirtualRequest>(new VirtualRequest(0, 25))
var myList = new List<MyType>();

myCache 
  .SortAndVirtualize(myComparer)
  .SortAndBind(myList)  \\ comparer is extracted from the output of virtualize
```

Once this is in, it's almost a direct lift of this code for Page.  Then we can mark Sort as obsolete, and at some point delete the old sort + remove the index and previous index from the change object.


src/DynamicData/Cache/Internal/UniquenessEnforcer.cs
```diff
@@ -11,20 +11,45 @@ internal sealed class UniquenessEnforcer<TObject, TKey>(IObservable<IChangeSet<T
...
-* If we handle refreshes, we cannot use .Last() as the last in the groupd may be a refresh,
-* and a previous in the group may add or update. Suddenly this scenario becomes very complicated
-* so for this phase we'll ignore these.
-*
+/*
+    For refresh,  we need to check whether there was a previous add or update in the batch. If not use refresh,
+    otherwise use the previous update.
 */
 
         source
```
RolandPheasant, line 19: For virtualize I had to ensure unique changes, which previously did not work for Refresh

src/DynamicData/Cache/ObservableCacheEx.VirtualiseAndPage.cs
```diff
@@ -0,0 +1,191 @@
+﻿// Copyright (c) 2011-2023 Roland Pheasant. All rights reserved.
+// Roland Pheasant licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using System.Reactive.Linq;
+using DynamicData.Cache.Internal;
+
+namespace DynamicData;
+
+/// <summary>
```
RolandPheasant, line 10: I have included Virtualize, Page and Top operators here. Maybe overkill to have one file per operator.  Thoughts?

src/DynamicData/Cache/ChangeSet.cs
```diff
@@ -5,6 +5,44 @@
 // ReSharper disable once CheckNamespace
 namespace DynamicData;
 
+/// <summary>
+/// A collection of changes with some arbitrary additional context
+/// </summary>
+/// <typeparam name="TObject">The type of the object.</typeparam>
+/// <typeparam name="TKey">The type of the key.</typeparam>
+/// <typeparam name="TContext">The additional context.</typeparam>
+public sealed class ChangeSet<TObject, TKey, TContext> : ChangeSet<TObject, TKey>, IChangeSet<TObject, TKey, TContext>
```
dwcullop, line 14: I'm going to have to keep in mind this is a thing now.  I can't quite place it, but I feel like I needed it before...

dwcullop (approved):

src/DynamicData/Cache/Internal/UniquenessEnforcer.cs
```diff
@@ -11,20 +11,45 @@ internal sealed class UniquenessEnforcer<TObject, TKey>(IObservable<IChangeSet<T
...
+                        // find the previous non-refresh and return if found
+                        for (var i = all.Length - 1; i >= 0; i--)
+                        {
+                            var candidate = all[i];
+                            if (candidate.Reason != ChangeReason.Refresh)
+                                return candidate;
+                        }
+
+                        // the entire batch are all  refresh events
+                        return all[0];
```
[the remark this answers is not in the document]
RolandPheasant: Actually it's less that I am using linq to avoid performance pitfalls, it more that I need to track backwards from the end of the collection. I also suspect you're right about avoiding using last but my rationale is that for 99% of all cases there will only ever be a single item in the grouping ... so I reasoned that avoiding a to list / array was possibly the best, and that .last() would only iterate once is almost all cases.

It's a tough balance to make, and it would be interesting to get some perf stats on it.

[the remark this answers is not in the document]
RolandPheasant: ...ignore me, I already do a to array. I should revisit that code!

src/DynamicData/Cache/ChangeSet.cs
RolandPheasant: Yes, I needed some extra params to be passed down the chain, and historically I have inherited and created a new changeset. However I think this is an improvement.

src/DynamicData/Cache/Internal/UniquenessEnforcer.cs
```diff
@@ -11,20 +11,45 @@ internal sealed class UniquenessEnforcer<TObject, TKey>(IObservable<IChangeSet<T
...
+                    cache ??= new ChangeAwareCache<TObject, TKey>(changes.Count);
+
+                    var grouped = changes.GroupBy(c => c.Key).Select(c =>
+                    {
+                        var last = c.Last();
+
+                        // If the last is a refresh we need to check whether, there was a previous add or update.
+                        // If so , use the previous.
+                        if (last.Reason != ChangeReason.Refresh)
+                            return last;
```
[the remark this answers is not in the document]
RolandPheasant: We need a benchmark here. Maybe a future improvement

src/DynamicData/Cache/ObservableCacheEx.VirtualiseAndPage.cs
```diff
@@ -0,0 +1,191 @@
...
+    /// Sort and virtualize the underlying data from the specified source.
+    /// </summary>
+    /// <typeparam name="TObject">The type of the object.</typeparam>
+    /// <typeparam name="TKey">The type of the key.</typeparam>
+    /// <param name="source">The source.</param>
+    /// <param name="comparer">The comparer to order the resulting dataset.</param>
+    /// <param name="virtualRequests">The virtualizing requests.</param>
+    /// <returns>An observable which will emit virtual change sets.</returns>
+    /// <exception cref="ArgumentNullException">source.</exception>
+    public static IObservable<IChangeSet<TObject, TKey, VirtualContext<TObject>>> SortAndVirtualize<TObject, TKey>(this IObservable<IChangeSet<TObject, TKey>> source,
```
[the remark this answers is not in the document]
RolandPheasant: I am torn on this one. Adding the SortAnd... clearly marks it apart from the old way. However, you're absolutely correct that it may not be required at all. That also holds true for the SortAndBind operator. Do we just rename these back to the original versions without the SortAnd prefix?  I think we probably need a face to face on this question

src/DynamicData.Tests/DynamicData.Tests.csproj
```diff
@@ -4,6 +4,9 @@
     <NoWarn>$(NoWarn);CS0618;CA1801;CA1812;CA1816;CA1063;CS8767;CS8602;CS8618;IDE1006</NoWarn>
     <XunitVersion>2.7.0</XunitVersion>
   </PropertyGroup>
+  <ItemGroup>
+    <Compile Remove="Cache\SortAndBindVirtualize.Fixture.cs" />
+  </ItemGroup>
```
JakenVeina, line 9: Leftover development noise?

src/DynamicData/Cache/ChangeSet.cs
JakenVeina: > I'm going to have to keep in mind this is a thing now. I can't quite place it, but I feel like I needed it before...

I'll wager it was for one of your various stream merging scenarios.

src/DynamicData/Cache/ObservableCacheEx.VirtualiseAndPage.cs
JakenVeina: Me, I would probably stick with one operator per-file, just for clarity.

src/DynamicData/Cache/ObservableCacheEx.VirtualiseAndPage.cs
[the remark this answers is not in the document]
JakenVeina: Using just `.Virtualize()` and `.Bind()` only works if there's no ambiguity at the call site.

```cs
using var subscription = someStream
    .SortAndBind(out var target)
    .Subscribe();
```

Right now, this is a valid invocation of the operator, if the items in `someStream` are `IComparable<TObject>`. If we were to swap `.SortAndBind()` to just `.Bind()` we lose a fair bit of clarity, I think, that there's a sort happening. And what if the user wants to opt-out of that implicit sort? Is that even possible?

I'm not sure what the equivalent scenario would be for virtualization, but that's the perspective we want to look at, to consider `.SortAndVirtualize()` versus just `.Virtualize()`.

src/DynamicData/Cache/ObservableCacheEx.cs
```diff
@@ -3592,24 +3592,6 @@ public static IObservable<T> FinallySafe<T>(this IObservable<T> source, Action f
...
-    public static IObservable<IPagedChangeSet<TObject, TKey>> Page<TObject, TKey>(this IObservable<ISortedChangeSet<TObject, TKey>> source, IObservable<IPageRequest> pageRequests)
-        where TObject : notnull
-        where TKey : notnull
-    {
-        source.ThrowArgumentNullExceptionIfNull(nameof(source));
-        pageRequests.ThrowArgumentNullExceptionIfNull(nameof(pageRequests));
-
-        return new Page<TObject, TKey>(source, pageRequests).Run();
-    }
-
```
JakenVeina, line 3612: If these operators truly are obsolete, I think we should just mark them as such, and remove them in the future, rather than removing them outright, right now.

src/DynamicData/Cache/Internal/UniquenessEnforcer.cs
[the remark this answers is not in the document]
RolandPheasant: I've changed the code as per your suggestion.

src/DynamicData.Tests/DynamicData.Tests.csproj
RolandPheasant: Yes, removed.

PR #917, opened 2024-07-06
New SortAndPage operator

RolandPheasant:
This is the following up to the SortAndVirtualize - see https://github.com/reactivemarbles/DynamicData/pull/888

If is an almost a direct lift of the code but using page based calculations.   This is part of the drive to eliminate the Sort operator altogether, and subsequently the need to propagate the index on `Change<TObject,TKey>`.  

**Old**

```cs
var pageRequests=new BehaviorSubject<IPageRequest>(new PageRequest(1, 25))

myCache 
  .Sort(myComparer)
  .Page(pageRequests)
```
**New**

```cs
var pageRequests=new BehaviorSubject<IPageRequest>(new PageRequest(1, 25))

myCache 
  .SortAndPage(myComparer, pageRequests)
```

There is a also a  new `Bind()` overload to bind the results of the page response.

Additionally I have renamed SortAndBind which acts on a a virtual response to `Bind()` .

I just need and replace `SortBy` with `SortByAndBind` operator and I am happy to make the big version release.

### DIFF
--- a/src/DynamicData/Cache/Internal/SortAndPage.cs
+++ b/src/DynamicData/Cache/Internal/SortAndPage.cs
@@ -1,0 +1,195 @@
+﻿// Copyright (c) 2011-2023 Roland Pheasant. All rights reserved.
+// Roland Pheasant licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using System.Reactive.Linq;
+using DynamicData.Binding;
+
+namespace DynamicData.Cache.Internal;
+
+internal sealed class SortAndPage<TObject, TKey>
+    where TObject : notnull
+    where TKey : notnull
+{
+    private static readonly KeyComparer<TObject, TKey> _keyComparer = new();
+
+    private readonly IObservable<IChangeSet<TObject, TKey>> _source;
+    private readonly IObservable<IComparer<TObject>> _comparerChanged;
+    private readonly IObservable<IPageRequest> _pageRequests;
+    private readonly SortAndPageOptions _options;
+
+    public SortAndPage(IObservable<IChangeSet<TObject, TKey>> source,
+        IComparer<TObject> comparer,
+        IObservable<IPageRequest> virtualRequests,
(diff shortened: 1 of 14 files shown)


src/DynamicData/Cache/Internal/SortAndPage.cs
```diff
@@ -0,0 +1,195 @@
...
+                        {
+                            applicator.ChangeComparer(keyValueComparer);
+                        }
+                        return ApplyPagedChanges();
+                    });
+
+                var paramsChanged = _pageRequests.Synchronize(locker)
+                    .DistinctUntilChanged()
+                    // exclude dodgy params
+                    .Where(parameters => parameters is { Page: > 0, Size: > 0 })
```
JakenVeina, line 85: Doesn't this filter out some valid requests? E.G. `{ Page: 0, Size: 10 }`? Or is `Page` here 1-indexed?

JakenVeina (commented):
On a high-level note, am I correct in reading that `Virtualize` and `Page` are just the same thing? Is there any actual difference, other than the indexing parameters?

src/DynamicData/Cache/Internal/SortAndPage.cs
RolandPheasant: It just ignores bad input. The other alternative would be just to throw an exception

RolandPheasant:
> On a high-level note, am I correct in reading that `Virtualize` and `Page` are just the same thing? Is there any actual difference, other than the indexing parameters?

Indeed so. It's almost the same but I've replaced it because it already exists and we can bet our bottom dollar that someone is suing it in the wild.

src/DynamicData/Cache/Internal/SortAndPage.cs
RolandPheasant: Dang, I just re-read your comments. Page zero is not valid. Paging always starts at 1.

JakenVeina (approved):

Issue #925, opened 2024-08-16
[Bug]: TransformImmutable: For ChangeReason.Add, a previous value cannot be specified


JakenVeina:
### Describe the bug 🐞

The following snippet produces an `ArgumentException`.

### Step to reproduce

```cs
using var sc = new SourceCache<string, string>(s => s);
using var sub = sc.Connect().TransformImmutable(s => s.Length).Subscribe();
sc.AddOrUpdate("Hey");
```

### Additional information ℹ️

This is related to a bogus use of `default` within a ternary expression, [here](https://github.com/reactivemarbles/DynamicData/blob/41a608d0a358add91e4403feb67ec02d688c8d81/src/DynamicData/Cache/Internal/TransformImmutable.cs#L43). `default` here resolves to `TDestination` instead of `Optional<TDestination>`, when `TDestination` is a value type.

PR #926, opened 2024-08-16
Fix for #925

JakenVeina:
Fixed a bogus use of the `default` keyword, within a ternary expression where implicit casting of a generic is in play. The expression compiles differently depending on whether the generic type in question is a value type, versus a reference type, and generates an exception in the case of a value type.


dwcullop (approved):
LGTM!  Thanks for fixing this so quickly.

Issue #932, opened 2024-08-30
[Bug]: SortAndBind with IObservable<IComparer> not working as expected


DevEngineReq:
### Describe the bug 🐞

If we use SortAndBind with an IObservable<IComparer> we have to use ObserveOn(RxApp.MainThreadScheduler) for the compare observable, otherwise the application will freeze. For the deprecated .Sort() method it was not required. 

This behavour can cause wired problems you switch from .Sort to SortAndBind and just copy the compare observable.

![image](https://github.com/user-attachments/assets/d96eaf50-dccb-4cab-b695-4ce0f2d094f7)

### Reproduction repository

https://github.com/reactivemarbles/DynamicData

### Expected behavior

See Descriptions

JakenVeina:
I can agree this looks unintuitive, but it's very much to be expected, and is in fact a behavior you can get with the separate `.Sort()` and `Bind()` operator.

Using the `.SortAndBind()` operator as in your example is the equivalent of doing the following:

```cs
this.cacheObservable = this.userCache.Connect()
    .ObserveOn(RxApp.MainThreadScheduler)
    .Sort(sorterObservable)
    .Bind(this.users)
    .Subscribe();
```

The issue is that there are 2 separate streams feeding input into the `.Sort()` operation, the `.Connect()` stream, and `sorterObservable`. When `.Sort()` and `.Bind()` are separate, it's trivial to insert the proper `.ObserveOn()` after the `.Sort()`, when the two separate streams have been merged into just one. But without that, you need to apply `.ObserveOn()` to both input streams, separately.

It sounds like we should add an optional `scheduler` parameter to `.SortAndBind()` to support this scenario, and to help clarify for folks migrating to it that a `scheduler` may be necessary.

DevEngineReq:
Like this

![image](https://github.com/user-attachments/assets/390b51e7-80f9-4123-aae8-a9ebc3b34e54)

It would be nice if the stream would go ahead with the entered scheduler after SortAndBind(....)

JakenVeina:
Yes, you'd end up with the downstream being driven by that scheduler.

DevEngineReq:
Any news on this issue?

Will there be an optional scheduler parameter to .SortAndBind() like described above?

JakenVeina:
In fact, this has just been added and released. You can now supply an `IScheduler` to `.SortAndBind()` via the `options` parameter.

DevEngineReq:
Works as expected. Thank you

PR #939, opened 2024-09-04
SortAndBind fixes and improvements

RolandPheasant:
- Enable opt-in to ResetOnFirstTimeLoad to ensure backwards compatibility with the old `Sort(...).Bind(...)` behaviour.

To opt-in system-wide after upgrading from v8 to v9 (and migrating to SortAndBind) like this:

```cs
DynamicDataOptions.SortAndBind = DynamicDataOptions.SortAndBind with { ResetOnFirstTimeLoad = true };
```

- Specialised handing for BindingList<T> when used with `SortAndBind`, to ensure reset is respected.

- Specify SortAndBind scheduler at a system with level.

```cs
DynamicDataOptions.SortAndBind = DynamicDataOptions.SortAndBindwith { Scheduler= myMainThreadScheduler };
```

With this ObserveOn(myMainThreadScheduler) will be a thing of the past.

### DIFF
--- a/src/DynamicData/Binding/SortAndBind.cs
+++ b/src/DynamicData/Binding/SortAndBind.cs
@@ -2,6 +2,7 @@
 // Roland Pheasant licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for full license information.
 
+using System.ComponentModel;
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using DynamicData.Cache;
@@ -33,12 +34,17 @@ internal sealed class SortAndBind<TObject, TKey>
         // static one time comparer
         var applicator = new SortApplicator(_cache, target, comparer, options);
 
-        _sorted = source.Do(changes =>
+        if (options.Scheduler is not null)
+            source = source.ObserveOn(options.Scheduler);
+
+        _sorted = source.Select((changes, index) =>
         {
             // clone to local cache so that we can sort the entire set when threshold is over a certain size.
             _cache.Clone(changes);
 
-            applicator.ProcessChanges(changes);
+            applicator.ProcessChanges(changes, index == 0);
+
+            return changes;
         });
     }
 
@@ -48,6 +54,9 @@ internal sealed class SortAndBind<TObject, TKey>
         IList<TObject> target)
         => _sorted = Observable.Create<IChangeSet<TObject, TKey>>(observer =>
         {
+            if (options.Scheduler is not null)
+                source = source.ObserveOn(options.Scheduler);
+
             var locker = new object();
             SortApplicator? sortApplicator = null;
 
@@ -61,14 +70,17 @@ internal sealed class SortAndBind<TObject, TKey>
 
             // Listen to changes and apply the sorting
             var subscriber = source.Synchronize(locker)
-                .Do(changes =>
+                .Select((changes, index) =>
                 {
                     _cache.Clone(changes);
 
                     // the sort applicator will be null until the comparer change observable fires.
                     if (sortApplicator is not null)
-                        sortApplicator.ProcessChanges(changes);
-                }).SubscribeSafe(observer);
+                        sortApplicator.ProcessChanges(changes, index == 0);
+
+                    return changes;
+                })
+                .SubscribeSafe(observer);
 
             return new CompositeDisposable(latestComparer, subscriber);
         });
@@ -92,10 +104,12 @@ internal sealed class SortAndBind<TObject, TKey>
         }
 
         // apply sorting as a side effect of the observable stream.
-        public void ProcessChanges(IChangeSet<TObject, TKey> changeSet)
+        public void ProcessChanges(IChangeSet<TObject, TKey> changeSet, bool isFirstTimeLoad)
         {
+            var forceReset = isFirstTimeLoad && options.ResetOnFirstTimeLoad;
+
             // apply sorted changes to the target collection
-            if (options.ResetThreshold > 0 && options.ResetThreshold < changeSet.Count)
+            if (forceReset || (options.ResetThreshold > 0 && options.ResetThreshold < changeSet.Count))
             {
                 Reset(cache.Items.OrderBy(t => t, comparer), true);
             }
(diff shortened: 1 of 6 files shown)


src/DynamicData/Binding/SortAndBind.cs
```diff
@@ -33,12 +34,17 @@ public SortAndBind(IObservable<IChangeSet<TObject, TKey>> source,
         // static one time comparer
         var applicator = new SortApplicator(_cache, target, comparer, options);
 
-        _sorted = source.Do(changes =>
+        if (options.Scheduler is not null)
+            source = source.ObserveOn(options.Scheduler);
```
JakenVeina, line 38: This is insufficient. Refer to #932, the logical query this needs to serve as a replacement for is...

```cs
.Sort(...)
.ObserveOn(...)
.Bind(...)
```

I.E. the scheduling needs to be applied to both input streams, `source` and `comparer`. Otherwise, notifications coming in on `comparer` can trigger the `ObservableCollection` be manipulated off of the main thread.

Another `.ObserveOn()` call applied to `comparer` should be sufficient here, since there is no change stream "between" the sorting and binding steps anymore.

JakenVeina (changes requested):

src/DynamicData/Binding/SortAndBind.cs
RolandPheasant: Fair call, I'll fix that.  

Also I was wondering whether to add a scheduler overloads to SortAndBind, or instead to instruct the use of

```cs
DynamicDataOptions.SortAndBind = DynamicDataOptions.SortAndBind with
{
    Scheduler = myMainThreadScheduler
};
```
which is neat but not obvious.

Or should I do both?  Thoughts?

RolandPheasant: I've applied the scheduler to the comparer changed observable, and moved the UI thread scheduler option to `DynamicDataOptions.MainThreadScheduler`.

So the one remaining question is whether it's worth adding 12 new overloads to pass the comparer to SortAndBind or to rely on the global main thread scheduler?

JakenVeina: I'd say both. There's value in being able to specify on a per-call basis, and in being able to set it globally and not have to worry about it.

Not sure about whether a global-level scheduler should be universal, or per-operator. Again, I think there's value in both, but in this case, I think it's a little cumbersome to HAVE both. If it's not per-operator, I would suggest naming it `BindScheduler` because that clarifies how it's actually being USED if you set it. It would only be pulled for binding operations.

RolandPheasant: Renamed to BindingScheduler + added scheduler overloads for new generation of binding operators.

dwcullop: > Also I was wondering whether to add scheduler overloads to SortAndBind, or instead to instruct the use of

I suggest both: Giving the ability to provide all the options for their special cases AND providing a helper method that facilitates the most common use case(s).

src/DynamicData/DynamicDataOptions.cs
```diff
@@ -20,4 +21,10 @@ public static class DynamicDataOptions
     /// Gets or sets the system-wide default values for all SortAndBind operations.
     /// </summary>
     public static SortAndBindOptions SortAndBind { get; set; } = new();
+
```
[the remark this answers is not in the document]
RolandPheasant: Dang.  My original PR had the scheduler on `SortAndBindOption` and I'm now going in circles.

In many respects, it makes sense to have the scheduler on `SortAndBindOption` as all the new binding operators use these whereas I would rather not add it to the original binding options at that would mean I'd have to retrospectively change those after.  

>If you do that way, then you don't need a separate IScheduler parameter in addition to the SortAndBindOptions parameter.
> You can have overloads that take EITHER a SortAndBindOptions OR an IScheduler parameter, and then the latter version can just call the former version with something like DynamicDataOptions.SortAndBind with { Scheduler: scheduler }.

Wise words. 

I'll sit on it for a day or two, but keeping the overloads to a minimum seems like a good idea.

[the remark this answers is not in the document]
JakenVeina: > It seems confusing to have ADDITIONAL "Binding" options at the top level when there is already a section for Binding specific options.

It doesn't seem confusing to me. `DynamicDataOptions.BindingScheduler` applies to all "Binding" operators, while `.SortAndBindOptions.Scheduler` applies to just the `.SortAndBind()` operator. I think there's value to be argued for both, but I definitely wouldn't add `DynamicDataOptions.BindingScheduler` without retrofitting it into all the existing operators. That would definitely be counter-intuitive. And I'd say that scope of change should be its own PR.

> If you do that way, then you don't need a separate IScheduler parameter in addition to the SortAndBindOptions parameter.

Agreed. I see no point in having overloads for both `IScheduler` and `SortAndBindOptions`. If there was already an `IScheduler` one, it would make sense to add one for `SortAndBindOptions`, and leave the existing one there to avoid breakage, but that's not the case here.

> In many respects, it makes sense to have the scheduler on SortAndBindOption as all the new binding operators use these

I think this is probably the way to go. That is, having each operator define its own `XXXOptions` struct, instead of just adding optional parameters. It makes adding new options much simpler. And, honestly, I think the point about being forced to retrofit some older operators when adding new options is a BENEFIT. It's a constraint we're placing on ourselves to keep the API a little more sane for consumers. I quite literally think there is only one downside to the `XXXOptions` pattern, and that's the increase in call site size (I.E. every call requires a full allocation of the options struct, regardless of which options are needed). I'm inclined to think this is a cost worth accepting, particularly since operator calls are not hot paths, only the operator innards are.

JakenVeina (approved):

dwcullop (approved):

RolandPheasant:
@dwcullop / @JakenVeina 

... a few days turned into several months. However I am back on this one.  

I've pushed a change which removes the scheduler overload, and include it on the SortAndBind options. Please review, and when I have merged this, I'll then bump the version and release.

giard-alexandre:
Since this PR is playing in SortAndBind already, I wonder if you migth have an easy fix for https://github.com/reactivemarbles/DynamicData/issues/975?
(I don't want to drastically creep the scope, just bringing it into view 😄 )

JakenVeina (approved):
Looks like nothing's really changed since the last review, except the `IScheduler` shuffling, so I'm good.

Issue #931, opened 2024-08-29
[Bug]: `ReferenceEquals` breaks value types


erri120:
### Describe the bug 🐞

DynamicData has many instances of using `ReferenceEquals` which breaks observable streams silently for value types.

The issue I ran into specifically was using `MergeManyChangeSets` for `IChangeSet<TValue, TKey>` where both the value and key are value types. When an item upstream is getting removed, the item was never removed from the cache because the equality check uses `ReferenceEquals`:

https://github.com/reactivemarbles/DynamicData/blob/afbbb4742e0caf99dc1343fe7bbcb23f4fc7ceba/src/DynamicData/Cache/Internal/ChangeSetMergeTracker.cs#L270-L271

The temporary workaround is to add a custom equality comparer, but a proper solution should use `EqualityComparer<T>.Default` or the custom equality comparer instead of `ReferenceEquals`.

### Step to reproduce

```csharp
var source1 = new SourceCache<int, int>(x => x);
var source2 = new SourceCache<bool, int>(_ => throw new NotSupportedException());

source1.Connect().MergeManyChangeSets(_ => source2.Connect()).Do(changeSet =>
{
    foreach (var change in changeSet)
    {
        Console.WriteLine($"{change.Reason}: {change.Current}");
    }
}).Subscribe();

source1.Edit(updater => updater.AddOrUpdate(1, key: 1));

// produces console log
source2.Edit(updater => updater.AddOrUpdate(false, key: 1));

// doesn't produce console log
source2.Edit(updater => updater.Remove(key: 1));
```

### Expected behavior

Value types shouldn't break DynamicData silently or loudly with exceptions, as seen in #925.

### DynamicData Version

9.0.4

JakenVeina:
I would agree. I can't imagine a scenario where `EqualityComparer<T>.Default` isn't the proper choice, but I also have never had a use-case for this operator, so I'll defer to @dwcullop for a second-opinion. He's the one who actually wrote this, and has use-cases for it all the time.

dwcullop:
Clearly a bug.  Will fix and also check for other uses of ReferenceEquals.

dwcullop:
> I would agree. I can't imagine a scenario where `EqualityComparer<T>.Default` isn't the proper choice

[...]

PR #940, opened 2024-09-06
Fix for ChangeSetMergeTracker so that it correctly works with Value Types

dwcullop:
## Description
- Addresses #931 by fixing `ChangeSetMergeTracker` (the basis for multiple operators including `MergeManyChangeSets`, `MergeChangeSets`, and `TransformManyAsync`)
- Replaces most instances of `ReferenceEquals` with `EqualityComparer<T>.Default.Equals`
- A specific helper method was created to use when `ReferenceEquals` is explicitly needed (which will gracefully handle value types)
- Added unit tests to verify the fix (tests failed without the additional changes)

## Other Fixes
- Updates `ToObservableOptional` with a more efficient implementation that does not use `ReferenceEquals`


src/DynamicData/Cache/Internal/ChangeSetMergeTracker.cs
```diff
@@ -268,9 +269,12 @@ private Optional<TObject> LookupBestValue(ChangeSetCache<TObject, TKey>[] source
     }
 
     private bool CheckEquality(TObject left, TObject right) =>
-        ReferenceEquals(left, right) || (equalityComparer?.Equals(left, right) ?? false);
+        _equalityComparer.Equals(left, right);
+
+    private bool CheckReferenceEquality(TObject left, TObject right) =>
+        ReferenceEquals(left, right) || (typeof(TObject).IsValueType && CheckEquality(left, right));
```
erri120, line 275: `ReferenceEquals` should never be called if the object is a value type, as it will always return false and will always box and allocate the values.

The check `typeof(TObject.IsValueType)` should be first, the compiler and JIT can optimize that branch away:

```csharp
typeof(TObject.IsValueType)) ? CheckEquality(left, right) : ReferenceEquals(left, right);
```

I don't believe `ReferenceEquals` should be used here at all, the definition of equality has to be consistent across all operators, otherwise it'd just lead to unexpected results. `EqualityComparer<T>.Default` will be a reference equals if `T` doesn't implement custom equality operators.

dwcullop: @erri120 I understand what you're saying about equality needing to be consistent, but I think that we really do need `ReferenceEquals` in a couple of places here to handle corner cases.

Consider:
1) Child ChangeSet A emits an Add Event with Key K
   Downstream Should See:  Add Event for K with value A
   
2) Child ChangeSet B emits an Add Event for the same Key.  Using the IComparer, it is determined that A's value is equal, so the downstream sees nothing from the Add event from B.
   Downstream Should See:  Nothing, because the current value is the same.
   
3) ChangeSet B emits a "Refresh" event for Key K.  The IComparer runs again it is determined that the two values are still equal.
   Downstream Should See:  Nothing, because the "Refresh" event came from a value that has not yet been observed.

However, if the Refresh had come from A, then we WOULD want to emit the "Refresh" downstream.
When checking to see if we should emit the "Refresh" downstream, we have to use `ReferenceEquals` because we want to make sure it came from the EXACT same value emitted downstream.  If we just use EqualityComparer, and it says the refreshed value (B) and the current value from (A) are equal, then we will emit a bad Refresh event downstream.

It is BAD because:
1) Downstream would see a Refresh when the value from A has not actually refreshed
2) Downstream would see a Refresh event when the Current value is the value from B, but they've never seen B before.  You shouldn't change the value using a Refresh event (if Current changes, it needs to be an Update event).

dwcullop: Keep in mind that this class only has the current changeset and the values that have already been emitted downstream, so it can't check to see if it came from the source or not.  It has to compare the values themselves.

JakenVeina: > When checking to see if we should emit the "Refresh" downstream, we have to use ReferenceEquals because we want to make sure it came from the EXACT same value emitted downstream.

I wholeheartedly disagree. The entire point of `IEquatable<T>` and `IEqualityComparer<T>` is to allow the consumer to TELL that we DON'T need to make sure it was the "EXACT same value emitted downstream". It's the consumer's way of telling us that certain objects which are different references actually have equivalent state **and are thus fully-interchangeable with each other**. If the consumer tells us two different references are interchangeable, and we treat them as interchangeable, and then something breaks, they were apparently NOT interchangeable, due to some amount of reference-specific state. I.E. it's a mistake by the consumer.

It took me a solid hour of attempting to write a detailed response to work this out, but the convincing argument for me is that there is actually a more fundamental defect in how you're handling `Refresh` changes (from what I read above, anyway). You're failing to consider that the fundamental POINT of a `Refresh` operation is to tell us that the state of a reference has changed, meaning it's equality compared to other objects may ALSO have changed. Based on what you described, you're just assuming that if a `Refresh` comes in you need to decide whether or not to forward that downstream, but in fact you may need to TRANSLATE it to an `Update`. Or, perhaps, translate some `Update`s into `Refresh`es.

Let's walk two slightly different versions of your scenario:
* Stream #1 emits `Add` for A@K
* Stream #2 emits `Add` for B@K
  * `EqualityComparer<T>.Default` says A and B are equal
    * We swallow the `Add` for B@K and emit nothing
      * "Equal" means the two instances should be fully-interchangeable, so downstream operators are free to keep using A, instead of B, at that key. If that breaks something downstream, it's the consumer's fault.
* Stream #2 emits Refresh for B@K
  1 `EqualityComparer<T>.Default` says A and B are equal
    * We forward the `Refresh`
      * "Equal" means the two instances should be fully-interchangeable, so even though downstream operators may have captured A previously, they are free to keep A after the refresh. If that breaks something downstream, it's the consumer's fault.
  2 `EqualityComparer<T>.Default` says A and B are not equal
    * We translate the `Refresh` into an `Update`
 
Bottom line, if the consumer is TELLING us to use a definition of equality other than reference equality, it's not our job to second-guess them. It is our job (if we're going to support mutability of objects, via `Refresh`) to recognize that equality is ALSO mutable, and act accordingly.

Al12rs (commented):
Bump, for rebase and review, feedback above seems to have been addressed

JakenVeina (approved):

dwcullop:
Sorry I was on vacation the past 10 days.  It will now just use the given comparer or EqualityComparer<T>.Default (if not provided).

JakenVeina (approved):

Issue #983, opened 2025-02-13
[Bug]: DynamicData is not AOT Friendly


dwcullop:
## Description

When using DynamicData in an WinAppSDK application that is compiled natively with AOT, the trimmer emitted the following warnings:

```
6>/_/src/DynamicData/Binding/BindingListEventsSuspender.cs(21,17): Trim analysis error IL2091: DynamicData.Binding.BindingListEventsSuspender<T>.<>c__DisplayClass1_0.<.ctor>b__0(): 'T' generic argument does not satisfy 'DynamicallyAccessedMemberTypes.All' in 'System.ComponentModel.BindingList<T>'. The generic parameter 'T' of 'DynamicData.Binding.BindingListEventsSuspender<T>.<>c__DisplayClass1_0' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
6>/_/src/DynamicData/Binding/BindingListEventsSuspender.cs(22,17): Trim analysis error IL2091: DynamicData.Binding.BindingListEventsSuspender<T>.<>c__DisplayClass1_0.<.ctor>b__0(): 'T' generic argument does not satisfy 'DynamicallyAccessedMemberTypes.All' in 'System.ComponentModel.BindingList<T>'. The generic parameter 'T' of 'DynamicData.Binding.BindingListEventsSuspender<T>.<>c__DisplayClass1_0' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
6>ILLink : Trim analysis error IL2091: DynamicData.Binding.BindingListEventsSuspender<T>.<>c__DisplayClass1_0.list: 'T' generic argument does not satisfy 'DynamicallyAccessedMemberTypes.All' in 'System.ComponentModel.BindingList<T>'. The generic parameter 'T' of 'DynamicData.Binding.BindingListEventsSuspender<T>.<>c__DisplayClass1_0' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
[...]

PR #987, opened 2025-02-28
Add AOT Friendly Attributes for BindingList<T> Compatibility

dwcullop:
## Description
1) Created a Polyfill for the `DynamicallyAccessedMembers` attribute / enum which isn't available until Net5.

2) Updated any class using `BindingList` to include the `DynamicallyAccessedMembers` attribute so that the type have the same attributes as `BindingList`.

3) Recursively walked up the code hierarchy to apply the attribute all the way up.

All of the warnings mentioned in #983 were related to `BindingList<T>`.  Any generic code that uses that class directly (or indirectly) needs to have the same AOT attributes so that the trimmer will be sure to not trim needed type information, so this should resolve all of those warnings.

### Bonus Fixes
Converted two files to use file-scoped namespaces.

### DIFF
--- a/src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet8_0.verified.txt
+++ b/src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet8_0.verified.txt
@@ -312,13 +312,13 @@ namespace DynamicData.Binding
         [System.Obsolete("This never worked properly in the first place")]
         public System.IDisposable SuspendNotifications(bool invokePropertyChangeEventWhenDisposed = true) { }
     }
-    public class BindingListAdaptor<T> : DynamicData.IChangeSetAdaptor<T>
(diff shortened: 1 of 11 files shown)


JakenVeina (approved):
Like I mentioned, I don't really have the context for whether this is functionally correct or appropriatte, but I don't see anything in the code that concerns me.

I will say that diffs on these files in GitHub are abysmal, I believe as a result of all the whitespace changes. Did something change recently with GitHub? Do we have any amount of configurability for how diffs are presented? If not, we maybe need to make a point to keep whitespace changes separate, for reviewability.

RolandPheasant:
> Those are the attributes that have gotten rid of AOT warnings for me in other projects. After the next release, I can confirm that they are indeed resolved.

I'll release tomorrow

xackus:
This also annotates `ReadOnlyObservableCollection` overloads of `Bind` and `SortAndBind`. Is this correct?

JakenVeina:
That seems to be correct, yes.

PR #1005, opened 2025-05-18
Use new lock primitive

RolandPheasant:
Utilize the net9 lock primitive for improved performance. 

To achieve this, I had to  add a version of Rx.Synchronize<T> which accepts the Lock object


src/DynamicData/Internal/Rxx.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) 2011-2023 Roland Pheasant. All rights reserved.
+// Roland Pheasant licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+#if NET9_0_OR_GREATER
+
+namespace System.Reactive.Linq;
+
+internal static class Rxx
```
RolandPheasant, line 9: @JakenVeina / @dwcullop 

This is the main part of the PR.  Can you please consider whether this is correctly implemented

src/DynamicData/Kernel/InternalEx.cs
```diff
@@ -14,6 +14,12 @@ namespace DynamicData.Kernel;
 /// </summary>
 public static class InternalEx
 {
+#if NET9_0_OR_GREATER
```
RolandPheasant, line 17: @JakenVeina / @dwcullop 

We may be able to use a poly fill, and use Lock everywhere.

src/DynamicData/Internal/Rxx.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        _source = source;
+        _locker = new Lock();
+    }
+
+    public IObservable<T> Run()
+    {
+        return Observable.Create<T>(observer =>
+        {
+            return _source.Subscribe(t =>
```
JakenVeina, line 39: Strictly speaking, this should be `.SubscribeSafe()`.

src/DynamicData/Internal/Rxx.cs
```diff
@@ -0,0 +1,61 @@
...
+        new Synchronize<T>(source, locker).Run();
+}
+
+/// <summary>
+/// Keep this class internal as it should be supplied by System.Reactive and probably will be one day.
+/// </summary>
+internal sealed class Synchronize<T>
+{
+    private readonly IObservable<T> _source;
+    private readonly Lock _locker;
```
JakenVeina, line 21: Recommend the name `_gate`, that's generally the terminology Microsoft and RX use for objects used to power a lock.

src/DynamicData/Internal/Rxx.cs
JakenVeina: Aside from the fact that you're not using the `internal` classes that RX uses for its operators, this looks identical to their `.Synchronize()` operator.

src/DynamicData/Internal/Rxx.cs
```diff
@@ -0,0 +1,61 @@
...
+internal static class Rxx
+{
+    public static IObservable<T> Synchronize<T>(this IObservable<T> source, Lock locker) =>
+        new Synchronize<T>(source, locker).Run();
+}
+
+/// <summary>
+/// Keep this class internal as it should be supplied by System.Reactive and probably will be one day.
+/// </summary>
+internal sealed class Synchronize<T>
```
JakenVeina, line 18: I don't see much point in pulling the operator guts out into a separate class here. It was my understanding that we were going to move away from this pattern for our own operators, in favor of doing `partial` classes for the extension fixtures, so we can just split each operator into its own file.

With this just being an attempt to extend RX, that'll likely be temporary, it's also not a big deal, either way.

src/DynamicData/Kernel/InternalEx.cs
JakenVeina: A "polyfill" as we've done before would require pulling the entire source code of .NET's `Lock` class into our own, and `#if` checking around that. Same for all the APIs that USE `Lock` that we'd want to consume.

The "polyfill"s that we have today work really well because those classes don't HAVE any actual code in them, nor any APIs that use them. The only thing that uses them is the compiler, for the sake of static analysis. on newer language features.

What we'd have to do to eliminate the `#if`s here would be to have an entire wrapper class of our own, that is implemented separately .NET 8- and .NET 9+. Which probably would end up beneficial. Would you want to take this PR first, and then circle back to that?

JakenVeina (approved):

JakenVeina:
So, do you want dibs on PR-ing a new `.Synchronize()` to RX?

src/DynamicData/Kernel/InternalEx.cs
RolandPheasant: It's probably not worth porting the Lock code. What do you think?

src/DynamicData/Internal/Rxx.cs
RolandPheasant: Changed.

src/DynamicData/Internal/Rxx.cs
RolandPheasant: I'll stick to locker as it's ubiquitous throughout the code base.

src/DynamicData/Internal/Rxx.cs
RolandPheasant: Agreed, and changed.

RolandPheasant:
> So, do you want dibs on PR-ing a new `.Synchronize()` to RX?

Feel free to do do yourself.  You'd also need to modify the targets in Rx as they only have specific builds up to Net6

dwcullop:
FYI... I hit deadlock the other day in a Net9 application.  Usually VS will tell you "Thread A is waiting on this lock held by Thread B" and "Thread B is waiting on THAT lock held by Thread A".

However, I did not get this behavior even though it was clearly that situation by looking at the threads/call stacks.

I suspect the reason is that VS doesn't know how to deal with the new lock type (yet).  Just a guess, but I was really surprised that I had to trace out the deadlock by hand.

PR #1013, opened 2025-05-29
Removed bogus `.Filter()` overload

JakenVeina:
Removed a bogus .Filter() overload that did not allow the consumer to supply filtering logic, resulting in all items always being filtered out.

This appears to have been mistakenly introduced during a refactor, in commit 3657fee1cb3144ad505c903466aa8c37b31f2214.


dwcullop:
Is it a breaking change if no one is using it? If it does not work, how could anyone be using it?

Better safe than sorry, I guess.

JakenVeina:
I'm a "technically correct is the best kind of correct" kinda person, so I say keep it a major bump. Except, I don't think there's any need to bother bumpibg it ONLY for this. I'd just have us hold onto it until the next time we have something to trigger a major bump that's actually meaningful.

dwcullop (approved):
LGTM.  Let's include it in the next major version bump release.

PR #1048, opened 2025-11-06
Operator and testing modernization for cache `.Filter()` operators, as described in #1014.

JakenVeina:
Includes full re-write of the static version of `.Filter()`, to promote consistent organization among all filtering operators, and to fix a missing null parameter check.

Includes refactoring and enhancement to combine the two separate backing classes for `.Filter()` operators into one, by adding `reapplyFilter` support to the newer one.

Includes full rewrite of the cache .Filter() testing suite.


RolandPheasant:
@JakenVeina there's some conflicts to resolve here

PR #1063, opened 2026-02-21
Re-wrote and optimized the static version of the list .Filter() operator,...

JakenVeina:
...persuant to #1014, and #1060. This operator now properly supports Refresh changes, and preserves ordering of items, for downstream consumers.


RolandPheasant (commented):
This makes sense, and looks reasonable.

Out of interest how does it perform for large lists of > 10 k items?

RolandPheasant (approved):

PR #1071, opened 2026-03-20
test(sum): add more sum tests

giard-alexandre:
This should address the first 3 bullet points in https://github.com/reactivemarbles/DynamicData/discussions/1031#discussioncomment-15777487

### DIFF
--- a/src/DynamicData.Tests/AggregationTests/SumFixture.ForCache.cs
+++ b/src/DynamicData.Tests/AggregationTests/SumFixture.ForCache.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Linq;
 
 using DynamicData.Aggregation;
 using DynamicData.Tests.Domain;
@@ -11,9 +10,9 @@ using Xunit;
 
 namespace DynamicData.Tests.AggregationTests;
 
-public class SumFixture
+public partial class SumFixture
 {
-    public class CacheSource
+    public class ForCache
     {
         [Theory]
         [InlineData(1, 10)]
@@ -222,6 +221,47 @@ public class SumFixture
         }
 
         [Fact]
+        public void SourceCompletesImmediately_InitialSumAndCompletionPropagate()
+        {
+            using var source = new TestSourceCache<Person, string>(p => p.Name);
+
+            source.AddOrUpdate(new Person("A", 10));
+            source.AddOrUpdate(new Person("B", 20));
+            source.AddOrUpdate(new Person("C", 30));
+            source.Complete();
+
+            // UUT Construction: source is already completed, with pre-existing items.
+            // Subscription should produce both an initial sum and a completion, synchronously.
+            using var subscription = source.Connect()
+                .Sum(p => p.Age)
+                .ValidateSynchronization()
+                .RecordValues(out var results);
+
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.HasCompleted.Should().BeTrue("the source was already completed at the time of subscription");
+            results.RecordedValues.Should().ContainSingle("an initial sum value should still be emitted, even when the source completes immediately upon subscription")
+                .Which.Should().Be(60, "the sum of ages 10 + 20 + 30 is 60");
+        }
+
+        [Fact]
+        public void SourceCompletesImmediatelyWithoutEmitting_CompletionPropagates()
+        {
+            using var source = new TestSourceCache<Person, string>(p => p.Name);
+
+            source.Complete();
+
+            // UUT Construction: source is already completed, with no items.
+            using var subscription = source.Connect()
+                .Sum(p => p.Age)
+                .ValidateSynchronization()
+                .RecordValues(out var results);
+
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.HasCompleted.Should().BeTrue("the source was already completed at the time of subscription");
+            results.RecordedValues.Should().BeEmpty("no items were added so no sum values should have been emitted");
+        }
+
+        [Fact]
         public void SourceErrorsAfterEmitting_ErrorPropagates()
         {
             using var source = new TestSourceCache<Person, string>(p => p.Name);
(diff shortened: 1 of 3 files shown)


src/DynamicData.Tests/AggregationTests/SumFixture.cs
```diff
@@ -1,227 +1,786 @@
...
+            });
+
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.HasCompleted.Should().BeFalse("the source can still publish notifications");
+            results.RecordedValues.Should().ContainSingle("a batched edit should produce exactly one sum emission")
+                .Which.Should().Be(60, "the sum of ages 10 + 20 + 30 is 60");
+        }
+
+        [Fact]
+        public void NoItemsAdded_NoSumEmitted()
```
JakenVeina: This reflects what the current behavior is? If so, it's one of the things that should be considered for a change. The most compelling argument, I think, is that `Enumerable.Sum()` emits `0` on an empty source, rather than, say, throwing an exception, so that's probably what consumers are going to expect, intuitively.

Not asking for a change, if this is the current behavior, that's what the test should reflect. Just musing about something to consider in the next phase.

JakenVeina: Me, I'd probably use `SourceIsEmpty` to describe the scenario here. It's slightly clearer about what's really being tested.

src/DynamicData.Tests/AggregationTests/SumFixture.ForCache.cs
```diff
@@ -1,227 +1,786 @@
...
+                .ValidateSynchronization()
+                .RecordValues(out var results);
+
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.HasCompleted.Should().BeFalse("the source can still publish notifications");
+            results.RecordedValues.Should().BeEmpty("no items were added so no sum values should have been emitted");
+        }
+
+        [Fact]
+        public void AllItemsRemoved_SumReturnsToZero()
```
JakenVeina, line 151: Yeah, this definitely confirms that we should be emitting `0` upon subscription to an empty source.

src/DynamicData.Tests/AggregationTests/SumFixture.cs
```diff
@@ -1,227 +1,786 @@
...
+            results.HasCompleted.Should().BeFalse("the source can still publish notifications");
+            results.RecordedValues.Should().ContainSingle("an AddRange produces a single changeset")
+                .Which.Should().Be(expectedSum, $"the sum of the first {itemCount} items should be {expectedSum}");
+        }
+
+        [Theory]
+        [InlineData(0, 50)]
+        [InlineData(1, 40)]
+        [InlineData(2, 30)]
+        public void ItemIsRemoved_SumDecreases(int removalIndex, int expectedSum)
```
JakenVeina: Probably `SumReflectsRemoval` works better, and matches what you've used elsewhere. This actually highlights that you should probably add test cases for negative values, for completeness' sake.

src/DynamicData.Tests/AggregationTests/SumFixture.cs
```diff
@@ -1,227 +1,786 @@
...
+                .Sum(p => (float?)p.Age)
+                .RecordValues(out var results);
+
+            results.RecordedValues[^1].Should().Be(60F, "the nullable float sum of ages 10 + 20 + 30 is 60");
+        }
     }
 
-    [Fact]
-    public void AddedItemsContributeToSumDouble()
+    public class ListSource
```
JakenVeina: Definitely split these into separate files. To match other operators where we've got multiple overloads, I'd do `SumFixture.ForCache.cs` and `SumFixture.ForList.cs`. Check out (if I recall correctly) `ToObservableChangeSetFixture`.

src/DynamicData.Tests/AggregationTests/SumFixture.ForCache.cs
```diff
@@ -1,227 +1,786 @@
...
+            source.Edit(updater => updater.Clear());
+
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.HasCompleted.Should().BeFalse("the source can still publish notifications");
+            results.RecordedValues.Should().HaveCount(2, "one additional sum value should have been emitted after clearing");
+            results.RecordedValues[^1].Should().Be(0, "all items were removed so the sum should return to zero");
+        }
+
+        [Fact]
+        public void SourceCompletesAfterEmitting_CompletionPropagates()
```
JakenVeina, line 178: Add equivalent tests for `SourceCompletesImmediately` and `SourceFailsImmediately`.

Quite often, these scenarios need a little bit of custom logic, to ensure correct behavior, especially for operators that emit an initial value, or that should always emit a value when the source emits one.

In the case of `Sum()`, a source can exist that emits an initial changeset, and then a completion, because no new item changes are allowed, and these would both occur immediately upon subscription. We'd want to make sure that `.Sum()` also emits an initial sum value, for those items, and then a completion, both immediately upon subscription.

src/DynamicData.Tests/AggregationTests/SumFixture.cs
```diff
@@ -1,227 +1,786 @@
...
+                .RecordValues(out var results);
+
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.HasCompleted.Should().BeFalse("the source can still publish notifications");
+            results.RecordedValues.Should().ContainSingle("one changeset was published containing all pre-existing items")
+                .Which.Should().Be(40, "null values should be treated as zero, so the sum should be 10 + 0 + 30 = 40");
+        }
+
+        [Fact]
+        public void ItemsAreAdded_SumIsCorrect_ForInt()
```
JakenVeina: Since we're targeting specific numeric types, we should probably add some test cases for `.MaxValue` and `.MinValue` to guarantee no overflows occurring where there shouldn't be. Maybe as new test cases to these existing tests, maybe as dedicated test methods.

JakenVeina (changes requested):
Looks solid, overall.  A few nitpicks, and a few suggested new tests/test cases.

src/DynamicData.Tests/AggregationTests/SumFixture.cs
giard-alexandre: I 100% agree with this and have hit up against this before. Sure, `StartWith(0)` is easy enough to add but yeah, it's confusing that Count doesnt default to 0.

giard-alexandre:
Alright sorry for the extended delay in getting this pushed out, long story but I moved countries so free time was at a premium.

 @JakenVeina Everything mentioned above should have been addressed, I'll let you resolve the conversations if it's done to your liking!

JakenVeina (approved):

PR #1077, opened 2026-04-10
Update SwappableLock to support NET9+ Lock type

dwcullop:
## Problem

`SwappableLock` is a `ref struct` used internally by `Filter.Dynamic` to atomically swap between lock objects during filtered changeset processing. It originally only supported `Monitor.Enter`/`Monitor.Exit` on `object` gates.

.NET 9 introduced `System.Threading.Lock`: a dedicated, more efficient lock type that is non-reentrant and does not support `Monitor.Enter`. Code paths that use `Lock` on .NET 9 were incompatible with `SwappableLock`.

## Approach

Instead of sprinkling `#if NET9_0_OR_GREATER` throughout every method (mixing `Lock` and `object` state in the same struct), we split `SwappableLock` into **two complete, independent implementations** behind a single top-level `#if`:

- **NET9+**: Uses `System.Threading.Lock` directly. Simpler (no `_hasLock` bool needed, `Lock` tracks its own state).
- **Pre-NET9**: Uses `Monitor.Enter`/`Exit` on `object` gates. Unchanged behavior from the original.

Each implementation is self-contained and readable without mental `#if` gymnastics.

## Changes

- **`Internal/SwappableLock.cs`** - Two complete implementations behind top-level `#if`. NET9 version uses `Lock`.
- **`Cache/Internal/Filter.Dynamic.cs`** - Gate properties return `Lock` on NET9, `object` otherwise. Dedicated `Lock` fields on NET9.
- **`Tests/Internal/SwappableLockFixture.cs`** - New. 7 tests covering CreateAndEnter, Dispose (release + idempotent), SwapTo (basic + chained), uninitialized throws, and Dispose after swap.

Zero behavioral change on pre-.NET 9 targets.

### DIFF
--- a/src/DynamicData/Internal/SwappableLock.cs
+++ b/src/DynamicData/Internal/SwappableLock.cs
@@ -4,6 +4,40 @@
 
 namespace DynamicData;
 
+#if NET9_0_OR_GREATER
+
+internal ref struct SwappableLock
+{
+    public static SwappableLock CreateAndEnter(Lock gate)
+    {
+        gate.Enter();
+        return new SwappableLock { _gate = gate };
+    }
+
+    public void SwapTo(Lock gate)
+    {
+        if (_gate is null)
+            throw new InvalidOperationException("Lock is not initialized");
+
+        gate.Enter();
+        _gate.Exit();
+        _gate = gate;
+    }
+
+    public void Dispose()
+    {
+        if (_gate is not null)
+        {
+            _gate.Exit();
+            _gate = null;
+        }
+    }
+
+    private Lock? _gate;
+}
+
+#else
+
 internal ref struct SwappableLock
 {
     public static SwappableLock CreateAndEnter(object gate)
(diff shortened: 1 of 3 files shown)


src/DynamicData/Internal/SwappableLock.cs
```diff
@@ -45,4 +95,8 @@ public void Dispose()
 
     private bool _hasLock;
     private object? _gate;
+
+#if NET9_0_OR_GREATER
```
JakenVeina: I think we really need to split the whole class for .NET 9 versus earlier versions.

It's weird enough to have both an `object?` gate and a `Lock?` gate within the same instance, when you'd never actually want to use both, but then it's even weirder that the API surface of `SwappableLock` allows it. So, you could end up accidentally having the same `SwappableLock` instance track both an `object`-based and a `Lock`-based lock at the same time, and get some really weird behavior.

Either we should just add a whole new `SwappableThreadingLock` class for `System.Threading.Lock` or actually DISABLE the `object`-based mechanics when `System.Threading.Lock` is available, and force all consumers to swap to using `System.Threading.Lock` for .NET 9+.

Since this is an internal utility, I kinda like option 2. Just only use `Lock`-based locks.

JakenVeina (changes requested):

src/DynamicData/Internal/SwappableLock.cs
RolandPheasant: I prefer option 2 aswell.

dwcullop: Yeah, a bridge too far.  I was also thinking we could make a Net9-Lock compatible drop-in replacement that we can use instead of "object" for pre-Net9 TFMs, so Option 2 aligns with that as well.

dwcullop: Done.  `SwappableLock` will EITHER be for `object` OR for `Lock` (depending on TFM).  Note that `Filter.Dynamic` was using `object` on all TFMs, so I changed it to use `Lock` on Net9.

JakenVeina (approved):

PR #1080, opened 2026-04-14
Overhaul XML documentation for ObservableCacheEx extension methods

dwcullop:
### What

Complete rewrite of XML doc comments across all three `ObservableCacheEx` partial class files, covering ~110 unique operators and ~306 total overloads. Every operator now has publication-quality documentation suitable for extraction to the ReactiveUI docs site.

### Why

The existing XML comments were shallow and generic ("Filters the specified source", "The source", "An observable which emits change sets"). Developers had to read the implementation source to understand how operators actually behave, especially around Refresh handling, error propagation, and completion semantics. This is a reactive collections library where the behavioral details matter: a Refresh event flowing through Filter can become an Add, Remove, forwarded Refresh, or get dropped entirely, and that's not something you can guess from a one-line summary.

### What changed

**Event behavior tables**: 77 operators now have a structured `<list type="table">` documenting how each event (Add, Update, Remove, Refresh, OnError, OnCompleted) is handled. Multi-source operators (joins, MergeManyChangeSets, FilterOnObservable, GroupOnObservable, TransformOnObservable) have separate tables per source, so developers can see the parent-side and child-side/per-item-observable behavior independently.

**Cross-references**: 129 `<seealso>` tags linking related operators: safe variants (Transform/TransformSafe), async variants (Transform/TransformAsync), similar operators (Filter/FilterImmutable/FilterOnObservable), complementary operators (Filter+AutoRefresh, Sort+Bind), and commonly confused operators (MergeMany vs MergeManyChangeSets vs MergeChangeSets).

**Secondary overloads**: 107 `<inheritdoc>` tags with delta `<remarks>` for overloads that delegate to a primary, so the behavioral docs aren't duplicated but every overload is still independently browsable.

**"Worth noting" sections**: 36 operators have a labeled section calling out non-obvious behavior: OnItemRemoved firing for all tracked items on disposal, MergeChangeSets emitting an Update (not Add) when falling back to another source's value, BatchIf losing buffered data if the source errors while paused, per-item observable operators requiring at least one emission before items become visible, etc.

**Summaries trimmed**: Every summary is now 1-3 sentences describing what the operator does. All behavioral detail lives in `<remarks>`.

### What didn't change

No code changes. No behavioral changes. No new APIs. All existing useful information in comments was preserved (additive only). Tests: 2,211 passed, 0 failed.


RolandPheasant (approved):
Can't properly review as there are so many comments.  However the ones I looked at make much sense  and are a considerable improvement.

PR #1081, opened 2026-04-14
Overhaul XML documentation for ObservableListEx extension methods

dwcullop:
### What

Complete rewrite of XML doc comments for ObservableListEx, covering ~60 unique operators and ~124 total overloads. Every operator now has publication-quality documentation suitable for extraction to the ReactiveUI docs site.

### Why

The existing XML comments were shallow and generic ("Filters source using the specified filter observable predicate", "The source", "An observable which emits the change set"). Developers had to read the implementation source to understand how list operators actually behave, especially around the list-specific change reasons (AddRange, RemoveRange, Moved, Clear) that have no cache equivalent.

### What changed

**Event behavior tables**: 31 operators now have a structured `<list type="table">` documenting how each `ListChangeReason` event is handled, including the list-specific reasons: **Add**, **AddRange**, **Replace**, **Remove**, **RemoveRange**, **Moved**, **Refresh**, **Clear**, **OnError**, and **OnCompleted**. Multi-source operators (FilterOnObservable, MergeManyChangeSets) have separate parent-side and per-item-observable tables.

**Cross-references**: 157 `<seealso>` tags linking related operators: async variants (Transform/TransformAsync), similar operators (Filter/FilterOnObservable/FilterOnProperty), complementary operators (Filter+AutoRefresh, Sort+Bind), and commonly confused operators (MergeMany vs MergeManyChangeSets vs MergeChangeSets).

**Secondary overloads**: 49 `<inheritdoc>` tags with delta `<remarks>` for overloads that delegate to a primary.

**"Worth noting" sections**: 20 operators have a labeled section calling out non-obvious behavior: OnItemRemoved firing for all tracked items on disposal, Transform silently appending when Add index is out of bounds, MergeChangeSets forwarding Replace via ReplaceOrAdd (not decomposing into Remove+Add), per-item observable operators requiring at least one emission before items become visible, ListFilterPolicy differences (CalculateDiff vs ClearAndReplace), and more.

**Summaries trimmed**: Every summary is 1-3 sentences describing what the operator does. All behavioral detail lives in `<remarks>`.

### Accuracy verification

Three review agents cross-checked the documentation against the actual implementation code:

- 10 operators traced end-to-end against their implementation files (Filter, Transform, OnItemRemoved, MergeChangeSets, Sort, GroupOn, etc.)
- 2 inaccuracies caught and fixed during review: MergeChangeSets Replace behavior (was incorrectly documented as Remove+Add, actually uses ReplaceOrAdd), Transform Moved exception behavior (undocumented `UnspecifiedIndexException`)
- 7 operators confirmed accurate on first pass

### DIFF
--- a/src/DynamicData/List/ObservableListEx.cs
+++ b/src/DynamicData/List/ObservableListEx.cs
@@ -1890,15 +1890,16 @@ public static class ObservableListEx
     }
 
     /// <summary>
-    /// Strips index information from all changes in the stream using <c>YieldWithoutIndex</c>.
+    /// Strips index information from all changes in the stream.
     /// </summary>
     /// <typeparam name="T">The type of the object.</typeparam>
     /// <param name="source">The source list changeset stream.</param>
     /// <returns>A list changeset stream with all index values removed from changes.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="source"/> is <c>null</c>.</exception>
     /// <remarks>
-    /// <para>This is a workaround operator introduced for internal use (e.g., creating Or operators via MergeMany). Most consumers should not need this.</para>
+    /// <para>Removes index positions from every change in each changeset. This is useful when downstream operators do not require or support index-based operations.</para>
     /// </remarks>
+    /// <seealso cref="ChangeSetEx.YieldWithoutIndex{T}(IEnumerable{Change{T}})"/>
     public static IObservable<IChangeSet<T>> RemoveIndex<T>(this IObservable<IChangeSet<T>> source)
         where T : notnull
     {
@@ -2686,7 +2687,7 @@ public static class ObservableListEx
 
     /// <summary>
     /// Filters the changeset stream to include only changes with the specified <see cref="ListChangeReason"/> values.
-    /// Index information is stripped from the output (via <c>YieldWithoutIndex</c>).
+    /// Index information is stripped from the output because removing some changes invalidates the original index positions.
     /// </summary>
     /// <typeparam name="T">The type of the item.</typeparam>
     /// <param name="source">The source list changeset stream.</param>
@@ -2707,6 +2708,7 @@ public static class ObservableListEx
     /// </remarks>
     /// <seealso cref="WhereReasonsAreNot{T}(IObservable{IChangeSet{T}}, ListChangeReason[])"/>
     /// <seealso cref="SuppressRefresh{T}(IObservable{IChangeSet{T}})"/>
+    /// <seealso cref="ChangeSetEx.YieldWithoutIndex{T}(IEnumerable{Change{T}})"/>
     public static IObservable<IChangeSet<T>> WhereReasonsAre<T>(this IObservable<IChangeSet<T>> source, params ListChangeReason[] reasons)
         where T : notnull
     {
@@ -2728,7 +2730,8 @@ public static class ObservableListEx
 
     /// <summary>
     /// Filters the changeset stream to exclude changes with the specified <see cref="ListChangeReason"/> values.
-    /// Index information is stripped from the output (via <c>YieldWithoutIndex</c>), except when only filtering out <see cref="ListChangeReason.Refresh"/>.
+    /// Index information is stripped from the output because removing some changes invalidates the original index positions.
+    /// The exception is when only <see cref="ListChangeReason.Refresh"/> is excluded, since removing Refresh does not affect index calculations.
     /// </summary>
     /// <typeparam name="T">The type of the item.</typeparam>
     /// <param name="source">The source list changeset stream.</param>
@@ -2739,11 +2742,12 @@ public static class ObservableListEx
     /// <remarks>
     /// <para>
     /// Empty changesets (after filtering) are automatically suppressed. When only <see cref="ListChangeReason.Refresh"/> is excluded,
-    /// indices are preserved (no <c>YieldWithoutIndex</c>), since removing Refresh does not affect index calculations.
+    /// indices are preserved, since removing Refresh does not affect index calculations.
     /// </para>
     /// </remarks>
     /// <seealso cref="WhereReasonsAre{T}(IObservable{IChangeSet{T}}, ListChangeReason[])"/>
     /// <seealso cref="SuppressRefresh{T}(IObservable{IChangeSet{T}})"/>
+    /// <seealso cref="ChangeSetEx.YieldWithoutIndex{T}(IEnumerable{Change{T}})"/>
     public static IObservable<IChangeSet<T>> WhereReasonsAreNot<T>(this IObservable<IChangeSet<T>> source, params ListChangeReason[] reasons)
         where T : notnull
     {
--- a/src/DynamicData/List/ObservableListEx.cs
+++ b/src/DynamicData/List/ObservableListEx.cs
@@ -28,7 +28,7 @@ public static class ObservableListEx
     /// The adaptor's <c>Adapt</c> method is called for each changeset under a synchronized lock, then the changeset is forwarded downstream.
     /// </summary>
     /// <typeparam name="T">The type of items in the list.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="adaptor">The <see cref="IChangeSetAdaptor{T}"/> adaptor whose <c>Adapt</c> method is invoked for each changeset.</param>
     /// <returns>A list changeset stream identical to the source, with the adaptor side effect applied.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="source"/> or <paramref name="adaptor"/> is <see langword="null"/>.</exception>
@@ -71,7 +71,7 @@ public static class ObservableListEx
     /// </summary>
     /// <typeparam name="TObject">The type of items in the list.</typeparam>
     /// <typeparam name="TKey">The type of the key.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TObject}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="keySelector">A <see cref="Func{T, TResult}"/> function to extract a unique key from each item.</param>
     /// <returns>A cache changeset stream (<see cref="IObservable{T}"/> of <see cref="IChangeSet{TObject, TKey}"/>) with keyed items.</returns>
     /// <remarks>
@@ -96,8 +96,8 @@ public static class ObservableListEx
     /// Only items present in ALL sources appear in the result.
     /// </summary>
     /// <typeparam name="T">The type of items in the lists.</typeparam>
-    /// <param name="source">The primary source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
-    /// <param name="others">Additional <see cref="IObservable{T}"/> changeset streams to intersect with.</param>
+    /// <param name="source">The primary source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="others">The additional <see cref="IObservable{IChangeSet{T}}"/> changeset streams to intersect with.</param>
     /// <returns>A list changeset stream containing items that exist in every source.</returns>
     /// <remarks>
     /// <para>
@@ -141,13 +141,13 @@ public static class ObservableListEx
         where T : notnull => sources.Combine(CombineOperator.And);
 
     /// <inheritdoc cref="And{T}(IObservable{IChangeSet{T}}, IObservable{IChangeSet{T}}[])"/>
-    /// <param name="sources">An <see cref="IObservableList{T}"/> of <see cref="IObservableList{T}"/>. Each inner list's changes are connected automatically.</param>
+    /// <param name="sources">An <see cref="IObservableList{IObservableList{T}}"/> of <see cref="IObservableList{IObservableList{T}}"/>. Each inner list's changes are connected automatically.</param>
     /// <remarks>This overload accepts <see cref="IObservableList{T}"/> instances directly, calling <c>Connect()</c> internally.</remarks>
     public static IObservable<IChangeSet<T>> And<T>(this IObservableList<IObservableList<T>> sources)
         where T : notnull => sources.Combine(CombineOperator.And);
 
     /// <inheritdoc cref="And{T}(IObservable{IChangeSet{T}}, IObservable{IChangeSet{T}}[])"/>
-    /// <param name="sources">An <see cref="IObservableList{T}"/> of <see cref="ISourceList{T}"/>. Each inner list's changes are connected automatically.</param>
+    /// <param name="sources">An <see cref="IObservableList{ISourceList{T}}"/> of <see cref="ISourceList{T}"/>. Each inner list's changes are connected automatically.</param>
     /// <remarks>This overload accepts <see cref="ISourceList{T}"/> instances directly, calling <c>Connect()</c> internally.</remarks>
     public static IObservable<IChangeSet<T>> And<T>(this IObservableList<ISourceList<T>> sources)
         where T : notnull => sources.Combine(CombineOperator.And);
@@ -172,7 +172,7 @@ public static class ObservableListEx
     /// The list is kept in sync with the source stream for the lifetime of the subscription.
     /// </summary>
     /// <typeparam name="T">The type of items in the list.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <returns>A read-only observable list reflecting the current state of the stream.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="source"/> is <see langword="null"/>.</exception>
     /// <seealso cref="AsObservableList{T}(ISourceList{T})"/>
@@ -189,9 +189,9 @@ public static class ObservableListEx
     /// changes when any property changes, causing downstream operators to re-evaluate.
     /// </summary>
     /// <typeparam name="TObject">The type of items, which must implement <see cref="INotifyPropertyChanged"/>.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
-    /// <param name="changeSetBuffer">Optional <see cref="TimeSpan"/> buffer duration to batch multiple refresh signals into a single changeset.</param>
-    /// <param name="propertyChangeThrottle">Optional <see cref="TimeSpan"/> throttle applied to each item's property change notifications.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TObject}}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="changeSetBuffer">An optional <see cref="TimeSpan"/> buffer duration to batch multiple refresh signals into a single changeset.</param>
+    /// <param name="propertyChangeThrottle">An optional <see cref="TimeSpan"/> throttle applied to each item's property change notifications.</param>
     /// <param name="scheduler">The scheduler for throttle and buffer timing. Defaults to <see cref="GlobalConfig.DefaultScheduler"/>.</param>
     /// <returns>A list changeset stream with additional <b>Refresh</b> changes injected when properties change.</returns>
     /// <remarks>
@@ -238,10 +238,10 @@ public static class ObservableListEx
     /// <inheritdoc cref="AutoRefresh{TObject}(IObservable{IChangeSet{TObject}}, TimeSpan?, TimeSpan?, IScheduler?)"/>
     /// <typeparam name="TObject">The type of items in the list. Must implement <see cref="INotifyPropertyChanged"/>.</typeparam>
     /// <typeparam name="TProperty">The type of the monitored property.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TObject}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="propertyAccessor">An <see cref="Expression{TDelegate}"/> expression selecting the specific property to monitor.</param>
-    /// <param name="changeSetBuffer">Optional <see cref="TimeSpan"/> buffer duration to batch refresh signals.</param>
-    /// <param name="propertyChangeThrottle">Optional <see cref="TimeSpan"/> throttle per item's property change notifications.</param>
+    /// <param name="changeSetBuffer">An optional <see cref="TimeSpan"/> buffer duration to batch refresh signals.</param>
+    /// <param name="propertyChangeThrottle">An optional <see cref="TimeSpan"/> throttle per item's property change notifications.</param>
     /// <param name="scheduler">The <see cref="IScheduler"/> for throttle and buffer timing.</param>
     /// <remarks>This overload monitors a single property instead of all properties. More efficient when only one property affects downstream operators.</remarks>
     public static IObservable<IChangeSet<TObject>> AutoRefresh<TObject, TProperty>(this IObservable<IChangeSet<TObject>> source, Expression<Func<TObject, TProperty>> propertyAccessor, TimeSpan? changeSetBuffer = null, TimeSpan? propertyChangeThrottle = null, IScheduler? scheduler = null)
@@ -270,9 +270,9 @@ public static class ObservableListEx
     /// </summary>
     /// <typeparam name="TObject">The type of items in the list.</typeparam>
     /// <typeparam name="TAny">The type emitted by the re-evaluator observable (value is ignored).</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TObject}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="reevaluator">A <see cref="Func{T, TResult}"/> factory that, given an item, returns an observable whose emissions trigger a <b>Refresh</b> for that item.</param>
-    /// <param name="changeSetBuffer">Optional <see cref="TimeSpan"/> buffer duration to batch refresh signals into a single changeset.</param>
+    /// <param name="changeSetBuffer">An optional <see cref="TimeSpan"/> buffer duration to batch refresh signals into a single changeset.</param>
     /// <param name="scheduler">The <see cref="IScheduler"/> for buffering.</param>
     /// <returns>A list changeset stream with additional <b>Refresh</b> changes injected when per-item observables fire.</returns>
     /// <remarks>
@@ -308,7 +308,7 @@ public static class ObservableListEx
     /// Applies changeset mutations to a target <see cref="IObservableCollection{T}"/> for UI data binding.
     /// </summary>
     /// <typeparam name="T">The type of items in the list.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="targetCollection">The <see cref="IObservableCollection{T}"/> target collection to keep in sync.</param>
     /// <param name="resetThreshold">When a changeset exceeds this many changes, the collection is reset instead of applying individual changes.</param>
     /// <returns>A continuation of the source changeset stream (allows further chaining).</returns>
@@ -355,7 +355,7 @@ public static class ObservableListEx
     }
 
     /// <inheritdoc cref="Bind{T}(IObservable{IChangeSet{T}}, IObservableCollection{T}, int)"/>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="targetCollection">The <see cref="IObservableCollection{T}"/> target collection to keep in sync.</param>
     /// <param name="options"><see cref="BindingOptions"/> options controlling reset threshold and other behaviors.</param>
     /// <remarks>This overload accepts a <see cref="BindingOptions"/> struct for fine-grained control over binding behavior.</remarks>
@@ -370,8 +370,8 @@ public static class ObservableListEx
     }
 
     /// <inheritdoc cref="Bind{T}(IObservable{IChangeSet{T}}, IObservableCollection{T}, int)"/>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
-    /// <param name="readOnlyObservableCollection">Output parameter receiving the created <see cref="ReadOnlyObservableCollection{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="readOnlyObservableCollection">An output parameter receiving the created <see cref="ReadOnlyObservableCollection{T}"/>.</param>
     /// <param name="resetThreshold">When a changeset exceeds this many changes, the collection is reset.</param>
     /// <remarks>This overload creates a <see cref="ReadOnlyObservableCollection{T}"/> via an <c>out</c> parameter, backed by an internal <c>ObservableCollectionExtended</c>.</remarks>
     public static IObservable<IChangeSet<T>> Bind<T>(this IObservable<IChangeSet<T>> source, out ReadOnlyObservableCollection<T> readOnlyObservableCollection, int resetThreshold = BindingOptions.DefaultResetThreshold)
@@ -390,8 +390,8 @@ public static class ObservableListEx
     }
 
     /// <inheritdoc cref="Bind{T}(IObservable{IChangeSet{T}}, IObservableCollection{T}, int)"/>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
-    /// <param name="readOnlyObservableCollection">Output parameter receiving the created <see cref="ReadOnlyObservableCollection{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="readOnlyObservableCollection">An output parameter receiving the created <see cref="ReadOnlyObservableCollection{T}"/>.</param>
     /// <param name="options"><see cref="BindingOptions"/> options controlling reset threshold and other behaviors.</param>
     /// <remarks>This overload creates a <see cref="ReadOnlyObservableCollection{T}"/> with <see cref="BindingOptions"/> for fine-grained control.</remarks>
     public static IObservable<IChangeSet<T>> Bind<T>(this IObservable<IChangeSet<T>> source, out ReadOnlyObservableCollection<T> readOnlyObservableCollection, BindingOptions options)
@@ -449,10 +449,10 @@ public static class ObservableListEx
     /// Buffers changeset notifications while a pause signal is active, then flushes all buffered changes when resumed.
     /// </summary>
     /// <typeparam name="T">The type of items in the list.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
-    /// <param name="pauseIfTrueSelector">An <see cref="IObservable{T}"/> of <see cref="bool"/> that controls buffering: <see langword="true"/> pauses (buffers), <see langword="false"/> resumes (flushes).</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="pauseIfTrueSelector">An <see cref="IObservable{bool}"/> of <see cref="bool"/> that controls buffering: <see langword="true"/> pauses (buffers), <see langword="false"/> resumes (flushes).</param>
     /// <param name="initialPauseState">The initial pause state. When <see langword="true"/>, buffering starts immediately.</param>
-    /// <param name="timeOut">Optional <see cref="TimeSpan"/> maximum duration to keep the buffer open. After this time, the buffer is flushed regardless of pause state.</param>
+    /// <param name="timeOut">An optional <see cref="TimeSpan"/> maximum duration to keep the buffer open. After this time, the buffer is flushed regardless of pause state.</param>
     /// <param name="scheduler">The <see cref="IScheduler"/> for timeout scheduling.</param>
     /// <returns>A list changeset stream that buffers during pause and emits combined changesets on resume.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="source"/> or <paramref name="pauseIfTrueSelector"/> is <see langword="null"/>.</exception>
@@ -486,7 +486,7 @@ public static class ObservableListEx
     /// Buffers changesets during an initial time window, then emits a single combined changeset and passes through subsequent changes.
     /// </summary>
     /// <typeparam name="TObject">The type of items in the list.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TObject}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="initialBuffer">The <see cref="TimeSpan"/> time period (measured from first emission) during which changes are buffered.</param>
     /// <param name="scheduler">The <see cref="IScheduler"/> for timing the buffer window.</param>
     /// <returns>A list changeset stream where the initial burst is combined into one changeset.</returns>
@@ -511,7 +511,7 @@ public static class ObservableListEx
     /// Casts each item in the changeset from <c>object</c> to <typeparamref name="TDestination"/> using a direct cast.
     /// </summary>
     /// <typeparam name="TDestination">The target type to cast to.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>. of <c>object</c> items.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{object}}"/> of <see cref="IChangeSet{T}"/>. of <c>object</c> items.</param>
     /// <returns>A list changeset stream of cast items.</returns>
     /// <seealso cref="Cast{TSource, TDestination}(IObservable{IChangeSet{TSource}}, Func{TSource, TDestination})"/>
     /// <seealso cref="CastToObject{T}(IObservable{IChangeSet{T}})"/>
@@ -528,7 +528,7 @@ public static class ObservableListEx
     /// </summary>
     /// <typeparam name="TSource">The source item type.</typeparam>
     /// <typeparam name="TDestination">The destination item type.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TSource}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="conversionFactory">A <see cref="Func{T, TResult}"/> function to convert each item from <typeparamref name="TSource"/> to <typeparamref name="TDestination"/>.</param>
     /// <returns>A list changeset stream of converted items.</returns>
     /// <remarks>Use this overload when type inference requires explicit specification of both source and destination types. Alternatively, call <see cref="CastToObject{T}"/> first, then the single-type-parameter <see cref="Cast{TDestination}"/> overload.</remarks>
@@ -549,7 +549,7 @@ public static class ObservableListEx
     /// Casts each item in the changeset to <c>object</c>. Typically used before <see cref="Cast{TDestination}(IObservable{IChangeSet{object}})"/> to work around type inference limitations.
     /// </summary>
     /// <typeparam name="T">The source item type (must be a reference type).</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <returns>A list changeset stream of <c>object</c> items.</returns>
     /// <seealso cref="Cast{TDestination}(IObservable{IChangeSet{object}})"/>
     public static IObservable<IChangeSet<object>> CastToObject<T>(this IObservable<IChangeSet<T>> source)
@@ -559,7 +559,7 @@ public static class ObservableListEx
     /// Applies each changeset to the target list as a side effect, keeping it synchronized with the source.
     /// </summary>
     /// <typeparam name="T">The type of items in the list.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="target">The <see cref="IList{T}"/> target list to clone changes into.</param>
     /// <returns>A continuation of the source changeset stream.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="source"/> is <see langword="null"/>.</exception>
@@ -582,7 +582,7 @@ public static class ObservableListEx
     /// </summary>
     /// <typeparam name="TObject">The type of the object.</typeparam>
     /// <typeparam name="TDestination">The type of the destination.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TObject}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="conversionFactory">The <see cref="Func{T, TResult}"/> conversion factory.</param>
     /// <returns>An observable which emits the change set.</returns>
     [Obsolete("Prefer Cast as it is does the same thing but is semantically correct")]
@@ -601,7 +601,7 @@ public static class ObservableListEx
     /// Defers downstream delivery until the source emits its first changeset, then forwards all subsequent changesets.
     /// </summary>
     /// <typeparam name="T">The type of the object.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <returns>A list changeset stream that begins emitting only after the source has produced its first changeset.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="source"/> is <see langword="null"/>.</exception>
     /// <remarks>
@@ -644,7 +644,7 @@ public static class ObservableListEx
     /// All remaining tracked items are disposed when the stream finalizes (OnCompleted, OnError, or subscription disposal).
     /// </summary>
     /// <typeparam name="T">The type of the object.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <returns>A continuation of the source changeset stream with disposal side effects applied.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="source"/> is <see langword="null"/>.</exception>
     /// <remarks>
@@ -679,7 +679,7 @@ public static class ObservableListEx
     /// </summary>
     /// <typeparam name="TObject">The type of items in the source list.</typeparam>
     /// <typeparam name="TValue">The type of distinct values produced.</typeparam>
-    /// <param name="source">The source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TObject}}"/> of <see cref="IChangeSet{T}"/>.</param>
     /// <param name="valueSelector">A <see cref="Func{T, TResult}"/> function that extracts the value to track from each source item.</param>
     /// <returns>A list changeset stream of distinct values.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="source"/> or <paramref name="valueSelector"/> is <see langword="null"/>.</exception>
@@ -718,8 +718,8 @@ public static class ObservableListEx
     /// Items present in the first source but not in any of the <paramref name="others"/> are included in the result.
     /// </summary>
     /// <typeparam name="T">The type of the item.</typeparam>
-    /// <param name="source">The primary source <see cref="IObservable{T}"/> of <see cref="IChangeSet{T}"/>.</param>
-    /// <param name="others">The other <see cref="IObservable{T}"/> changeset streams to exclude from the result.</param>
+    /// <param name="source">The primary source <see cref="IObservable{IChangeSet{T}}"/> of <see cref="IChangeSet{T}"/>.</param>
+    /// <param name="others">The other <see cref="IObservable{IChangeSet{T}}"/> changeset streams to exclude from the result.</param>
     /// <returns>A list changeset stream containing items from <paramref name="source"/> that are not in any of <paramref name="others"/>.</returns>
     /// <exception cref="ArgumentNullException"><paramref name="others"/> is <see langword="null"/>.</exception>
     /// <remarks>
@@ -788,7 +788,7 @@ public static class ObservableListEx
     /// <typeparam name="T">The type of the item.</typeparam>
     /// <param name="source">The <see cref="ISourceList{T}"/> source list to monitor and remove expired items from.</param>
     /// <param name="timeSelector">A <see cref="Func{T, TResult}"/> function returning the time-to-live for each item. Return <see langword="null"/> for items that should never expire.</param>
-    /// <param name="pollingInterval">Optional <see cref="TimeSpan"/> polling interval to batch expiry checks. If omitted, a separate timer is created for each unique expiry time.</param>
+    /// <param name="pollingInterval">An optional <see cref="TimeSpan"/> polling interval to batch expiry checks. If omitted, a separate timer is created for each unique expiry time.</param>
     /// <param name="scheduler">The scheduler for scheduling expiry timers. Defaults to <see cref="GlobalConfig.DefaultScheduler"/>.</param>
     /// <returns>An observable that emits collections of items each time expired items are removed from the source list.</returns>
     /// <remarks>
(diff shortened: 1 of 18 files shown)


src/DynamicData/List/ObservableListEx.cs
```diff
@@ -24,17 +24,29 @@ namespace DynamicData;
...
-    /// adaptor.
-    /// </exception>
+    /// <typeparam name="T">The type of items in the list.</typeparam>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> to observe and adapt.</param>
+    /// <param name="adaptor">The <see cref="IChangeSetAdaptor{T}"/> adaptor whose <c>Adapt</c> method is invoked for each changeset.</param>
+    /// <returns>A list changeset stream identical to the source, with the adaptor side effect applied.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="source"/> or <paramref name="adaptor"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// <para>
+    /// This operator synchronizes access with a lock, then calls <c>adaptor.Adapt(changes)</c> before forwarding the changeset.
```
JakenVeina: A single-input operator that doesn't do any scheduling should not be doing any locking. If it actually is doing locking, it's superfluous and should be removed. In general, I don't think we should be documenting unintended behavior.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -24,17 +24,29 @@ namespace DynamicData;
...
+    /// It is the primary extension point for custom UI binding adaptors (e.g., <see cref="Bind{T}(IObservable{IChangeSet{T}}, IObservableCollection{T}, BindingOptions)"/>
+    /// delegates to this operator).
+    /// </para>
+    /// <list type="table">
+    /// <listheader><term>Event</term><description>Behavior</description></listheader>
+    /// <item><term>Add/AddRange/Replace/Remove/RemoveRange/Moved/Refresh/Clear</term><description>The adaptor's <c>Adapt</c> method is called with the full changeset, then it is forwarded downstream unchanged.</description></item>
+    /// <item><term>OnError</term><description>Forwarded to the downstream observer. If the adaptor throws, the exception propagates as OnError.</description></item>
```
JakenVeina: I do still think this is pointless overkill. What we end up with is identical copy/pasted `OnError` and `OnCompleted` listings for virtually every operator, which ends up just becoming distracting noise for readers. On the rare occasion that these notifications don't just flow immediately downstream, sure, document that.

Given that this is already in place on the cache side, I can be okay with leaving it for now.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -24,17 +24,29 @@ namespace DynamicData;
...
+    /// <item><term>OnCompleted</term><description>Forwarded to the downstream observer.</description></item>
+    /// </list>
+    /// </remarks>
+    /// <seealso cref="Bind{T}(IObservable{IChangeSet{T}}, IObservableCollection{T}, BindingOptions)"/>
+    /// <seealso cref="Clone{T}(IObservable{IChangeSet{T}}, IList{T})"/>
```
JakenVeina: What's the relevance of referencing `.Clone()` here?

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -24,17 +24,29 @@ namespace DynamicData;
...
+    /// <exception cref="ArgumentNullException"><paramref name="source"/> or <paramref name="adaptor"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// <para>
+    /// This operator synchronizes access with a lock, then calls <c>adaptor.Adapt(changes)</c> before forwarding the changeset.
+    /// It is the primary extension point for custom UI binding adaptors (e.g., <see cref="Bind{T}(IObservable{IChangeSet{T}}, IObservableCollection{T}, BindingOptions)"/>
+    /// delegates to this operator).
+    /// </para>
+    /// <list type="table">
+    /// <listheader><term>Event</term><description>Behavior</description></listheader>
+    /// <item><term>Add/AddRange/Replace/Remove/RemoveRange/Moved/Refresh/Clear</term><description>The adaptor's <c>Adapt</c> method is called with the full changeset, then it is forwarded downstream unchanged.</description></item>
```
JakenVeina: Listing the handling behavior of every type of change is redundant if all changes are handled identically, and we've already stated what that handling behavior is.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -76,13 +92,34 @@ public static IObservable<IChangeSet<TObject, TKey>> AddKey<TObject, TKey>(this
...
-    /// Items which are in all of the sources are included in the result.
+    /// Applies a logical AND (intersection) between multiple list changeset streams.
+    /// Only items present in ALL sources appear in the result.
     /// </summary>
-    /// <typeparam name="T">The type of the item.</typeparam>
-    /// <param name="source">The source.</param>
-    /// <param name="others">The others.</param>
-    /// <returns>An observable which emits the change set.</returns>
+    /// <typeparam name="T">The type of items in the lists.</typeparam>
+    /// <param name="source">The primary source <see cref="IObservable{IChangeSet{T}}"/> to intersect.</param>
```
JakenVeina: Not sure that I like labeling the source as "primary". Intersect operations are, by nature, commutative, there is no "primary". You could say "the first of two" or just "the first".

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -76,13 +92,34 @@ public static IObservable<IChangeSet<TObject, TKey>> AddKey<TObject, TKey>(this
...
+    /// <returns>A list changeset stream containing items that exist in every source.</returns>
+    /// <remarks>
+    /// <para>
+    /// Uses reference counting per item across all sources. An item appears downstream only when
+    /// its reference count is non-zero in ALL sources. Item identity is determined by the default equality comparer.
+    /// </para>
+    /// <list type="table">
+    /// <listheader><term>Event</term><description>Behavior</description></listheader>
+    /// <item><term>Add/AddRange</term><description>The item's reference count is incremented in its source tracker. If the item is now present in all sources, an <b>Add</b> is emitted.</description></item>
+    /// <item><term>Replace</term><description>The old item is removed from the tracker and the new item is added. Membership is recalculated for both.</description></item>
```
JakenVeina: Should maybe also mention that it will emit an `Add`, `Remove`, `Replace`, or nothing

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -76,13 +92,34 @@ public static IObservable<IChangeSet<TObject, TKey>> AddKey<TObject, TKey>(this
...
+    /// <item><term>Moved</term><description>Ignored (set operations are position-independent).</description></item>
+    /// <item><term>OnError</term><description>Forwarded from any source.</description></item>
+    /// <item><term>OnCompleted</term><description>Forwarded when all sources complete.</description></item>
+    /// </list>
+    /// <para><b>Worth noting:</b> Item identity uses object equality, not position. Duplicate items in a single source are reference-counted independently.</para>
+    /// </remarks>
+    /// <seealso cref="Or{T}(IObservable{IChangeSet{T}}, IObservable{IChangeSet{T}}[])"/>
+    /// <seealso cref="Except{T}(IObservable{IChangeSet{T}}, IObservable{IChangeSet{T}}[])"/>
+    /// <seealso cref="Xor{T}(IObservable{IChangeSet{T}}, IObservable{IChangeSet{T}}[])"/>
+    /// <seealso cref="ObservableCacheEx.And"/>
```
JakenVeina: Shouldn't `ObservableCacheEx.And` here also reference the method's parameter types?

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -91,53 +128,37 @@ public static IObservable<IChangeSet<T>> And<T>(this IObservable<IChangeSet<T>>
...
-    /// <summary>
-    /// Apply a logical And operator between the collections.
-    /// Items which are in all of the sources are included in the result.
-    /// </summary>
-    /// <typeparam name="T">The type of the item.</typeparam>
-    /// <param name="sources">The sources.</param>
-    /// <returns>An observable which emits the change set.</returns>
+    /// <inheritdoc cref="And{T}(IObservable{IChangeSet{T}}, IObservable{IChangeSet{T}}[])"/>
+    /// <param name="sources">A <see cref="ICollection{T}"/> of changeset streams to intersect.</param>
+    /// <remarks>This overload accepts a pre-built collection of sources instead of params array.</remarks>
```
JakenVeina: Does additional remarks on an inherited doc list get parsed as additional remarks, or does it override the original remarks? If it overrides, we should get rid of this, it's not valuable info. I'm assuming overriding is the general behavior, like with the override for the `"sources"` parameter, but it's possible remarks are supposed to be handled differently.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -147,12 +168,14 @@ public static IObservableList<T> AsObservableList<T>(this ISourceList<T> source)
...
-    /// <typeparam name="T">The type of the item.</typeparam>
-    /// <param name="source">The source.</param>
-    /// <returns>An observable list.</returns>
-    /// <exception cref="ArgumentNullException">source.</exception>
+    /// <typeparam name="T">The type of items in the list.</typeparam>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> to materialize into a read-only list.</param>
+    /// <returns>A read-only observable list reflecting the current state of the stream.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="source"/> is <see langword="null"/>.</exception>
+    /// <seealso cref="AsObservableList{T}(ISourceList{T})"/>
     public static IObservableList<T> AsObservableList<T>(this IObservable<IChangeSet<T>> source)
```
JakenVeina, line 190: Should add in either summary or remarks that this allows for subscription multicasting. I.E. it allows for multiple subscriptions to a source, while ensuring that the source operators are only evaluated once. Something to that effect.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -189,17 +235,15 @@ public static IObservable<IChangeSet<TObject>> AutoRefresh<TObject>(this IObserv
...
-    /// <typeparam name="TObject">The type of object.</typeparam>
-    /// <typeparam name="TProperty">The type of property.</typeparam>
-    /// <param name="source">The source observable.</param>
-    /// <param name="propertyAccessor">Specify a property to observe changes. When it changes a Refresh is invoked.</param>
-    /// <param name="changeSetBuffer">Batch up changes by specifying the buffer. This greatly increases performance when many elements have successive property changes.</param>
-    /// <param name="propertyChangeThrottle">When observing on multiple property changes, apply a throttle to prevent excessive refresh invocations.</param>
-    /// <param name="scheduler">The scheduler.</param>
-    /// <returns>An observable change set with additional refresh changes.</returns>
+    /// <inheritdoc cref="AutoRefresh{TObject}(IObservable{IChangeSet{TObject}}, TimeSpan?, TimeSpan?, IScheduler?)"/>
+    /// <typeparam name="TObject">The type of items in the list. Must implement <see cref="INotifyPropertyChanged"/>.</typeparam>
```
JakenVeina: No need to override the doc for `TObject`.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -189,17 +235,15 @@ public static IObservable<IChangeSet<TObject>> AutoRefresh<TObject>(this IObserv
...
-    /// </summary>
-    /// <typeparam name="TObject">The type of object.</typeparam>
-    /// <typeparam name="TProperty">The type of property.</typeparam>
-    /// <param name="source">The source observable.</param>
-    /// <param name="propertyAccessor">Specify a property to observe changes. When it changes a Refresh is invoked.</param>
-    /// <param name="changeSetBuffer">Batch up changes by specifying the buffer. This greatly increases performance when many elements have successive property changes.</param>
-    /// <param name="propertyChangeThrottle">When observing on multiple property changes, apply a throttle to prevent excessive refresh invocations.</param>
-    /// <param name="scheduler">The scheduler.</param>
-    /// <returns>An observable change set with additional refresh changes.</returns>
+    /// <inheritdoc cref="AutoRefresh{TObject}(IObservable{IChangeSet{TObject}}, TimeSpan?, TimeSpan?, IScheduler?)"/>
```
JakenVeina, line 253: `<summary>` should be overridden, to clarify the different behavior of this operator.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -221,15 +265,36 @@ public static IObservable<IChangeSet<TObject>> AutoRefresh<TObject, TProperty>(t
...
+    /// <param name="scheduler">The <see cref="IScheduler"/> for buffering.</param>
+    /// <returns>A list changeset stream with additional <b>Refresh</b> changes injected when per-item observables fire.</returns>
+    /// <remarks>
+    /// <para>
+    /// This is the general-purpose refresh mechanism. <see cref="AutoRefresh{TObject}(IObservable{IChangeSet{TObject}}, TimeSpan?, TimeSpan?, IScheduler?)"/>
+    /// is a convenience wrapper that uses <c>WhenAnyPropertyChanged()</c> as the re-evaluator.
+    /// </para>
+    /// <list type="table">
+    /// <listheader><term>Event</term><description>Behavior</description></listheader>
+    /// <item><term>Add/AddRange</term><description>Subscribes to the re-evaluator observable for each new item via <c>MergeMany</c>. The original change is forwarded.</description></item>
```
JakenVeina: Docs should not be mentioning internal implementation details that are subject to change (I.E. the fact that the operator uses `.MergeMany()` internally). The mention for `AutoRefresh()` above is fine, because, as mentioned there, that's a convenience wrapper.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -221,15 +265,36 @@ public static IObservable<IChangeSet<TObject>> AutoRefresh<TObject, TProperty>(t
...
+    /// <item><term>Replace</term><description>Unsubscribes from the old item's observable, subscribes to the new. The original change is forwarded.</description></item>
+    /// <item><term>Remove/RemoveRange/Clear</term><description>Unsubscribes from removed items. The original change is forwarded.</description></item>
+    /// <item><term>Moved/Refresh</term><description>Forwarded unchanged.</description></item>
+    /// <item><term>Re-evaluator fires</term><description>The item's current index is looked up and a <b>Refresh</b> change is emitted.</description></item>
+    /// <item><term>OnError</term><description>Forwarded from source or from any re-evaluator observable.</description></item>
+    /// <item><term>OnCompleted</term><description>Forwarded when the source completes.</description></item>
+    /// </list>
+    /// <para><b>Worth noting:</b> The internal index lookup (to find where each item is for the Refresh change) requires maintaining a cloned list, adding memory overhead proportional to list size.</para>
```
JakenVeina: The vast majority of our operators do this. Many of them multiple times. Is there a reason to single out this operator in particular for doing so?

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -240,18 +305,38 @@ public static IObservable<IChangeSet<TObject>> AutoRefreshOnObservable<TObject,
...
-    /// source
-    /// or
-    /// targetCollection.
-    /// </exception>
+    /// <typeparam name="T">The type of items in the list.</typeparam>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> to bind to a collection.</param>
+    /// <param name="targetCollection">The <see cref="IObservableCollection{T}"/> target collection to keep in sync.</param>
+    /// <param name="resetThreshold">When a changeset exceeds this many changes, the collection is reset instead of applying individual changes.</param>
+    /// <returns>A continuation of the source changeset stream (allows further chaining).</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="source"/> or <paramref name="targetCollection"/> is <see langword="null"/>.</exception>
```
JakenVeina, line 320: This seems to be the first operator for which intended exceptions are listed. If we want to list intended exceptions (I would say we do), it needs to be on all of them.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -269,19 +354,11 @@ public static IObservable<IChangeSet<T>> Bind<T>(this IObservable<IChangeSet<T>>
...
-    /// <param name="targetCollection">The target collection.</param>
-    /// <param name="options">The binding options.</param>
-    /// <returns>An observable which emits the change set.</returns>
-    /// <exception cref="System.ArgumentNullException">
-    /// source
-    /// or
-    /// targetCollection.
-    /// </exception>
+    /// <inheritdoc cref="Bind{T}(IObservable{IChangeSet{T}}, IObservableCollection{T}, int)"/>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> to bind to a collection.</param>
```
JakenVeina: Override is identical to what's being overridden.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -269,19 +354,11 @@ public static IObservable<IChangeSet<T>> Bind<T>(this IObservable<IChangeSet<T>>
...
+    /// <param name="targetCollection">The <see cref="IObservableCollection{T}"/> target collection to keep in sync.</param>
```
JakenVeina: Override is identical to what's being overridden.

There's a bunch more like this, I'm not gonna spam comments on all of them.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -315,14 +389,11 @@ public static IObservable<IChangeSet<T>> Bind<T>(this IObservable<IChangeSet<T>>
...
 
-    /// <summary>
-    /// Creates a binding to a readonly observable collection which is specified as an 'out' parameter.
-    /// </summary>
-    /// <typeparam name="T">The type of the item.</typeparam>
-    /// <param name="source">The source.</param>
-    /// <param name="readOnlyObservableCollection">The resulting read only observable collection.</param>
-    /// <param name="options">The binding options.</param>
-    /// <returns>A continuation of the source stream.</returns>
+    /// <inheritdoc cref="Bind{T}(IObservable{IChangeSet{T}}, IObservableCollection{T}, int)"/>
```
JakenVeina, line 404: Should override `<summary>` here to explain the purpose of this particular overload, in constructing a target collection for you.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -337,19 +408,11 @@ public static IObservable<IChangeSet<T>> Bind<T>(this IObservable<IChangeSet<T>>
...
-    /// source
-    /// or
-    /// targetCollection.
-    /// </exception>
-    /// <returns>An observable which emits the change set.</returns>
+    /// <inheritdoc cref="Bind{T}(IObservable{IChangeSet{T}}, IObservableCollection{T}, int)"/>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> to bind to a collection.</param>
+    /// <param name="bindingList">The target <see cref="BindingList{T}"/> to keep in sync.</param>
+    /// <param name="resetThreshold">When a changeset exceeds this many changes, the list is reset.</param>
+    /// <remarks>This overload binds to a <see cref="BindingList{T}"/> (WinForms binding). Uses a an internal binding list adaptor internally.</remarks>
```
JakenVeina: Typos & phrasing.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -430,13 +483,21 @@ public static IObservable<IChangeSet<T>> BufferIf<T>(this IObservable<IChangeSet
...
-    /// <param name="scheduler">The scheduler to buffer on.</param>
-    /// <returns>An observable which emits the change set.</returns>
+    /// <typeparam name="TObject">The type of items in the list.</typeparam>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TObject}}"/> to buffer during the initial loading period.</param>
+    /// <param name="initialBuffer">The <see cref="TimeSpan"/> time period (measured from first emission) during which changes are buffered.</param>
+    /// <param name="scheduler">The <see cref="IScheduler"/> for timing the buffer window.</param>
+    /// <returns>A list changeset stream where the initial burst is combined into one changeset.</returns>
+    /// <remarks>
+    /// <para>
+    /// Composed from <see cref="DeferUntilLoaded{T}(IObservable{IChangeSet{T}})"/>, <c>Buffer</c>, and <see cref="FlattenBufferResult{T}"/>.
```
JakenVeina: Internal implementation details. Remove.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -574,18 +675,33 @@ public static IObservable<IChangeSet<T>> DisposeMany<T>(this IObservable<IChange
...
+    /// <item><term><b>Remove</b>/<b>RemoveRange</b></term><description>Reference count decremented. If the count reaches zero, a <b>Remove</b> is emitted for that distinct value.</description></item>
+    /// <item><term><b>Refresh</b></term><description>Value is re-extracted. If changed, old value decremented and new value incremented (same as Replace logic).</description></item>
+    /// <item><term><b>Clear</b></term><description>All reference counts cleared. <b>Remove</b> emitted for every tracked distinct value.</description></item>
+    /// <item><term>OnError</term><description>Forwarded to the downstream observer.</description></item>
+    /// <item><term>OnCompleted</term><description>Forwarded to the downstream observer.</description></item>
+    /// </list>
+    /// </remarks>
+    /// <seealso cref="Filter{T}(IObservable{IChangeSet{T}}, Func{T, bool})"/>
+    /// <seealso cref="GroupOn{TObject, TGroup}(IObservable{IChangeSet{TObject}}, Func{TObject, TGroup}, IObservable{Unit}?)"/>
```
JakenVeina: I'm guessing `.Filter()` and `.GroupOn()` are referenced because they're used internally? If so, remove.

JakenVeina (changes requested):
Will pick back up tomorrow. I got through operators A-D.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -598,13 +714,37 @@ public static IObservable<IChangeSet<TValue>> DistinctValues<TObject, TValue>(th
...
-    /// <param name="source">The source.</param>
-    /// <param name="others">The others.</param>
-    /// <returns>An observable which emits the change set.</returns>
+    /// <param name="source">The primary <see cref="IObservable{IChangeSet{T}}"/> from which other streams are subtracted.</param>
+    /// <param name="others">The other <see cref="IObservable{IChangeSet{T}}"/> changeset streams to exclude from the result.</param>
+    /// <returns>A list changeset stream containing items from <paramref name="source"/> that are not in any of <paramref name="others"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="others"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// <para>
+    /// Uses reference-counted equality comparison across all sources. Items are compared by equality (not index position).
```
JakenVeina: Should probably mention use of the default equality comparer, like `.And()` did.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -613,52 +753,57 @@ public static IObservable<IChangeSet<T>> Except<T>(this IObservable<IChangeSet<T
...
+    /// This operator acts directly on an <see cref="ISourceList{T}"/>, not on a changeset stream. It monitors items as they are added,
+    /// schedules their removal, and physically removes them from the source list when their time expires.
+    /// </para>
+    /// <para>
+    /// When <paramref name="pollingInterval"/> is specified, all items due for removal are batched into a single removal at each polling tick,
+    /// which can improve performance when many items expire around the same time.
+    /// </para>
+    /// <para><b>Worth noting:</b> The returned observable emits the expired items (not changesets). Subscribe to this observable to trigger the expiry mechanism; if not subscribed, no items will be removed.</para>
+    /// </remarks>
+    /// <seealso cref="LimitSizeTo{T}(ISourceList{T}, int, IScheduler?)"/>
```
JakenVeina, line 820: Are these references relevant, or just generated because the operator uses them internally?

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -672,14 +817,34 @@ public static IObservable<IEnumerable<T>> ExpireAfter<T>(
...
-    /// <returns>A list stream, containing only the items matched by <paramref name="predicate"/>.</returns>
-    /// <exception cref="ArgumentNullException">Throws for <paramref name="source"/> and <paramref name="predicate"/>.</exception>
-    /// <remarks>Note that, unlike some other overloads of this operator, ordering of items is preserved.</remarks>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> to filter.</param>
+    /// <param name="predicate">A <see cref="Func{T, TResult}"/> predicate that determines which items are included. Items returning <see langword="true"/> appear downstream; items returning <see langword="false"/> are excluded.</param>
+    /// <returns>A list changeset stream containing only items that satisfy <paramref name="predicate"/>.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="source"/> or <paramref name="predicate"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// <para>Use this overload when the predicate is fixed for the lifetime of the subscription. Item ordering is preserved.</para>
```
JakenVeina: "Use this overload when the predicate is fixed for the lifetime of the subscription." strikes me as ever so slightly misleading. The method signature makes it obvious that the predicate cannot change, so what some people might interpret this as is that the predicate RESULT should not change. Which is simply not true, that would be `.FilterImmutable()` Here, items can have their inclusion or exclusion changed by a refresh.

JakenVeina: I don't see any value in removing the "unlike other overloads of this operator" note.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -690,16 +855,36 @@ public static IObservable<IChangeSet<T>> Filter<T>(
...
-    /// <param name="source">The source.</param>
-    /// <param name="predicate">The predicate which indicates which items should be included.</param>
-    /// <param name="filterPolicy">Should the filter clear and replace, or calculate a diff-set.</param>
-    /// <returns>An observable which emits the change set.</returns>
-    /// <exception cref="ArgumentNullException">source
-    /// or
-    /// filterController.</exception>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> to filter.</param>
+    /// <param name="predicate">An <see cref="IObservable{Func{T, bool}}"/> that emits new predicate functions. Each emission triggers a full re-evaluation of all items.</param>
+    /// <param name="filterPolicy">The <see cref="ListFilterPolicy"/> that controls re-filtering behavior: <see cref="ListFilterPolicy.CalculateDiff"/> (default) computes the minimal diff between old and new results; <see cref="ListFilterPolicy.ClearAndReplace"/> clears and repopulates entirely.</param>
```
JakenVeina: I don't think re-stating the definition of `ListFilterPolicy` here is a good idea. It means that if that enum ever changes, we now have to make updates here as well.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -690,16 +855,36 @@ public static IObservable<IChangeSet<T>> Filter<T>(
...
+    /// <returns>A list changeset stream containing only items that satisfy the most recent predicate.</returns>
+    /// <remarks>
+    /// <para>
+    /// Each time <paramref name="predicate"/> emits, every item is re-evaluated. The <paramref name="filterPolicy"/> controls
```
JakenVeina: This doesn't really add any new information, just re-states info from above.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -690,16 +855,36 @@ public static IObservable<IChangeSet<T>> Filter<T>(
...
+    /// <item><term>Clear</term><description>All downstream items are cleared.</description></item>
+    /// <item><term>Predicate changed</term><description>All items re-evaluated against the new predicate. With <see cref="ListFilterPolicy.CalculateDiff"/>, only items that changed status emit Add/Remove. With <see cref="ListFilterPolicy.ClearAndReplace"/>, a Clear is emitted followed by AddRange of all matching items.</description></item>
+    /// <item><term>OnError</term><description>Forwarded from the source or from <paramref name="predicate"/>.</description></item>
+    /// <item><term>OnCompleted</term><description>Forwarded when the source completes. Independent completion of <paramref name="predicate"/> does not terminate the filter.</description></item>
+    /// </list>
+    /// <para><b>Worth noting:</b> No items are included until <paramref name="predicate"/> emits its first function. <see cref="ListFilterPolicy.CalculateDiff"/> is generally preferred for performance; <see cref="ListFilterPolicy.ClearAndReplace"/> is useful when downstream consumers (like UI bindings) handle full resets more efficiently than individual changes.</para>
```
JakenVeina: The notes on `ListFilterPolicy` should probably be moved to the docs for `ListFilterPolicy`

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -711,20 +896,38 @@ public static IObservable<IChangeSet<T>> Filter<T>(this IObservable<IChangeSet<T
...
-    /// <param name="predicate">A static predicate to be used to determine which items should be included or excluded by the filter.</param>
-    /// <param name="filterPolicy">The policy that the operator should use when performing re-filtering operations.</param>
-    /// <param name="suppressEmptyChangeSets">By default empty changeset notifications are suppressed for performance reasons.  Set to false to publish empty changesets.  Doing so can be useful for monitoring loading status.</param>
-    /// <returns>An observable which emits change sets.</returns>
-    /// <exception cref="ArgumentNullException">Throws for <paramref name="source"/>, <paramref name="predicateState"/>, and <paramref name="predicate"/>.</exception>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> to filter.</param>
+    /// <param name="predicateState">An <see cref="IObservable{TState}"/> stream of state values to be passed to <paramref name="predicate"/>.</param>
+    /// <param name="predicate">A <see cref="Func{T, TResult}"/> predicate receiving the current state and an item, returning <see langword="true"/> to include or <see langword="false"/> to exclude.</param>
```
JakenVeina: Should put the word "static" back in here, describing the predicate. It's important that the predicate be functionally static, even if not literally a `static` method in source.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -711,20 +896,38 @@ public static IObservable<IChangeSet<T>> Filter<T>(this IObservable<IChangeSet<T
...
+    /// <param name="filterPolicy">The <see cref="ListFilterPolicy"/> that controls re-filtering behavior: <see cref="ListFilterPolicy.CalculateDiff"/> (default) computes minimal diff; <see cref="ListFilterPolicy.ClearAndReplace"/> clears and repopulates.</param>
```
JakenVeina: Same as before, don't just re-state the definition of `ListFilterPolicy` everywhere.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -711,20 +896,38 @@ public static IObservable<IChangeSet<T>> Filter<T>(this IObservable<IChangeSet<T
...
+    /// <item><term><b>Remove</b>/<b>RemoveRange</b></term><description>If the item was downstream, a <b>Remove</b> is emitted.</description></item>
+    /// <item><term><b>Refresh</b></term><description>Re-evaluated against current state. Inclusion status may change.</description></item>
+    /// <item><term><b>Clear</b></term><description>All downstream items are cleared.</description></item>
+    /// <item><term>State changed</term><description>All items re-evaluated with new state value. <see cref="ListFilterPolicy.CalculateDiff"/> emits minimal Add/Remove; <see cref="ListFilterPolicy.ClearAndReplace"/> emits Clear then AddRange.</description></item>
+    /// <item><term>OnError</term><description>Forwarded from the source or from <paramref name="predicateState"/>.</description></item>
+    /// <item><term>OnCompleted</term><description>Forwarded when the source completes.</description></item>
+    /// </list>
+    /// <para><b>Worth noting:</b> <paramref name="predicateState"/> should emit an initial value immediately upon subscription. No items are included until the first state value arrives.</para>
```
JakenVeina: Already stated above.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -783,21 +1015,43 @@ public static IObservable<IChangeSet<TObject>> FilterOnProperty<TObject, TProper
     }
 
     /// <summary>
-    /// Convert the result of a buffer operation to a change set.
+    /// Flattens a buffered list of changesets (from Rx's <c>Buffer</c> operator) back into a single changeset stream.
```
JakenVeina: This description isn't really any better than the old one. It doesn't flatten "a" buffered list into a "stream", that's just nonsense.

How about...

> Flattens buffered changesets (E.G. from `System.Reactive.Linq.Observable.Buffer()`) back into single changesets.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -783,21 +1015,43 @@ public static IObservable<IChangeSet<TObject>> FilterOnProperty<TObject, TProper
...
+    /// <para>Use this after applying <c>Observable.Buffer()</c> to a changeset stream to re-merge the batched changesets into a single stream.</para>
+    /// </remarks>
+    /// <seealso cref="BufferIf{T}(IObservable{IChangeSet{T}}, IObservable{bool}, IScheduler?)"/>
+    /// <seealso cref="BufferInitial{T}(IObservable{IChangeSet{T}}, TimeSpan, IScheduler?)"/>
     public static IObservable<IChangeSet<T>> FlattenBufferResult<T>(this IObservable<IList<IChangeSet<T>>> source)
         where T : notnull => source.Where(x => x.Count != 0).Select(updates => new ChangeSet<T>(updates.SelectMany(u => u)));
 
     /// <summary>
-    /// Provides a call back for each item change.
+    /// Invokes <paramref name="action"/> for every <see cref="Change{T}"/> in each changeset, including range changes as-is.
```
JakenVeina: What is "range changes as-is" supposed to mean, here?

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -783,21 +1015,43 @@ public static IObservable<IChangeSet<TObject>> FilterOnProperty<TObject, TProper
...
+    /// The changeset is forwarded downstream unchanged.
     /// </summary>
     /// <typeparam name="TObject">The type of the object.</typeparam>
```
JakenVeina: I probably missed a bunch of these from prior operators, but the question I intuitively have to ask is "What object?" Some of the other operators above have the right idea, it should be something like "The type of items in the list". Whatever it is, the wording should be consistent across all operators.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -783,21 +1015,43 @@ public static IObservable<IChangeSet<TObject>> FilterOnProperty<TObject, TProper
...
-    /// <param name="source">The source.</param>
-    /// <param name="action">The action.</param>
-    /// <returns>An observable which emits the change set.</returns>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TObject}}"/> to observe each change in.</param>
+    /// <param name="action">The action invoked for each <see cref="Change{T}"/>. Range changes (AddRange, RemoveRange, Clear) are received as a single <see cref="Change{T}"/> with a populated <c>Range</c> property.</param>
```
JakenVeina: No need to re-state how `Change<T>` works, unless it's somehow different here.

JakenVeina (changes requested):
Got through E and F today.

src/DynamicData/List/ObservableListEx.cs
dwcullop: People don't see the documentation for multiple operators at once.  Usually it's shown just one at a time in their IDE or each method has it's own HTML page.  Having the duplicated data allows them to vary independently in the future and puts it in one place for consumption.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Right, the lock claim was wrong. Reworded to just say the adaptor is invoked before forwarding downstream.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Yeah you were right. Did a sweep across both files removing OnError/OnCompleted rows that just repeated standard Rx behavior. Only kept them where something useful or unexpected was happening (e.g. MergeManyChangeSets, which waits for all child streams to complete before completing the output).

src/DynamicData/List/ObservableListEx.cs
dwcullop: Dropped.

src/DynamicData/List/ObservableListEx.cs
dwcullop: You called it. Sweep at the end: dropped the Adapt table along with a handful of others (DeferUntilLoaded, QueryWhenChanged, WhereReasonsAre, Refresh/Remove helpers, Switch) where every row was tautological.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Fixed.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Reworded the Replace row.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Done. Did a sweep across the file and fully qualified every `ObservableCacheEx.X` cref with its parameter types. Killed ~70 CS1584/CS1658 warnings in the process.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Both, depending on syntax. A sibling `<remarks>short text</remarks>` next to `<inheritdoc cref=...>` silently overrides the inherited remarks (that's the anti-pattern you're flagging). To append, wrap as `<remarks><inheritdoc cref=...><para>delta</para></remarks>`. The nested inheritdoc pulls in just the parent's remarks and the para adds the delta. Switched every overload that was overriding to the append pattern.

Spec: https://ewsoftware.github.io/XMLCommentsGuide/html/86453FFB-B978-4A2A-9EB5-70E118CA8073.htm

src/DynamicData/List/ObservableListEx.cs
dwcullop: Added.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Agreed. Went further and dropped all explicit typeparam/param tags on this overload and the 4 Bind overloads; new params just get a `<paramref>` mention inside the summary prose. Cref-based inheritdoc covers the rest and compiler stays clean.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Done.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Reworded to behavior, not impl.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Fair point. Generalized.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Did a sweep. Every operator that calls `ThrowArgumentNullExceptionIfNull` now has matching `<exception cref="ArgumentNullException">` tags.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Dropped the redundant param tags and fixed the remarks override.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Same fix.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Done on both out-param overloads.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Fixed both `an an` and the surrounding phrasing.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Rewritten to describe behavior.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Removed. They were impl-only.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Added.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Kept both intentionally. `LimitSizeTo` is the size-counterpart and `ToObservableChangeSet` (the overload taking an `expireAfter` selector) is the time-counterpart. Conceptually related from a user's perspective, not impl artifacts.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Reworded. The predicate function itself is fixed for the lifetime; inclusion can still change via Refresh.

dwcullop: Kept it. Phrased it via the predicate-function-fixed framing above.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Stripped here and in the other Filter overloads. Moved the performance trade-off prose into the enum member docs on `ListFilterPolicy` itself.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Stripped.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Done. `ListFilterPolicy.cs` now carries the `ClearAndReplace` / `CalculateDiff` trade-off notes.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Restored.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Stripped.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Removed.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Took your wording.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Reworded.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Did a TObject doc sweep across all operators in the file.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Dropped the Range-property restatement from the action param.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -952,17 +1273,33 @@ public static IObservable<IEnumerable<T>> LimitSizeTo<T>(this ISourceList<T> sou
...
-    /// observableSelector.</exception>
+    /// <typeparam name="T">The type of items in the source list.</typeparam>
+    /// <typeparam name="TDestination">The type of values emitted by per-item observables.</typeparam>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{T}}"/> whose items each produce an observable.</param>
+    /// <param name="observableSelector">A <see cref="Func{T, TResult}"/> function that returns an observable for each source item.</param>
+    /// <returns>An observable that emits values from all per-item observables, merged together.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="source"/> or <paramref name="observableSelector"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// <para>
+    /// Internally uses <see cref="SubscribeMany{T}(IObservable{IChangeSet{T}}, Func{T, IDisposable})"/> to manage per-item subscriptions.
```
JakenVeina: Remove internal implementation details.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -974,13 +1311,34 @@ public static IObservable<TDestination> MergeMany<T, TDestination>(this IObserva
     }
 
     /// <summary>
-    /// Operator similiar to Merge except it is ChangeSet aware.  All of the observable changesets are merged together into a single stream of ChangeSet events.
+    /// Merges multiple list changeset streams from an observable-of-observables into a single unified changeset stream.
+    /// Unlike cache MergeChangeSets, list merging performs no key-based deduplication.
```
JakenVeina: ~Swap "cache" for maybe "the cache equivalent". Reads better for people who aren't already familiar with the cache/list duality.~

Actually, how about just link to the operation you're referring to, with a `<see>`?

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1030,15 +1384,24 @@ public static IObservable<IChangeSet<TObject>> MergeChangeSets<TObject>(this IOb
     }
 
     /// <summary>
-    /// Operator similiar to Merge except it is ChangeSet aware.  All of the observable changesets are merged together into a single stream of ChangeSet events.
+    /// Merges a collection of list changeset streams into a single unified changeset stream.
+    /// This is the primary overload that all other list MergeChangeSets overloads delegate to.
```
JakenVeina: If this is the "primary" overload, you should probably de-duplicate the docs from that first one down to here. The first one, in particular, includes a change-type/behavior table.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1080,14 +1436,20 @@ public static IObservable<IChangeSet<TObject>> MergeChangeSets<TObject>(this IOb
...
     /// </summary>
-    /// <typeparam name="TObject">The type of the object.</typeparam>
+    /// <typeparam name="TObject">The type of items in the list.</typeparam>
     /// <typeparam name="TKey">The type of the object key.</typeparam>
-    /// <param name="source">The SourceList of Observable Cache ChangeSets.</param>
-    /// <param name="comparer"><see cref="IComparer{T}"/> instance to determine which element to emit if the same key is emitted from multiple child changesets.</param>
-    /// <returns>The result from merging the child changesets together.</returns>
-    /// <exception cref="ArgumentNullException">Parameter was null.</exception>
+    /// <param name="source">The <see cref="IObservableList{T}"/> of cache changeset observables.</param>
+    /// <param name="comparer"><see cref="IComparer{TObject}"/> to resolve which value wins when the same key appears in multiple sources.</param>
```
JakenVeina, line 1428: This seems like a prime example of where a change-type/behavior table is great, to elaborate on what "conflict resolution" looks like.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1295,24 +1738,41 @@ public static IObservable<IChangeSet<T>> OnItemRemoved<T>(
...
-    /// <param name="source">The source.</param>
-    /// <param name="others">The others.</param>
-    /// <returns>An observable which emits the change set.</returns>
+    /// <param name="source">The primary source <see cref="IObservable{IChangeSet{T}}"/> to union.</param>
+    /// <param name="others">The other <see cref="IObservable{IChangeSet{T}}"/> changeset streams to combine with.</param>
+    /// <returns>A list changeset stream containing items that exist in at least one source.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="others"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// <para>
+    /// Uses reference-counted equality comparison. An item is included when it first appears in any source and removed when it no longer exists in any source.
```
JakenVeina: Same as `.And()`, should reference that the default equality comparer is used.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1321,43 +1781,46 @@ public static IObservable<IChangeSet<T>> Or<T>(this IObservable<IChangeSet<T>> s
...
+    /// <param name="requests">An observable of <see cref="IPageRequest"/> controlling which page to display (page number and page size).</param>
+    /// <returns>An <see cref="IPageChangeSet{T}"/> stream containing only items within the current page window.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="source"/> or <paramref name="requests"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// <para>
+    /// Maintains the full source list internally and calculates the page window on each change or page request.
+    /// Items entering the page window produce <b>Add</b>; items leaving produce <b>Remove</b>. A new page request triggers
+    /// a full recalculation of the page contents.
+    /// </para>
+    /// <para><b>Worth noting:</b> The source should ideally be sorted before paging, as list order determines page contents. Duplicate items are removed from the result via <c>Distinct()</c>.</para>
```
JakenVeina: Are we saying that duplicate items are removed if the list is not sorted, because the operator has to fall back to equality comparison? I'm not sure how to interpret this, otherwise.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1475,12 +1961,15 @@ public static IObservable<IChangeSet<T>> Reverse<T>(this IObservable<IChangeSet<
     }
 
     /// <summary>
-    /// Defer the subscription until loaded and skip initial change set.
+    /// Skips the initial changeset (the snapshot emitted on subscription) and forwards all subsequent changesets.
```
JakenVeina, line 1956: I feel like this needs some remarks to clarify that it's REALLY dangerous. If the initial changeset isn't empty, then you have a corrupt collection that will throw if any of those initial items get refreshed, replaced, or removed.

Either that, or this description is just wrong, or misleading.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1490,19 +1979,37 @@ public static IObservable<IChangeSet<T>> SkipInitial<T>(this IObservable<IChange
     }
 
     /// <summary>
-    /// Sorts the sequence using the specified comparer.
+    /// Sorts the list using the specified comparer, maintaining a sorted output that incrementally updates as items change.
     /// </summary>
     /// <typeparam name="T">The type of the item.</typeparam>
```
JakenVeina, line 1985: More inconsistencies in type docs. Since this one's written as `T` instead of `TObject`. That in itself should probably change, but that'd be a non-docs change.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1490,19 +1979,37 @@ public static IObservable<IChangeSet<T>> SkipInitial<T>(this IObservable<IChange
...
+    /// removes are removed by index, and refreshes re-evaluate position (emitting <b>Moved</b> if changed).
+    /// </para>
+    /// <list type="table">
+    /// <listheader><term>Event</term><description>Behavior</description></listheader>
+    /// <item><term><b>Add</b>/<b>AddRange</b></term><description>Inserted at the correct sorted position. May trigger a full reset if the count exceeds <paramref name="resetThreshold"/>.</description></item>
+    /// <item><term><b>Replace</b></term><description>Old item removed, new item inserted at sorted position.</description></item>
+    /// <item><term><b>Remove</b>/<b>RemoveRange</b>/<b>Clear</b></term><description>Removed from sorted list.</description></item>
+    /// <item><term><b>Refresh</b></term><description>Sort position re-evaluated. If position changed, a <b>Moved</b> is emitted.</description></item>
+    /// <item><term>Comparer changed</term><description>Full re-sort of all items.</description></item>
+    /// <item><term>Resort signal</term><description>Full re-sort using the current comparer.</description></item>
```
JakenVeina: Should probably be "Re-sort".

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1512,19 +2019,15 @@ public static IObservable<IChangeSet<T>> Sort<T>(this IObservable<IChangeSet<T>>
         return new Sort<T>(source, comparer, options, resort, comparerChanged, resetThreshold).Run();
     }
 
+    /// <inheritdoc cref="Sort{T}(IObservable{IChangeSet{T}}, IComparer{T}, SortOptions, IObservable{Unit}?, IObservable{IComparer{T}}?, int)"/>
     /// <summary>
-    /// Sorts the sequence using the specified observable comparer.
+    /// Sorts the list using an observable comparer. The initial comparer is taken from the first emission; subsequent emissions trigger a full re-sort.
```
JakenVeina, line 2025: Should probably have a little extra clarity, in remarks, of what happens if no initial comparer is emitted. Do items just retain their existing upstream sorting? Does the downstream stay empty?

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1566,26 +2083,27 @@ public static IObservable<IChangeSet<T>> SubscribeMany<T>(this IObservable<IChan
...
     public static IObservable<IChangeSet<T>> SuppressRefresh<T>(this IObservable<IChangeSet<T>> source)
         where T : notnull => source.WhereReasonsAreNot(ListChangeReason.Refresh);
 
     /// <summary>
-    /// Transforms an observable sequence of observable lists into a single sequence
-    /// producing values only from the most recent observable sequence.
-    /// Each time a new inner observable sequence is received, unsubscribe from the
-    /// previous inner observable sequence and clear the existing result set.
+    /// Subscribes to the latest inner <see cref="IObservableList{T}"/>, switching to each new source and clearing the result when switching.
     /// </summary>
```
JakenVeina, line 2103: Maybe reference that this is the DynamicData equivalent of RX's native `.Switch()`, since you can't just use the native `.Switch()` on a changeset stream.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1866,20 +2413,13 @@ public static IObservable<IChangeSet<TDestination>> Transform<TSource, TDestinat
...
-    /// <param name="transformOnRefresh">Should a new transform be applied when a refresh event is received.</param>
-    /// <returns>A an observable change set of the transformed object.</returns>
-    /// <exception cref="ArgumentNullException">
-    /// source
-    /// or
-    /// valueSelector.
-    /// </exception>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TSource}}"/> to transform.</param>
+    /// <param name="transformFactory">A <see cref="Func{T, TResult}"/> function receiving the source item and its index, returning the transformed item.</param>
+    /// <param name="transformOnRefresh">When <see langword="true"/>, Refresh events re-invoke the factory.</param>
```
JakenVeina: No need to re-document.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1866,20 +2413,13 @@ public static IObservable<IChangeSet<TDestination>> Transform<TSource, TDestinat
...
-    /// <param name="source">The source.</param>
-    /// <param name="transformFactory">The transform function.</param>
-    /// <param name="transformOnRefresh">Should a new transform be applied when a refresh event is received.</param>
-    /// <returns>A an observable change set of the transformed object.</returns>
-    /// <exception cref="ArgumentNullException">
-    /// source
-    /// or
-    /// valueSelector.
-    /// </exception>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TSource}}"/> to transform.</param>
```
JakenVeina: No need to re-document.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1890,21 +2430,14 @@ public static IObservable<IChangeSet<TDestination>> Transform<TSource, TDestinat
...
```
JakenVeina: No need to re-document.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1890,21 +2430,14 @@ public static IObservable<IChangeSet<TDestination>> Transform<TSource, TDestinat
...
+    /// <param name="transformFactory">A function receiving the source item and the previous transformed value (as <see cref="Optional{T}"/>), returning the new transformed item.</param>
+    /// <param name="transformOnRefresh">When <see langword="true"/>, Refresh events re-invoke the factory.</param>
```
JakenVeina: No need to re-document.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1915,21 +2448,14 @@ public static IObservable<IChangeSet<TDestination>> Transform<TSource, TDestinat
...
-    /// <param name="source">The source.</param>
-    /// <param name="transformFactory">The transform factory.</param>
-    /// <param name="transformOnRefresh">Should a new transform be applied when a refresh event is received.</param>
-    /// <returns>A an observable change set of the transformed object.</returns>
-    /// <exception cref="ArgumentNullException">
-    /// source
-    /// or
-    /// valueSelector.
-    /// </exception>
+    /// <param name="source">The source <see cref="IObservable{IChangeSet{TSource}}"/> to transform.</param>
```
JakenVeina: No need to re-document.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -1915,21 +2448,14 @@ public static IObservable<IChangeSet<TDestination>> Transform<TSource, TDestinat
...
+    /// <param name="transformFactory">A <see cref="Func{T, TResult}"/> function receiving the source item, previous transformed value, and index.</param>
+    /// <param name="transformOnRefresh">When <see langword="true"/>, Refresh events re-invoke the factory.</param>
```
JakenVeina: No need to re-document.

src/DynamicData/List/ObservableListEx.cs
```diff
@@ -2250,13 +2809,33 @@ public static IObservable<IChangeSet<T>> WhereReasonsAreNot<T>(this IObservable<
...
-    /// <param name="source">The source.</param>
-    /// <param name="others">The others.</param>
-    /// <returns>An observable which emits the change set.</returns>
+    /// <param name="source">The primary source <see cref="IObservable{IChangeSet{T}}"/> to exclusively combine.</param>
+    /// <param name="others">The other <see cref="IObservable{IChangeSet{T}}"/> changeset streams to combine with.</param>
+    /// <returns>A list changeset stream containing items that exist in exactly one source.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="others"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// <para>
+    /// Uses reference-counted equality. An item is included when it exists in exactly one source.
```
JakenVeina: Default equality comparer is used.

JakenVeina (changes requested):

src/DynamicData/List/ObservableListEx.cs
dwcullop: Done.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Good call, swapped for a cref.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Done. Moved the canonical docs (table, remarks, exception) down to the IEnumerable overload and made the IObservable-source one inherit. Also tightened up the "primary" claim since multiple overloads call the operator directly.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Added a per-change table covering the conflict resolution semantics and source list add/remove.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Done. Added the explicit "default equality comparer for T" reference like And has.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Verified in Pager.cs: Distinct() is called unconditionally. Rewrote the remark so the sort guidance and the Distinct fact stand independently. They were reading as if linked.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Agreed. Added a prominent warning that any later refresh/replace/remove/move on a dropped initial item will throw because downstream has no record of it.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Agreed it's inconsistent. Leaving the param name change for a separate non-docs PR so this one stays scoped.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Done.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Documented: items sort by Comparer<T>.Default until the first comparer emission, then a full re-sort happens once it arrives.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Added a reference to Observable.Switch with a note that it cannot be applied directly to changeset streams.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Stripped all redundant param tags from the three variant overloads. Inheritdoc handles them; the differentiated transformFactory signature is visible in the method signature.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Same fix as above. Stripped.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Stripped.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Stripped.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Stripped.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Stripped.

src/DynamicData/List/ObservableListEx.cs
dwcullop: Done, parity with Or/And.

JakenVeina (approved):

PR #1088, opened 2026-05-17
Add Preview workflows and enforce versioning rules

dwcullop:
Adds pre-release publishing built on NBGV, plus three workflow_dispatch actions that handle every `version.json` edit so we don't have to touch it by hand.

## How versions work

| Branch | `version.json` | Publishes |
|---|---|---|
| `main` | `"X.Y-preview.{height}"` | `X.Y.0-preview.N` on every PR merge |
| `release/<major>.x` | `"X.Y"` + `"versionHeightOffset": -1` | `X.Y.N` stable on every PR merge (first published patch is `X.Y.0`, per semver convention; previews stay at `preview.1` per .NET ecosystem convention) |

NBGV's `{height}` is the first-parent commit count since the `version` field in `version.json` last changed. Two PRs merged the same day get distinct heights. No NuGet querying, no race conditions, no manual patch increments.

## Automation workflows

All run from the Actions tab.

| Workflow | When | What it does |
|---|---|---|
| Promote main to stable minor | Ready to ship the next minor stable | Opens two PRs: (1) merges main into `release/<major>.x` and sets the stable version (this one has the full diff, review it carefully); (2) bumps main to the next preview |
| Cut major release | Ready to ship a new major stable | Opens a main-bump PR, creates `release/<major>.x`, dispatches `release.yml` against the new branch |
| Bump main to next major preview | First breaking change is about to land | Opens a PR bumping main from `X.Y-preview` to `(X+1).0-preview`. Rejects skips |

Passive guards:

| Workflow | What it guards |
|---|---|
| `pr-version-check.yml` | Rejects any human-authored PR that modifies `version.json` unless labeled `manual-version-edit` (bot PRs from `bot/bump-*` and `bot/promote-*` are exempt). For PRs labeled `breaking-change`, also enforces that main's major is exactly `latest_stable_major + 1`. Any PR (labeled or not) that changes the major must carry the `breaking-change` label |
| Prerelease regression guard (in `release.yml`) | Refuses to publish stable from main, refuses to publish prerelease from `release/*.x`, refuses prerelease for `X.Y` if a stable `X.Y.*` tag already exists |

## Day-to-day

| Want to | Do this |
|---|---|
| Ship a stable patch | PR to `release/<major>.x`, merge |
| Cherry-pick a fix from main to a release branch | `git cherry-pick` onto a branch off `release/<major>.x`, PR, merge |
| Ship a preview | PR to `main`, merge |
| Ship the next stable minor | Run "Promote main to stable minor", merge both PRs (promotion first, then bump immediately) |
| Land a breaking change | Run "Bump main to next major preview", merge, then label and merge the breaking PR |
| Ship a new stable major | Run "Cut major release", merge the main-bump PR right after |

### 1. Cut `release/9.x` by hand (the only manual `version.json` edit ever)

`release/9.x` doesn't exist yet, and it needs to continue the existing `9.4.x` tag history (currently at `9.4.37`). The `pr-version-check` guard requires the `manual-version-edit` label on any PR that edits `version.json`, but the cutover is a direct push to a new branch (no PR), so the label isn't needed.

```sh
git fetch origin
git checkout -b release/9.x origin/main
# Edit version.json:
#   - Change "version" from "9.5-preview.{height}" back to "9.4"
#   - Set "versionHeightOffset" to <N> where N is the latest stable 9.4.x patch number
#     Verify: git tag --list '9.4.*' | sort -V | tail -1  (e.g. 37 -> set to 37; next publish = 9.4.38)
git add version.json
git commit -m "Cut release/9.x continuing 9.4.<N+1>"
git push -u origin release/9.x
```

The next build on `release/9.x` produces `9.4.<N+1>`. After this, every release goes through the automation. The `manual-version-edit` label exists for recovery cases (a failed automation run that needs hand-cleanup); regular work never touches `version.json`.

### 2. Make `PR version check` a required status check

In repo settings -> Branches -> branch protection rule for `main` (and `release/*.x`), add `PR version check / check` as a required status check. This converts the guard from advisory to binding.

## Process

The basic infra was straightforward. The automation went through three rounds of adversarial multi-agent review (six reviewers in round 1, then four in round 2, then two in round 3 to verify the round-2 fixes). The reviewers found ~50 unique issues across the rounds, most of which got fixed: PowerShell injection through inputs, the second-promotion always-conflicts merge, `cut-major` race conditions and ordering bugs, error handling that swallowed git failures silently, a few subtle off-by-ones in the version checks, supply-chain risk from floating action versions, and the prerelease-regresses-past-stable case. The commit history walks through all of it if you want details.

## Known limitation

Bot PRs are authored by `GITHUB_TOKEN`. GitHub deliberately suppresses workflows triggered by that token to prevent recursion, so `ci-build.yml` and `pr-version-check.yml` won't run on auto-generated PRs unless you close and reopen them (or wire up a PAT). The main-bump PRs are one-line `version.json` edits so this is usually fine. The promotion PRs from `promote-minor` carry the full diff of main and should be reviewed and have CI run before merging.

### DIFF
--- a/.github/workflows/cut-major.yml
+++ b/.github/workflows/cut-major.yml
@@ -57,6 +57,22 @@ jobs:
         if ($nextMain -notmatch '^(\d+)\.(\d+)$') {
           throw "next_main_version must be 'major.minor' (got: '$nextMain')"
         }
+        $nextMainMajor = [int]$matches[1]
+        $nextMainMinor = [int]$matches[2]
+        $cutMajor = [int]$major
+
+        if ($nextMainMajor -eq $cutMajor) {
+          if ($nextMainMinor -lt 1) {
+            throw "next_main_version ($nextMain) for the same major as the cut release must have minor >= 1. After cutting $major.0, the next main version should be $major.1 or later."
+          }
+        } elseif ($nextMainMajor -eq $cutMajor + 1) {
+          if ($nextMainMinor -ne 0) {
+            throw "next_main_version ($nextMain) for the next major must be $($cutMajor + 1).0. Use sequential major bumps (no skipping)."
+          }
+        } else {
+          throw "next_main_version ($nextMain) must be either '$major.<minor>' (continued minors of the cut major) or '$($cutMajor + 1).0' (next sequential major). To skip a major, edit version.json by hand."
+        }
+
         $nextPreview = "$nextMain-preview.{height}"
 
         Add-Content -Path $env:GITHUB_ENV -Value "RELEASE_BRANCH=$releaseBranch"
@@ -75,8 +91,10 @@ jobs:
       run: |
         $path = 'version.json'
         $content = [System.IO.File]::ReadAllText($path)
-        $new = [regex]::Replace($content, '"version":\s*"[^"]+"', "`"version`": `"$env:STABLE_VERSION`"")
-        [System.IO.File]::WriteAllText($path, $new)
+        $content = [regex]::Replace($content, '"version":\s*"[^"]+"', "`"version`": `"$env:STABLE_VERSION`"")
+        $content = [regex]::Replace($content, '(?m)^\s*"versionHeightOffset"\s*:\s*-?\d+\s*,?\s*\r?\n', '')
+        $content = [regex]::Replace($content, ',(\s*[}\]])', '$1')
+        [System.IO.File]::WriteAllText($path, $content)
 
     - name: Commit and push release branch
       run: |
--- a/.github/workflows/cut-major.yml
+++ b/.github/workflows/cut-major.yml
@@ -1,0 +1,134 @@
+name: Cut major release
+
+on:
+  workflow_dispatch:
+    inputs:
+      major_version:
+        description: 'Major version to ship as stable (e.g., 10). Main must currently be on this major''s preview line.'
+        required: true
+        type: string
+      next_main_version:
+        description: 'Next preview version for main as major.minor (default: <major>.1). Use the next major if more breaking changes are queued.'
+        required: false
+        type: string
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  cut:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v6
+      with:
+        fetch-depth: 0
+
+    - name: Configure git
+      run: |
+        git config user.email "github-actions[bot]@users.noreply.github.com"
+        git config user.name "github-actions[bot]"
+
+    - name: Validate inputs
+      shell: pwsh
+      run: |
+        $major = '${{ inputs.major_version }}'
+        if ($major -notmatch '^\d+$') {
+          throw "major_version must be a single integer (got: '$major')"
+        }
+
+        $releaseBranch = "release/$major.x"
+        $stableVersion = "$major.0"
+
+        $mainVersionJson = git show "origin/main:version.json" | ConvertFrom-Json
+        $mainVer = $mainVersionJson.version
+        if ($mainVer -notmatch "^$major\.0-preview") {
+          throw "main's version is '$mainVer' but should be '$major.0-preview.{height}' to cut $major.0 stable. Run 'Bump main to next major preview' first if needed."
+        }
+
+        $existing = git ls-remote --heads origin "refs/heads/$releaseBranch"
+        if ($existing) {
+          throw "Branch '$releaseBranch' already exists. Cannot cut a new major release branch that exists."
+        }
+
+        $nextMain = '${{ inputs.next_main_version }}'
+        if (-not $nextMain) { $nextMain = "$major.1" }
+        if ($nextMain -notmatch '^(\d+)\.(\d+)$') {
+          throw "next_main_version must be 'major.minor' (got: '$nextMain')"
+        }
+        $nextPreview = "$nextMain-preview.{height}"
+
+        Add-Content -Path $env:GITHUB_ENV -Value "RELEASE_BRANCH=$releaseBranch"
+        Add-Content -Path $env:GITHUB_ENV -Value "STABLE_VERSION=$stableVersion"
+        Add-Content -Path $env:GITHUB_ENV -Value "NEXT_PREVIEW=$nextPreview"
+        Add-Content -Path $env:GITHUB_ENV -Value "NEXT_MAIN_VERSION=$nextMain"
+        Add-Content -Path $env:GITHUB_ENV -Value "BUMP_BRANCH=bot/bump-main-after-$stableVersion"
+
+    - name: Create release branch from main
+      run: |
+        git checkout main
+        git checkout -b "$RELEASE_BRANCH"
+
+    - name: Set stable version on release branch
+      shell: pwsh
+      run: |
+        $path = 'version.json'
+        $content = [System.IO.File]::ReadAllText($path)
+        $new = [regex]::Replace($content, '"version":\s*"[^"]+"', "`"version`": `"$env:STABLE_VERSION`"")
+        [System.IO.File]::WriteAllText($path, $new)
+
+    - name: Commit and push release branch
+      run: |
+        git add version.json
+        git commit -m "Cut $RELEASE_BRANCH at $STABLE_VERSION stable"
+        git push -u origin "$RELEASE_BRANCH"
+
+    - name: Create main bump branch
+      run: |
+        git checkout main
+        git checkout -b "$BUMP_BRANCH"
+
+    - name: Bump main to next preview
+      shell: pwsh
+      run: |
+        $path = 'version.json'
+        $content = [System.IO.File]::ReadAllText($path)
+        $new = [regex]::Replace($content, '"version":\s*"[^"]+"', "`"version`": `"$env:NEXT_PREVIEW`"")
+        [System.IO.File]::WriteAllText($path, $new)
+
+    - name: Commit and push main bump branch
+      run: |
+        git add version.json
+        git commit -m "Bump main to $NEXT_PREVIEW after cutting $RELEASE_BRANCH"
+        git push -u origin "$BUMP_BRANCH"
+
+    - name: Open main bump PR
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      shell: pwsh
+      run: |
+        $body = @"
+        Companion PR to cutting ``$env:RELEASE_BRANCH``.
+
+        The release branch ``$env:RELEASE_BRANCH`` has been pushed and will publish ``$env:STABLE_VERSION.0`` stable.
+
+        This PR advances ``main`` to ``$env:NEXT_PREVIEW`` so subsequent PRs publish that preview line.
+        "@
+        $url = gh pr create `
+          --base main `
+          --head "$env:BUMP_BRANCH" `
+          --title "Bump main to $env:NEXT_PREVIEW after $env:RELEASE_BRANCH cut" `
+          --body $body
+        Add-Content -Path $env:GITHUB_ENV -Value "BUMP_PR_URL=$url"
+
+    - name: Summary
+      shell: pwsh
+      run: |
+        $summary = @"
+        ## Major release cut
+
+        - **New release branch**: ``$env:RELEASE_BRANCH`` (will publish ``$env:STABLE_VERSION.0`` stable on push).
+        - **Main bump PR** (merge to advance main to next preview): $env:BUMP_PR_URL
+        "@
+        Add-Content -Path $env:GITHUB_STEP_SUMMARY -Value $summary
--- a/.github/workflows/cut-major.yml
+++ b/.github/workflows/cut-major.yml
@@ -18,7 +18,7 @@ permissions:
   actions: write
 
 concurrency:
-  group: cut-major-${{ inputs.major_version }}
+  group: main-version-mutator
   cancel-in-progress: false
 
 jobs:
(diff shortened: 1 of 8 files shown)


.github/workflows/promote-minor.yml
```diff
@@ -0,0 +1,171 @@
...
+        git checkout -b "$PROMOTE_BRANCH"
+        git merge origin/main --no-edit
+
+    - name: Set stable version on promotion branch
+      shell: pwsh
+      run: |
+        $path = 'version.json'
+        $content = [System.IO.File]::ReadAllText($path)
+        $new = [regex]::Replace($content, '"version":\s*"[^"]+"', "`"version`": `"$env:STABLE_VERSION`"")
+        [System.IO.File]::WriteAllText($path, $new)
```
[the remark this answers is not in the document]
dwcullop: Fixed in 124e51d2. The promote-minor.yml `Set stable version` step now also regex-strips any `versionHeightOffset` line in version.json (with a follow-up trailing-comma cleanup so the result remains valid JSON). Verified across three cases: offset in middle, offset at end, offset absent. Same fix applied defensively in cut-major.yml.

.github/workflows/promote-minor.yml
```diff
@@ -0,0 +1,171 @@
...
+          throw "stable_version ($version) must be greater than current version on '$branch' ($branchVer)"
+        }
+
+        $mainVersionJson = git show "origin/main:version.json" | ConvertFrom-Json
+        $mainVer = $mainVersionJson.version
+        if ($mainVer -notmatch '^(\d+)\.(\d+)-preview') {
+          throw "main's version.json must be in '<major>.<minor>-preview.{height}' form (got: '$mainVer')"
+        }
+        $mainMajor = [int]$matches[1]
+        $mainMinor = [int]$matches[2]
```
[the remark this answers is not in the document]
dwcullop: Fixed in 124e51d2. The `Validate inputs` step in promote-minor.yml now requires `stable_version` to exactly match main's current preview major.minor. Mismatched inputs throw with an error pointing to bump-major-preview as the right workflow if main needs advancing first.

version.json
```diff
@@ -1,9 +1,8 @@
 {
-    "version": "9.4",
+    "version": "9.5-preview.{height}",
     "publicReleaseRefSpec": [
-        "^refs/heads/main$", // we release out of master
-        "^refs/heads/preview/.*", // we release previews
-        "^refs/heads/rel/\\d+\\.\\d+\\.\\d+" // we also release branches starting with rel/N.N.N
+        "^refs/heads/main$", // main publishes pre-release packages
+        "^refs/heads/release/.+" // release/<major>.x branches publish stable packages
```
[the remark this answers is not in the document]
dwcullop: Fixed in 124e51d2. Tightened `publicReleaseRefSpec` from `^refs/heads/release/.+` to `^refs/heads/release/\d+\.x$` so only branches like release/9.x, release/10.x match. Branches like release/test or release/9.x-experiment now produce non-public builds (with the -g<commit> NBGV suffix) and are not treated as stable releases.

.github/workflows/cut-major.yml
```diff
@@ -0,0 +1,134 @@
...
+        if ($existing) {
+          throw "Branch '$releaseBranch' already exists. Cannot cut a new major release branch that exists."
+        }
+
+        $nextMain = '${{ inputs.next_main_version }}'
+        if (-not $nextMain) { $nextMain = "$major.1" }
+        if ($nextMain -notmatch '^(\d+)\.(\d+)$') {
+          throw "next_main_version must be 'major.minor' (got: '$nextMain')"
+        }
+        $nextPreview = "$nextMain-preview.{height}"
```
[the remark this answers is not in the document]
dwcullop: Fixed in 124e51d2. The cut-major.yml `Validate inputs` step now requires `next_main_version` to be either `<cutMajor>.<minor>` with minor >= 1 (continued minors of the cut major) or `(cutMajor+1).0` (next sequential major). Inputs like `next_main_version=12.0` after cutting 10.0 are rejected with an error message that points to the manual escape hatch (edit version.json by hand).

.github/workflows/release.yml
```diff
@@ -2,17 +2,19 @@ name: Build and Release
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, release/** ]
```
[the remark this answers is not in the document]
dwcullop: Fixed in 124e51d2. Narrowed the push trigger from `release/**` to `release/*.x` (the best the GH Actions glob can express) and added a `Validate publish branch` step early in the job that fails fast if `github.ref_name` is not 'main' or doesn't match the regex `^release/\d+\.x$`. Branches like release/test won't even start the job; release/foo.x would start it and fail before NBGV runs. Same trigger narrowing applied to ci-build.yml and pr-version-check.yml for consistency.

RELEASING.md
```diff
@@ -0,0 +1,106 @@
...
+| `release/<major>.x` | Lifetime branch for an entire major version (e.g. `release/9.x` hosts all 9.x minors). | Stable: `X.Y.N` |
+
+Every PR merged to either branch publishes a NuGet package automatically via `.github/workflows/release.yml`.
+
+## How patch numbers are computed
+
+NBGV walks the first-parent history from `HEAD` back to the commit where `version.json`'s `version` field last changed. The count of commits is the "height". `{height}` in `version.json` is replaced with that count.
+
+- Two PRs merging the same day get distinct heights, no collisions.
+- A `version.json` change resets height to 1 (so the commit that sets the version is published as `X.Y.1`, not `X.Y.0`).
```
JakenVeina: A major/minor version bump should definitely yield a published version of X.Y.0, not X.Y.1. Is this a limitation of NBGV? Do we need to swap a `height` for `height - 1` somewhere?

RELEASING.md
```diff
@@ -0,0 +1,106 @@
...
+- `fetch-depth: 0` in CI is required for height to be correct.
+
+## Automation workflows
+
+All `version.json` edits and release-branch creation are scripted via `workflow_dispatch` actions in the GitHub Actions UI. Maintainers should not edit `version.json` by hand.
```
JakenVeina, line 24: We should setup a check step in the PR build workflow, to reject PRs that change this. Maybe with an exception for PRs by a maintainer.

RELEASING.md
```diff
@@ -0,0 +1,106 @@
...
+
+| Workflow | When to run | What it does |
+|---|---|---|
+| **Promote main to stable minor** (`promote-minor.yml`) | Ready to ship the next minor (first published patch is `X.Y.1`) from `main` to an existing release branch. | Opens two PRs: (1) merges `main` into `release/<major>.x` and sets stable version (contains the full diff, NOT mechanical); (2) bumps `main` to the next preview minor (mechanical). |
```
JakenVeina: This doesn't accommodate cherrypicking from main to a release branch. That's definitely something we need to be able to do.

RELEASING.md
```diff
@@ -0,0 +1,106 @@
...
+Two passive guards run on every PR / release:
+
+| Workflow | Trigger | What it does |
+|---|---|---|
+| **PR version check** (`pr-version-check.yml`) | Pull request to `main` / `release/*.x`. | For PRs to `main` labeled `breaking-change`, fails the check unless the PR's `version.json` major is **exactly one greater** than the latest stable tag's major (no skipping). For PRs to `release/*.x`, the check is skipped (breaking changes are a main-only concern). |
+| **Prerelease regression guard** (in `release.yml`) | Every push to `main` and dispatched run on `release/*.x`. | Refuses to publish a prerelease from a `release/*.x` branch (those must be stable only). Refuses to publish a prerelease for `X.Y` if a stable `X.Y.*` tag already exists. |
+
+### Recommended branch protection
+
+To make the **PR version check** binding (not advisory), add `PR version check / check` as a required status check on `main` in branch protection settings.
```
JakenVeina: This doesn't belong here.

RELEASING.md
```diff
@@ -0,0 +1,106 @@
...
+2. The workflow opens a main-bump PR, creates `release/10.x`, and dispatches `release.yml` against the new branch to publish the first `10.0.x` patch. The dispatched run URL is in the workflow summary.
+3. Merge the main-bump PR. Don't wait: every push to `main` after the new stable ships will fail the prerelease-regression guard until this PR merges.
+
+### Manual escape hatch
+The automation workflows are thin wrappers around `version.json` edits. If something goes wrong, you can always perform the equivalent edits by hand. See the workflow YAML files for the exact operations. Recovery scenarios:
+
+- **`cut-major` failed after pushing the release branch but before dispatching `release.yml`**: manually run `release.yml` against the new `release/<major>.x` branch from the Actions tab.
+- **`promote-minor` failed mid-flight**: the bot branches `bot/promote-<version>-<run_id>` and `bot/bump-main-after-<version>-<run_id>` carry the run ID, so a retry produces fresh branches. Delete any stale bot branches or PRs from the failed run before re-running.
+
+## Initial cutover (one-time, when introducing this infrastructure)
```
JakenVeina: This whole section doesn't belong here.

.github/workflows/release.yml
```diff
@@ -2,17 +2,23 @@ name: Build and Release
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, 'release/*.x' ]
+  workflow_dispatch:
 
 env:
   configuration: Release
-  productNamespacePrefix: "ReactiveMarbles"
```
JakenVeina, line 9: Why was this removed?

.github/workflows/release.yml
```diff
@@ -41,35 +58,83 @@ jobs:
...
+        $packages = Get-ChildItem -Path src -Recurse -Filter '*.nupkg' | Where-Object { $_.FullName -like '*\bin\Release\*' }
+        if (-not $packages) { throw "No .nupkg files found under src/**/bin/Release/." }
+        foreach ($pkg in $packages) {
+          dotnet nuget push -s $env:SOURCE_URL -k $env:NUGET_AUTH_TOKEN --skip-duplicate $pkg.FullName
+          if ($LASTEXITCODE -ne 0) { throw "dotnet nuget push failed for $($pkg.Name) (exit $LASTEXITCODE)." }
+        }
+
     - name: Changelog
-      uses: glennawatson/ChangeLog@v1
+      uses: glennawatson/ChangeLog@0464dd89b26f61fecf24b41d675f8ffdb11c4c3f # v1
```
JakenVeina, line 130: This looks very odd. What's going on here?

.github/workflows/cut-major.yml
```diff
@@ -0,0 +1,270 @@
...
+      shell: pwsh
+      run: |
+        $ErrorActionPreference = 'Stop'
+
+        $major = $env:MAJOR_VERSION_INPUT
+        if ($major -notmatch '^(0|[1-9]\d*)$') {
+          throw "major_version must be a positive integer with no leading zeros (got: '$major')"
+        }
+        $cutMajor = [int]$major
+        $major = [string]$cutMajor
```
JakenVeina: If you're already validating no leading 0's or spaces on the input, what does int-cycling accomplish?

JakenVeina (changes requested):

RELEASING.md
dwcullop: Good catch. Switched the release branches to write `versionHeightOffset: -1` so the commit that sets the new stable version publishes as X.Y.0 instead of X.Y.1. Both cut-major.yml and promote-minor.yml do this now. Previews stay on the preview.1 convention (main has no offset) since .NET ecosystem convention starts previews at .1 rather than .0 ([see .NET 10 Preview 1](https://devblogs.microsoft.com/dotnet/dotnet-10-preview-1/) and the [NuGet versioning examples](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#pre-release-versions)).

RELEASING.md
dwcullop: Done. Extended pr-version-check.yml to reject any human-authored PR that touches version.json unless explicitly labeled 'manual-version-edit'. Bot PRs (bot/bump-* and bot/promote-* opened by github-actions[bot]) are exempted automatically. Maintainers who genuinely need a hand-edit (cutover, recovery) add the label.

RELEASING.md
dwcullop: Already supported, just needed to be called out. release.yml triggers on every push to release/*.x, so you can cherry-pick and PR to the release branch directly. Added a 'Cherry-picking a fix from main to a release branch' section under Day-to-day flows with the explicit git commands.

RELEASING.md
dwcullop: Removed. Branch-protection is a one-time admin task and doesn't belong in operational docs.

RELEASING.md
dwcullop: Removed. The PR body already has these steps as the short version; no need to keep them in RELEASING.md after cutover.

.github/workflows/release.yml
dwcullop: Dead code. Nothing in the repo referenced productNamespacePrefix. The multi-agent review flagged it and I dropped it in 6bc57b20.

.github/workflows/release.yml
dwcullop: SHA pin for supply-chain security. GitHub's official guidance is to pin third-party actions to a full-length commit SHA because tags are mutable and a compromised maintainer could re-point `v1` to a malicious commit. The `# v1` trailing comment is the standard human-readable annotation Dependabot uses when it rewrites tag refs to SHAs. Refs: [Secure use reference: Using third-party actions](https://docs.github.com/en/actions/reference/secure-use-reference#using-third-party-actions) and [Using SHAs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/find-and-customize-actions#using-shas).

.github/workflows/cut-major.yml
dwcullop: Right, the `[string]` round-trip was a no-op since the regex already validates no leading zeros / no whitespace. Dropped it here and in bump-major-preview.yml which had the same pattern.

JakenVeina (approved):

PR #1079, opened 2026-04-12
Fix cross-cache deadlocks via queue-drain delivery pattern

dwcullop:
**This PR replaces #1074 and provides a more robust fix**

## The Problem

Many DynamicData cache operators use `Synchronize(lock)` to serialize concurrent access. This acquires a lock and **holds it during the entire downstream `observer.OnNext()` delivery**. When a pipeline crosses cache boundaries (one cache's output feeds into another cache), the downstream cache's lock can be acquired while the upstream operator's lock is still held. If two threads follow reverse paths through the same pair of caches, this creates a classic ABBA deadlock.

This is not a theoretical issue — any application that composes complex DynamicData pipelines across multiple `SourceCache` instances with concurrent writers can deadlock.

## The Solution: Queue-Drain Delivery

Replace `Synchronize(lock)` with a queue-drain pattern that **releases the lock before delivering to downstream subscribers**. This eliminates ABBA deadlocks because no operator lock is held during cross-cache delivery.

### Two Delivery Queue Classes

**`DeliveryQueue<T>`** — typed, single-source queue. Implements `IObserver<T>`.
- Internally stores `Queue<Notification<T>>` (value/error/completed)
- On enqueue: acquires the gate, queues the notification, releases the gate, THEN delivers outside the lock
- Single-deliverer token (`_isDelivering`) ensures Rx serialization contract
- Used by `MergeMany` (N children of the same type), `DisposeMany`/`OnBeingRemoved` (need `EnsureDeliveryComplete` for disposal), and `ObservableCache` itself (the core fix)
- `EnsureDeliveryComplete()` with `_drainThreadId` re-entrant safety for disposal callbacks

**`SharedDeliveryQueue`** — type-erased, multi-source queue.
- One queue shared across multiple sources of different types (e.g., Sort subscribes to both data source and comparer changes)
- Each source gets a typed `DeliverySubQueue<T>` via `CreateQueue<T>(observer)`
- Single drain loop delivers from all sub-queues outside the lock
- Used by 24+ operators that subscribe to multiple observables

### Drop-In Replacement: `SynchronizeSafe`

Three extension method overloads in `SynchronizeSafeExtensions` provide surgical drop-in replacements for `Synchronize(lock)`:

```csharp
// Multi-source (most operators): replace Synchronize(locker) 
source.SynchronizeSafe(queue)     // SharedDeliveryQueue

// Single-source: replace Synchronize(locker) with implicit queue creation
[...]

PR #1097, opened 2026-05-27
Eliminate Rx Merge gate in queue-serialized operators

dwcullop:
## Problem

PR #1079 moved cross-cache operators from `Synchronize(lock)` to `SynchronizeSafe`, which routes every notification through a `SharedDeliveryQueue` that releases the lock before invoking the downstream observer. The goal was that no operator-level lock would be held across a cross-cache call, so two operators on a bidirectional pipeline could not form an ABBA cycle.

Several operators completed the queue routing but then combined their already-serialized inputs with `Observable.Merge` (or `Observable.CombineLatest`, or `Observable.Switch`) before delivery. Rx's multi-input combinators install a private `_gate` and hold it for the full duration of every downstream `OnNext`. When that downstream delivery walks into another cache's writer lock, two such gates on two operators reconstruct the ABBA cycle that the queue-drain design was supposed to eliminate.

`DeadlockTortureTest.Page_DoesNotDeadlock` (added in #1079) caught this for `Page` as an intermittent CI failure. The same latent bug existed in every other site that combined queue-serialized inputs through a gate-holding Rx combinator.

## Fix

Introduce three internal Rx helpers and apply them at every queue-serialized site:

**`UnsynchronizedMerge<T>(this IObservable<T>, params IObservable<T>[])`** in `SynchronizeSafeExtensions.cs`. Drop-in alternative to `Observable.Merge` that performs no synchronization of its own. Preserves Merge's terminal semantics (completes only after every source completes; first error terminates; subscription happens in argument order) but does not install a gate. Each source is subscribed through its own `Observer.Create` instance because Rx's `ObserverBase` sets a one-shot stopped flag on the first `OnCompleted`/`OnError`; a single shared observer would silently drop terminal notifications from every source after the first. The `OnCompleted` handler decrements a shared counter; only the last surviving source's completion fires downstream.

Safe to use only when every input is already serialized; in this library that precondition is satisfied by routing each input through the same `SharedDeliveryQueue` via `SynchronizeSafe(queue)` before the merge.

**`DeliveryQueueMerge<T>(this IObservable<T>, params IObservable<T>[])`** in `DeliveryQueueMergeExtensions.cs`. Same-type Rx merge that owns its own `DeliveryQueue<T>` so the call site reads like an ordinary `Observable.Merge` and never has to mention queue plumbing. Used where every input has the same element type and no per-input projection is needed inside the drain (`AutoRefresh`).

**`UnsynchronizedCombineLatest<TFirst, TSecond, TResult>`** in `SynchronizeSafeExtensions.cs`. Two-input drop-in for `Observable.CombineLatest` with the same precondition as `UnsynchronizedMerge`: both inputs must be pre-serialized through the same gate. Uses `Optional<T>` for the latest-value state so there is no `default!` / null-forgiving fallback and the "has produced a value yet" question is answered by the same field that holds the value.

### Sites addressed

| Operator | Shape used |
|---|---|
| `AutoRefresh` | `shared.DeliveryQueueMerge(refreshChanges)` |
| `Page` | `SharedDeliveryQueue` + `SynchronizeSafe(queue).Select(projection)` per input + `UnsynchronizedMerge` |
| `Virtualise` | same shape as `Page` |
| `Sort` (conditional branch) | three-input variant of the same shape |
| `SortAndPage` | `UnsynchronizedMerge` |
| `SortAndVirtualize` | `UnsynchronizedMerge` (collapsing chained `.Merge().Merge()` into one call) |
| `GroupOnImmutable` | same shape as `Page` |
| `QueryWhenChanged` (itemChangedTrigger branch) | same shape as `Page` |
| `TransformWithForcedTransform` | `UnsynchronizedMerge` directly (queue shared with a `Publish`/cacheLoader subscription that lives outside the merge) |
| `GroupOn` | `groups.UnsynchronizedMerge(regroup)` (both queue-serialized) |
| `GroupOnDynamic` | three-source `UnsynchronizedMerge` for downstream completion signal |
| `TransformAsync` (forceTransform branch) | `transformer.UnsynchronizedMerge(forced)` |
| `TransformMany` (childChanges path) | `initial.UnsynchronizedMerge(subsequent)` |
| `TreeBuilder` | `predicateChanged.UnsynchronizedCombineLatest(reFilterObservable)` (reFilterObservable now also routed through the queue) |
| `Switch` | inlined switch logic with `SerialDisposable` + `UnsynchronizedMerge` of `destination.Connect()` and the errors `Subject` |

`Sort`'s three-source case becomes a single `UnsynchronizedMerge(a, b, c)` call instead of nested `.Merge().Merge()`, which also removes one of the two gates the chained form would have created.

`Switch` is restructured rather than helper-substituted: `Observable.Switch` is replaced by an inline `SerialDisposable` that swaps the inner subscription on each outer notification, both outer and inner are routed through the same queue, and the final destination/errors merge uses `UnsynchronizedMerge` (both inputs originate inside the queue drain).

### Coverage matrix for Rx gate-holding combinators

The audit of every cache-side usage now lives in [`rx-gate-holding-operators.instructions.md`](https://github.com/reactivemarbles/DynamicData/blob/fix/operator-merge-gate-deadlock/.github/instructions/rx-gate-holding-operators.instructions.md), alongside the rule it supports and the helpers that implement it.

## Test coverage

`DeadlockTortureTest` is expanded so the same fixture catches a future regression in any of the addressed operators:

- New `[Fact] GroupWithImmutableState_DoesNotDeadlock`.
- New `[Fact] QueryWhenChanged_DoesNotDeadlock` - uses a side-channel `Subscribe(_ => otherCache.AddOrUpdate(...))` to close the ABBA cycle, since `QueryWhenChanged` does not produce a changeset that `PopulateInto` can consume.
- New `[Fact] SortAndPage_DoesNotDeadlock`.
- New `[Fact] SortAndVirtualize_DoesNotDeadlock`.
- New `[Fact] GroupOnWithRegrouper_DoesNotDeadlock` - drives the `groups.Merge(regroup)` branch of `GroupOn` via a subject pusher pumping the regrouper.
- New `[Fact] GroupOnDynamicSelector_DoesNotDeadlock` - drives the three-source completion-signal merge in `GroupOnDynamic` via a `BehaviorSubject` selector and a regrouper subject.
- New `[Fact] TransformAsyncWithForce_DoesNotDeadlock` - drives the `transformer.Merge(forced)` branch of `TransformAsync` via a force subject.
- New `[Fact] TransformToTree_DoesNotDeadlock` - drives the `predicateChanged.CombineLatest(reFilterObservable)` site in `TreeBuilder`.
- New `[Fact] Switch_DoesNotDeadlock` - drives the refactored switch logic via `Observable.Return(s).Switch()`.
- `AllDangerous_Stacked_DoNotDeadlock` now stacks `GroupWithImmutableState` and `Virtualise` into the kitchen-sink pipeline.
- `MultiplePairs_Simultaneous_NoDeadlock` gains `GroupWithImmutableState`, `SortAndPage`, and `SortAndVirtualize` lanes.

Tests that pass a subject input (`Page`, `Virtualise`, `BatchIf`, `TransformWithForce`, `AllDangerous_Stacked`, `MultiplePairs`) previously created the subject and let it sit idle. `BehaviorSubject` initial values reached the operator, but no test ever pushed during the race, so the operator's subject-driven branch was never exercised. The deadlock still formed via source-side flow alone, which is why `Page`/`Virtualise` did fail on main, but the test could miss a regression that only manifested on the subject-driven path. `RunBidirectionalDeadlockTest` now takes an optional `Action? subjectPusher` that runs on a third worker gated by the same `Barrier`, and each subject-bearing test pushes its own pattern on the subject while the writer threads are pushing on the sources.

The per-iteration `TimeoutSeconds` is raised from 15 to 60. The CI runner is intentionally stripped down; the test budget should accommodate it. A real deadlock hangs forever, not 60s, so the timeout still distinguishes deadlock from slow.

New focused fixtures for the merge / combine helpers themselves:

- `UnsynchronizedMergeFixture` (9 tests) - arrival-order forwarding, all-must-complete OnCompleted, first-error-wins OnError, late-OnError-after-first-error suppression, late-OnCompleted-after-OnError suppression, argument-order subscription, synchronous-`Empty<T>`/`Throw<T>` at subscribe time, no-others fallback.
- `DeliveryQueueMergeFixture` (matching contract plus a concurrent serialization check: two producers race 1,000 items each and the test asserts `maxInFlight == 1`).
- `UnsynchronizedCombineLatestFixture` (11 tests) - emit-only-after-both-have-produced, every-subsequent-OnNext, all-must-complete OnCompleted, first-error-wins, OnNext suppression after error, late-OnCompleted-after-OnError suppression, latest-value semantics, synchronous initial values via `BehaviorSubject`, synchronous-`Empty`/`Throw` boundary cases.

## Verification

- `DeadlockTortureTest` fixture: 21/21 pass, full suite in 15s with `xunit.parallelizeTestCollections=false`.
- Targeted unit tests for every touched operator (`Sort` + `Virtualise` + `Page` + `AutoRefresh` + `Group*` + `QueryWhenChanged` + `Transform*` + `Tree*` + `Switch` + `EditDiff*` + the three internal merge fixtures): all pass.
- Cache + Internal full sweep: 1486 passed, 0 failed, 9 skipped.
- Last CI build green.

## Additional Fix: `EditDiffChangeSetOptional`

`EditDiffChangeSetOptional` wrapped its consumer-provided source in `_source.Synchronize()` to defensively serialize `OnNext` deliveries against its own `previous` closure state. Per the Rx contract (rules `§6.8` and `§5.8` in [`rx-design-guide.instructions.md`](https://github.com/reactivemarbles/DynamicData/blob/fix/operator-merge-gate-deadlock/.github/instructions/rx-design-guide.instructions.md)) a single-source operator must not redundantly serialize; the source already guarantees serialization by `§4.2`. The sister class `EditDiffChangeSet` already follows this rule (no `Synchronize`). The defensive call is removed and the file is reindented from its previous deep-nesting layout.


src/DynamicData/Internal/DeliveryQueueMergeExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// source through a single <see cref="DeliveryQueue{T}"/>. Functionally equivalent
+    /// to <see cref="Observable.Merge{TSource}(IObservable{TSource}[])"/>: completes
+    /// only after every source completes; the first error terminates; subscription
+    /// occurs in argument order.
+    /// </summary>
+    /// <typeparam name="T">The element type, common to every input.</typeparam>
+    /// <param name="first">The first input observable.</param>
+    /// <param name="others">Additional input observables.</param>
+    /// <returns>An observable that emits items from every input, serialized through the queue.</returns>
+    public static IObservable<T> DeliveryQueueMerge<T>(this IObservable<T> first, params IObservable<T>[] others) =>
```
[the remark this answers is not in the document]
dwcullop: Added `src/DynamicData.Tests/Internal/DeliveryQueueMergeFixture.cs` (commit 88c9f2fc). Same behavioural contract as the `UnsynchronizedMerge` fixture plus a concurrent serialization check: two producers race 1,000 items each through the merged stream while the observer increments/decrements an in-flight counter and tracks the max. The test asserts `maxInFlight == 1` (the queue never lets two OnNexts overlap downstream) and that the full 2,000-item bag is delivered exactly once.

src/DynamicData/Internal/SynchronizeSafeExtensions.cs
```diff
@@ -78,4 +79,74 @@ public static IObservable<T> SynchronizeSafe<T>(this IObservable<T> source) =>
...
+    /// <para>Each source is subscribed through its own <see cref="Observer.Create{T}(Action{T}, Action{Exception}, Action)"/>
+    /// instance. The actions close over shared <c>pending</c> and <c>terminated</c> counters, but
+    /// the observer instances must be distinct because Rx's <c>ObserverBase</c> sets a one-shot
+    /// stopped flag on the first <c>OnCompleted</c>/<c>OnError</c>; a single shared observer
+    /// would silently drop terminal notifications from every source after the first.</para>
+    /// <para>Without the external serialization precondition, concurrent <c>OnNext</c>
+    /// calls into the shared observer will race. Do not use as a general-purpose
+    /// <c>Observable.Merge</c> replacement.</para>
+    /// </remarks>
+    public static IObservable<T> UnsynchronizedMerge<T>(this IObservable<T> first, params IObservable<T>[] others) =>
```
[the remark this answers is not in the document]
dwcullop: Added `src/DynamicData.Tests/Internal/UnsynchronizedMergeFixture.cs` (commit 88c9f2fc). Covers: arrival-order forwarding across sources, all-must-complete OnCompleted (3 sources), first-error-wins OnError, late-OnError-after-first-error suppression, late-OnCompleted-after-OnError suppression, argument-order subscription, synchronous `Empty<T>` at subscribe still counts toward the completion gate, synchronous `Throw<T>` at subscribe propagates immediately, and the no-others fallback (params array empty).

src/DynamicData.Tests/Cache/DeadlockTortureTest.cs
```diff
@@ -64,26 +67,68 @@ [Fact] public async Task AutoRefresh_DoesNotDeadlock() =>
...
+            subjectPusher: () => { for (var j = 0; j < ItemCount; j++) req.OnNext(new PageRequest(1 + (j % 4), 25 + (j % 4) * 25)); })).Should().BeTrue();
     }
 
     [Fact] public async Task Virtualise_DoesNotDeadlock()
     {
         using var req = new BehaviorSubject<IVirtualRequest>(new VirtualRequest(0, 50));
-        (await RunBidirectionalDeadlockTest(s => s.Sort(SortExpressionComparer<Person>.Ascending(p => p.Age)).Virtualise(req))).Should().BeTrue();
+        (await RunBidirectionalDeadlockTest(
+            s => s.Sort(SortExpressionComparer<Person>.Ascending(p => p.Age)).Virtualise(req),
+            subjectPusher: () => { for (var j = 0; j < ItemCount; j++) req.OnNext(new VirtualRequest(j * 5, 25 + (j % 4) * 25)); })).Should().BeTrue();
```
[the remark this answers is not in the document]
dwcullop: Good catch. Applied `UnsynchronizedMerge` to both `SortAndPage` and `SortAndVirtualize` (commit 88c9f2fc): `SortAndPage` swaps the static `Observable.Merge(a,b,c)` form for the extension-method form, and `SortAndVirtualize` collapses chained `.Merge(b).Merge(c)` into a single `UnsynchronizedMerge(b, c)` (removing the second gate too). Added `SortAndPage_DoesNotDeadlock` and `SortAndVirtualize_DoesNotDeadlock` to `DeadlockTortureTest` alongside the existing `Sort().Page()` and `Sort().Virtualise()` tests (older coverage preserved). `MultiplePairs_Simultaneous_NoDeadlock` also gains parallel `SortAndPage` and `SortAndVirtualize` lanes with their own subjects.

src/DynamicData/Internal/DeliveryQueueMergeExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+            var subscriptions = new CompositeDisposable(totalSources + 1);
+            var pending = totalSources;
+
+            // Each source needs its own inner observer instance because Rx's ObserverBase
+            // sets a one-shot stopped flag on the first OnCompleted/OnError; a single shared
+            // observer would silently drop terminal notifications from every source after
+            // the first. OnNext and OnError forward straight to the queue (the queue's gate
+            // serializes concurrent calls); OnCompleted is counter-gated so only the last
+            // surviving source's completion terminates the merged stream.
+            IObserver<T> CreateInner() =>
```
JakenVeina, line 46: General convention is that local methods should be placed at the end of their containing method, is it not?

src/DynamicData/Internal/DeliveryQueueMergeExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+/// concurrent producers but releases its gate before delivering, so a downstream
+/// observer that walks into another cache's writer lock cannot deadlock with this
+/// operator's serialization point.</para>
+/// <para>Every input must share the same element type. When the inputs have different
+/// element types or require operator-private projections invoked inside the queue's
+/// drain, use <see cref="SynchronizeSafeExtensions.SynchronizeSafe{T}(IObservable{T}, SharedDeliveryQueue)"/>
+/// with a <see cref="SharedDeliveryQueue"/> and finish with
+/// <see cref="SynchronizeSafeExtensions.UnsynchronizedMerge{T}(IObservable{T}, IObservable{T}[])"/>.</para>
+/// </remarks>
+internal static class DeliveryQueueMergeExtensions
```
JakenVeina, line 18: Internal code shouldn't be XML-doc'd. Just put some normal comments to clarify whatever's not obvious in the code.

src/DynamicData/Internal/DeliveryQueueMergeExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// </summary>
+    /// <typeparam name="T">The element type, common to every input.</typeparam>
+    /// <param name="first">The first input observable.</param>
+    /// <param name="others">Additional input observables.</param>
+    /// <returns>An observable that emits items from every input, serialized through the queue.</returns>
+    public static IObservable<T> DeliveryQueueMerge<T>(this IObservable<T> first, params IObservable<T>[] others) =>
+        Observable.Create<T>(observer =>
+        {
+            var queue = new DeliveryQueue<T>(observer);
+            var totalSources = others.Length + 1;
```
JakenVeina: Redundant variable.

src/DynamicData/Internal/DeliveryQueueMergeExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+            var subscriptions = new CompositeDisposable(totalSources + 1);
+            var pending = totalSources;
```
JakenVeina: Poor variable name. I'd go with something like `frameDepth` or `nestingDepth`.

src/DynamicData/Internal/SynchronizeSafeExtensions.cs
JakenVeina, line 87: I find it hard to believe that the native `.Merge()` has enough overhead in non-contested scenarios to make this worthwhile. Got any numbers to back this up?

JakenVeina (changes requested):
I'm gonna be honest, this puts another point on my "I feel like this whole DrainQueue thing is a bad idea" column.

I mentioned before that I think the fundamental reason it makes me queasy is that it feels like we're trying to re-invent how synchronization/scheduling works, when RX has already established a paradigm for that. And now, because it's such a fundamental paradigm, we're creeping into "re-inventing RX" territory.

Like, we already had to re-implement `.Synchronize()` with `SynchronizeSafe()` operator, and now we're doing the same for `.Merge()`. How many other native RX operators use locking internally that are going to need this same treatment?

That being said, let's still move forward with getting this into a preview, but I think any more issues like this, where we discover that the DQ approach is incompatible with native RX (or at least, that it doesn't fully resolve the ABBA looping problem), and that's going to confirm for me that the we need to roll things back.

src/DynamicData/Internal/SynchronizeSafeExtensions.cs
dwcullop: Fair point.  We only hit it rarely on the deliberately brutal kitchen sink stress test, which is supposed to finesse such problems.  It's not about the overhead.  It's about holding a lock.  We can't do it without allowing an ABBA situation.

That said, if we can hit it, then end users will hit it too.  Maybe even only rarely, which is worse, because that's how it gets into production.

dwcullop:
> I mentioned before that I think the fundamental reason it makes me queasy is that it feels like we're trying to re-invent how synchronization/scheduling works, when RX has already established a paradigm for that. And now, because it's such a fundamental paradigm, we're creeping into "re-inventing RX" territory.

I understand your concern.  I really do, but the standard Rx "hold a lock when delivering" model is fundamentally insufficient and problematic, especially for complex usages of Rx like DD.  Rx has no other solution.  There have been multiple issues opened:
- [Possible Deadlock in BehaviorSubject.Subscribe](https://github.com/dotnet/reactive/issues/2080) (Still open)
- [Deadlock on OSX](https://github.com/dotnet/reactive/issues/1092) (Closed without a fix) 
**Calls out DynamicData specifically**

[Even in the Rx code itself](https://github.com/dotnet/reactive/blob/a363e7942b6b22e4f3603e84b9c90cfe025e5bc8/Rx.NET/Source/src/System.Reactive/Linq/Observable/RefCount.cs#L176-L190) they talk about how they have to work around the issue we're actually fixing.

Your broader concern, "how much more time will we spend playing whack-a-mole?", is very valid.  I will do more digging and see if I can find all the places that:
A) Use an Rx operator that acquires a lock
B) Mixes it with a DQ operation

There might be one or two more we have to implement (but the implementations are trivial because they don't require locking).

A bigger question is "Does this fix belong in DD?" and the answer is clearly no, this is something Rx should provide for.  But it doesn't.  Not yet.  Maybe we can drive that once we have a solution that works.

src/DynamicData/Internal/DeliveryQueueMergeExtensions.cs
dwcullop: It needs to indicate that's the number that's left because it counts down.  I'll pick a better name though.

src/DynamicData/Internal/DeliveryQueueMergeExtensions.cs
dwcullop: Fixed

src/DynamicData/Internal/DeliveryQueueMergeExtensions.cs
dwcullop: Fixed

src/DynamicData/Internal/DeliveryQueueMergeExtensions.cs
dwcullop: Fixed

src/DynamicData/Internal/DeliveryQueueMergeExtensions.cs
dwcullop: Fixed

src/DynamicData/Internal/SynchronizeSafeExtensions.cs
dwcullop: I only found one more operator we need to implement:  CombineLatest.  None of the other 13 Rx operators that hold a gate are being used.   I did find other places we need to use the updated Merge though.

JakenVeina: No, I mean like.... what's the purpose of having an unsynchronized `.Merge()`? I.E. a `.Merge()` where we skip any locking or queueing because we assume that all the upstreams are already subject to locking or queueing. That's how I read this, and my interpretation was that we'd want this for the sake of optimizing-away synchronization overhead, in such scenarios. Is that not what's going on here?

JakenVeina: > None of the other 13 Rx operators that hold a gate are being used. 

For reference, what are all these other operators? We might want to notate that somewhere, in the event we ever DO want to use them, internally.

JakenVeina:
I suppose there is actually one more aspect that makes me quesy: I've never experienced it as a limitation, myself. I actively avoid ABBA scenarios, because I think they're generally poor design, and in my mind, RX isn't SUPPOSED to support them. And every time I've actually encountered a case like that, it was easily solved with proper use of an `IScheduler`.

> this is something Rx should provide for. But it doesn't

Whereas, my opinion is the opposite: I agree with RX not supporting it.

That being said, "it's not a problem for me personally" isn't really a good argument, so I'm gonna keep trying to suppress that impulse.

So, really, it boils down to "How much effort should we spend trying to solve a problem that isn't really ours to solve?"

dwcullop:
Okay, consider this:  Rx doesn't have this problem, but DD does.  DD inherently has to maintain the state of the cache AND emit the changes.  This requires synchronization.  Some of the operators require further synchronization.  So, then the question is "Is cross-cache manipulation a valid use-case".  If it is, then we need to deal with the synchronization because Rx doesn't.

I would say it is valid.  It might not be common, but I don't think it is unreasonable to have two caches that consume each other.

The fix we've got now is incomplete.  This PR fills in some gaps.  In a lot of cases, the gap is that we're using a lock where we no longer need one.  Removing it will eliminate a chance for a deadlock and improve throughput.

src/DynamicData/Internal/SynchronizeSafeExtensions.cs
dwcullop: I updated the PR description with the results of the audit I did for use of synchronizing operators used in DD.  Let me know if you think we should pull it out as repo instructions.

dwcullop: To be clear `UnsynchronizedMerge` isn't about optimizing away a lock where we don't need one.  It's about removing a lock where we can't have one.  We *also* don't need one but that's not the motivation.  It's needed because adding a lock AFTER the delivery queue just re-introduces the problem that the delivery queue is meant to solve.

JakenVeina:
>  So, then the question is "Is cross-cache manipulation a valid use-case".

I mean, you know my position on this. It's "no".

I've said before that side-effects are considered bad-practice in RX, but I'll be more specific: It's been a well-understood problem with RX as a mechanism of state-maintenance since Redux became popular. And DD is basically exactly the same architecture as redux: central state stores that publish state change notifications to be transformed and fed into a UI.

Here's their take on the matter: https://redux.js.org/usage/side-effects-approaches

TL;DR don't do side-effects within your main queries. Redux has mechanisms explicitly for doing side-effects, where they happen asynchronously from the main query. That's why I point to `IScheduler`s in RX.NET as the proper mechanism for handling side-effects.

src/DynamicData/Internal/SynchronizeSafeExtensions.cs
JakenVeina: I don't see how that's correct. The issue is the synchronization between upstream and downstream, which the drain queue breaks, no matter what the downstream is doing. If all inputs to a `.Merge()` are already synchronized and serialized via a drain queue, then any `lock`ing within the `.Merge()` becomes effectively a NOP. The only purpose that can be served by removing the `lock`ing is to eliminate overhead. I see the cost of maintaining a new operator as way worse than the cost of just accepting that tiny bit of overhead.

dwcullop:
@JakenVeina  Can we proceed with this one?  I know you're feelings on DQ, but what we have now in `main` is incomplete.  This PR closes the gap and will prevent some of the test flakes we're seeing.

src/DynamicData/Internal/SynchronizeSafeExtensions.cs
dwcullop: Any operator that holds a lock while emitting downstream adds an edge to the lock ordering, and two of them is all it takes for ABBA: A holds its gate and emits into B where it needs B's gate, while a second path has B holding its gate and emitting into A. Rx's `.Merge()` holds its gate across `ForwardOnNext`, so every one we use is another edge in that graph. The unsynchronized versions hold nothing while emitting, so they can't take part in a cycle at all.

dwcullop: Pulled the audit out of the PR description and into `.github/instructions/rx-gate-holding-operators.instructions.md`, since a PR description is not somewhere anyone is going to find it later. It carries the rule, the three helpers and their precondition, the full coverage matrix, and a note to update the table when an operator starts or stops using one of these combinators.

JakenVeina:
I'd really rather not, just yet. The higher priority for me is all these defects we're trying to fix, and that we'll THEN have to backport to 9.x. The more DQ work we merge in, the more work that backporting is going to be. And there's no way the DQ implementation is going to be staying as it is.

The Primitives migration and getting a VNext prototype published for RUI is also a higher priority, for me.

I promise I have been keeping the DQ stuff in the back of my mind through all of this, and intend to give it REAL attention soon.

dwcullop:
Understood on priorities, and I'm not reopening the DQ debate.

But I'd push back on parking this behind Primitives specifically. Seventeen of the eighteen library files this touches are also modified by #1116. If Primitives lands first, this has to be redone against a branch that rewrote nearly all of them. If this lands first, Primitives picks it up in a rebase it's already doing. Waiting doesn't defer the work, it multiplies it, and it lands on whoever rebases afterwards.

It's ready now: CI is green and I've merged main into it ten times keeping it current.

What I don't want to ship is the half-converted state. Main has thirteen cache operators routing inputs through a `SharedDeliveryQueue` and then recombining them with `Merge` or `CombineLatest`, which puts a held gate straight back across delivery. They pay the cost of the queue design and get none of the benefit. `Page` is the clearest case.

I'd like this in before the v10 preview. Happy to take another review pass if you want one.

PR #1101, opened 2026-05-29
Rewrote testing for the cache variant of the AutoRefresh() and AutoRefreshOnObservable() operators,...

### LINKED ISSUE #1099: [Bug]: AutoRefreshOnObservable produces refreshes for immediately-removed items
Feeding the following sequence of changesets into `.AutoRefreshOnObservable()` (for the cache version, the list version doesn't seem to have this issue, it just emits an empty changeset), when the refresh observable emits a value immediately upon subscription, results in the following downstream sequence changesets, which includes an invalid `Refresh` for an item that was previously removed from the collection.

**Upstream**
```
2 Changes
    Add #1
    Remove #1
```

**Downstream**
```
2 Changes
    Add #1
    Remove #1
1 Change
    Refresh #1
```

If the upstream changeset is split up to have the `Add` and `Remove` be in separate changesets, there is no issue, as follows:

**Upstream**
```
1 Change
    Add #1
1 Change
    Remove #1
```

**Downstream**
```
1 Change
    Add #1
1 Change
    Refresh #1
1 Change
    Remove #1
```

This doesn't seem to be an issue with any of the operators used internally within `AutoRefreshOnObservable()` (E.G. `.MergeMany()` and `.SubscribeMany()`), but with the sequencing of subscriptions within `.AutoRefresh()` itself.

I discovered this while expanding our testing suite for `.FilterOnObservable()`, which ends up occasionally including items in its downstream that were removed upstream.

JakenVeina:
…in accordance with #1014, and in a preliminary effort to resolve #1099.

### DIFF
--- a/src/DynamicData.Tests/Cache/AutoRefreshOnObservableFixture.Base.cs
+++ b/src/DynamicData.Tests/Cache/AutoRefreshOnObservableFixture.Base.cs
@@ -1,0 +1,777 @@
+using System;
+using System.Linq;
+using System.Reactive;
+using System.Reactive.Concurrency;
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
+
+using Microsoft.Reactive.Testing;
+
+using FluentAssertions;
+using Xunit;
+
+using DynamicData.Tests.Utilities;
+
+namespace DynamicData.Tests.Cache;
+
+public static partial class AutoRefreshOnObservableFixture
+{
+    public abstract class Base
+    {
+        [Fact]
+        public void ChangeSetBufferIsGiven_ReevaluatorNotificationsAreBufferedOnScheduler()
+        {
+            // Setup
+            using var source = new TestSourceCache<Item, int>(Item.SelectId);
+
+            using var item1 = new Item() { Id = 1 };
+            using var item2 = new Item() { Id = 2 };
+            using var item3 = new Item() { Id = 3 };
+            
+            source.AddOrUpdate(new[] { item1, item2, item3 });
+
+            var scheduler = new TestScheduler();
+
+
+            // UUT Initialization
+            using var subscription = BuildUut(
+                    source:             source.Connect(),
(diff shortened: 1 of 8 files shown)


src/DynamicData.Tests/Cache/AutoRefreshOnObservableFixture.Base.cs
```diff
@@ -0,0 +1,765 @@
...
+            results.Error.Should().BeNull();
+            results.RecordedChangeSets.Skip(1).Count().Should().Be(1, "one source operation was performed");
+            results.RecordedItemsByKey.Values.Should().BeEquivalentTo(source.Items, "1 item was replaced within the source");
+            results.HasCompleted.Should().BeFalse("the source has not completed");
+            
+            item2.HasObservers.Should().BeFalse("replacing an item should trigger unsubscription from its reevaluator");
+            item4.HasObservers.Should().BeTrue("adding an item should invoke its reevaluator and subscribe to it");
+            item1.HasObservers.Should().BeTrue("the item was not removed from the source");
+            item3.HasObservers.Should().BeTrue("the item was not removed from the source");
+        }
```
dwcullop, line 364: This test stops after asserting item4.HasObservers == true, but never verifies that the new reevaluator's notifications actually flow through end-to-end. A regression where the operator subscribes to the new item's reevaluator but mishandles the resulting notification (wrong key, dropped Refresh, etc.) would slip through. Consider adding ++item4.Value followed by an assertion that the next recorded changeset contains a Refresh for item4.

dwcullop (approved):
Minor nit but looks good to me.

src/DynamicData.Tests/Cache/AutoRefreshOnObservableFixture.Base.cs
JakenVeina: Agreed.

PR #1113, opened 2026-06-14
FilterImmutable: emit correct Current value when Update transitions to Remove

### LINKED ISSUE #1112: FilterImmutable emits Remove with the wrong Current value for Update->Remove transitions
### Description

`FilterImmutable` emits `ChangeReason.Remove` with the wrong value in `Current` when an Update entry transitions from matching the predicate to not matching it.

### Repro

For an Update where the previous value satisfies the predicate and the current value does not, the operator emits a `Remove` whose `Current` holds the *new* (non-matching) value. The previously-included value is lost.

```csharp
using var source = new Subject<IChangeSet<Item, int>>();
using var results = source.FilterImmutable(predicate: item => item.IsIncluded).AsAggregator();

var included = new Item { Id = 1, IsIncluded = true };
var excluded = new Item { Id = 1, IsIncluded = false };

source.OnNext(new ChangeSet<Item, int> {
    new(ChangeReason.Add, included.Id, current: included, index: 0)
});

source.OnNext(new ChangeSet<Item, int> {
    new(ChangeReason.Update, excluded.Id, current: excluded, previous: included, currentIndex: 0, previousIndex: 0)
});

var removeChange = results.Messages.Last().Single();
// removeChange.Reason   == ChangeReason.Remove        // correct
// removeChange.Current  == excluded (IsIncluded=false) // WRONG, should be `included`
```

### Expected

Per the `Change<T,K>` contract, `Remove.Current` is the value that just left downstream. For Update->Remove, that is `Previous`: `Current` never reached downstream.

### Impact

Consumers reading `Current` on a Remove get garbage references:

- Composition with `TransformImmutable` throws `InvalidCastException` if the downstream transform is type-narrowing.
- Side-effect handlers (e.g. disposing, releasing handles, unsubscribing) act on the wrong object.
- Re-keying or projecting on Remove silently produces wrong results.

The stateful `Filter` handles this correctly via `ChangeAwareCache`; the bug is specific to `FilterImmutable`.

### Affected versions

All versions that include `FilterImmutable`.

dwcullop:
Fixes #1112.

### Problem

`FilterImmutable` emits `ChangeReason.Remove` carrying the wrong value in `Current` when an Update changeset entry transitions from matching to not matching the predicate. The emitted Remove carries the new (non-matching) `Current` instead of the previous (matching) value that actually left downstream.

This violates the `Change<T,K>` contract that `Remove.Current` is the item being removed. Consumers reading `Current` on Remove receive the wrong reference, producing silent corruption (wrong side effects, incorrect projections) or `InvalidCastException` when composed with type-narrowing operators.

### Fix

When emitting Update->Remove, carry `change.Previous.Value` as `Current`. `Previous` is dropped (Remove is not an Update).

### Test

`Update_PreviousMatchedCurrentDoesNot_EmitsRemoveCarryingPreviousAsCurrent` in `FilterImmutableFixture` covers the contract. RED before the fix, GREEN after.

### Verification

- All 12 tests in `FilterImmutableFixture` pass.
- All 1365 tests in `DynamicData.Tests.Cache` pass.
- Full suite has 2 pre-existing flakes in `DeadlockTortureTest` (`TransformWithForce_DoesNotDeadlock`, `AllDangerous_Stacked_DoNotDeadlock`) that reproduce on unmodified `main`. Unrelated to this change.

### DIFF
--- a/src/DynamicData.Tests/Cache/FilterImmutableFixture.cs
+++ b/src/DynamicData.Tests/Cache/FilterImmutableFixture.cs
@@ -4,6 +4,8 @@ using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 
+using DynamicData.Tests.Utilities;
+
 using FluentAssertions;
 using Xunit;
 
@@ -16,9 +18,10 @@ public sealed class FilterImmutableFixture
     {
         using var source = new Subject<IChangeSet<Item, int>>();
 
-        using var results = source
+        using var subscription = source
             .FilterImmutable(predicate: Item.Predicate)
-            .AsAggregator();
+            .ValidateChangeSets(Item.KeySelector)
+            .RecordCacheItems(out var results);
 
 
         // Add items
@@ -31,8 +34,8 @@ public sealed class FilterImmutableFixture
         });
 
         results.Error.Should().BeNull();
-        results.Messages.Count.Should().Be(1, "1 source operation was performed");
-        results.Data.Items.Should().BeEquivalentTo(new[] { item1 }, "2 items were added, with 1 excluded");
+        results.RecordedChangeSets.Count.Should().Be(1, "1 source operation was performed");
+        results.RecordedItemsByKey.Values.Should().BeEquivalentTo(new[] { item1 }, "2 items were added, with 1 excluded");
 
 
         // Replace items, changing inclusion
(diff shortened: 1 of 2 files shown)


src/DynamicData.Tests/Cache/FilterImmutableFixture.cs
```diff
@@ -364,6 +364,39 @@ private static void ManipulateExcludedItems(ISubject<IChangeSet<Item, int>> sour
         });
     }
 
+    [Fact]
```
JakenVeina, line 376: Instead of adding a new test, we can just add `.ValidateChangeSets()` to all the existing tests. That's got all the proper changeset contract rules baked in. And probably swap `.AsAggregator()` for `.RecordCacheItems()` while you're at it. That elicits a failure in `ItemsAreManipulated` on my end.

JakenVeina (changes requested):

src/DynamicData.Tests/Cache/FilterImmutableFixture.cs
dwcullop: Did both. Swapped .AsAggregator() for .ValidateChangeSets(Item.KeySelector).RecordCacheItems(out var results) across all existing tests, and kept the new focused Remove.Current regression test. All 12 tests pass with the fix; ItemsAreManipulated does fail under ValidateChangeSets without the fix. Pushed in 5bb84c81.

JakenVeina (approved):

PR #1114, opened 2026-06-14
Orchestrator: Operator for Composing Per-Item Observables

dwcullop:
## What this is

A common pattern in cache operators: for each item in a source changeset, maintain a separate external subscription whose values feed into the operator's output. `MergeMany` subscribes to one observable per item. `AutoRefresh` subscribes to one re-evaluator per item. `FilterOnObservable` subscribes to one bool stream per item. `TransformManyAsync`, `GroupOnObservable`, and several others do the same. The shape is universal: 1:1 between source items and external subscriptions, with results flowing back into a single output.

Getting that right involves a lot of boilerplate: spin up a subscription on Add/Update, tear it down on Remove, route everything through the same serialization gate to avoid the kind of cross-cache deadlocks that motivated #1097, keep the downstream stream alive until both the source and every per-item subscription have completed, dispose cleanly on error or unsubscribe.

`CacheParentSubscription` was the previous primitive for cache operators that needed parent/child subscription accounting. It worked, but every consumer had to subclass it and own per-subscription state as subclass fields. That made it easy to leak state across subscribers or forget to wire up the queue-drain machinery from #1097. It also didn't compose: operators wanting parent/child accounting AND a `ChangeAwareCache`, or AND a merge tracker, had to inline both concerns.

This PR replaces it with **Orchestrator**: one primitive that handles subscription accounting, queue routing, completion accounting, and disposal. Operator authors write a small implementation describing what their operator does. The runtime handles how.

Two interfaces and a driver:

- `ICacheOrchestrator<TSource, TKey, TInner, TResult>`: authors implement this. Three methods (`OnSourceChangeSet`, `OnInner`, `OnDrainComplete`).
- `ICacheOrchestratorContext<TKey, TInner>`: authors call this. Three methods (`Track`, `Untrack`, `Serialize`).
- `CacheOrchestration<...>`: the driver. Per-subscription instances built by a caller-supplied factory, so all state is naturally isolated.

## Architecture

```mermaid
flowchart TB
    subgraph public[Public extension methods]
        ops[AutoRefresh · MergeMany · MergeManyChangeSets ·<br/>FilterOnObservable · TransformManyAsync ·<br/>GroupOnObservable · TransformOnObservable · MergeManyItems ·<br/>MergeManyListChangeSets]
    end

    subgraph specializations[Internal specializations]
        oc[OrchestrateChangeSets<br/>mirror-manipulator with ChangeAwareCache]
        omc1[OrchestrateManyChangeSets cache<br/>merges child cache changesets]
        omc2[OrchestrateManyChangeSets list<br/>merges child list changesets]
    end

    subgraph base[Internal base entry points]
        ot[Orchestrate&lt;TOrch&gt;<br/>typed factory]
        ol[Orchestrate lambdas<br/>convenience overload]
    end

    subgraph driver[Driver]
        co[CacheOrchestration<br/>per-subscription state owner]
    end

    subgraph contract[Contract]
        ico[ICacheOrchestrator<br/>what you implement]
        ictx[ICacheOrchestratorContext<br/>what you call]
        cob[CacheOrchestratorBase<br/>per-ChangeReason dispatch convenience]
    end

    subgraph runtime[Runtime]
        sdq[SharedDeliveryQueue + DeliverySubQueue<br/>lock-released serialized delivery]
    end

    ops --> specializations
    ops --> ol
    ops --> ot
    specializations --> ot
    ot --> co
    ol --> co
    co --> sdq
    co -.implements.-> ictx
    co -.calls.-> ico
    cob -.implements.-> ico

    style public          fill:#FFCDD2,stroke:#C62828,color:#000
    style specializations fill:#FFE0B2,stroke:#E65100,color:#000
    style base            fill:#FFF9C4,stroke:#F57F17,color:#000
    style driver          fill:#C8E6C9,stroke:#2E7D32,color:#000
    style contract        fill:#BBDEFB,stroke:#1565C0,color:#000
    style runtime         fill:#E1BEE7,stroke:#6A1B9A,color:#000
```

### `ICacheOrchestrator<TSource, TKey, TInner, TResult>` (you implement)

| Method | Called when | What you do |
|---|---|---|
| `OnSourceChangeSet(IChangeSet<TSource, TKey> changes)` | A changeset arrives from the keyed source | Walk the changes. Usually `context.Track(key, ...)` on Add/Update, `context.Untrack(key)` on Remove. Update any per-key state you keep. |
| `OnInner(TInner value, TKey key)` | A tracked inner observable emits, paired with its source key | Accumulate into output state, or forward to the emitter. |
| `OnDrainComplete(bool isFinal, bool wasReentrant)` | A drain cycle of the shared queue finished | Flush accumulated state. `isFinal == true` means the source and every tracked inner have terminated, so flush synchronously even if you'd normally defer (timer buffering, etc.). `wasReentrant` distinguishes a reentrant drain from a clean cycle; most orchestrators ignore it. |

### `ICacheOrchestratorContext<TKey, TInner>` (you call)

| Method | Called when | What it does |
|---|---|---|
| `Track(TKey key, IObservable<TInner> observable)` | A new source item appears (Add/Update) and needs a per-key inner observable | Subscribes the observable through the shared queue. If a prior subscription exists for the key, it gets disposed and replaced. Participates in completion accounting: downstream stays alive until every tracked subscription terminates. |
| `Untrack(TKey key)` | A source item disappears (Remove) | Disposes the inner subscription for that key. No-op if none. |
| `Serialize<T>(IObservable<T> observable)` | An ad-hoc observable (timer, separate cache stream) needs to go through the same gate as source and inner notifications | Wraps the observable with the queue's synchronization gate. Does NOT participate in completion accounting; downstream can complete while a Serialize-wrapped subscription is still active. |

## Lifecycle

```mermaid
%%{init: {"themeVariables": {"sequenceRectStroke": "#000000", "sequenceRectStrokeWidth": "1px"}}}%%
sequenceDiagram
    box rgba(255, 205, 210, 1) Consumer
        participant Sub as Subscriber
    end
    box rgba(255, 224, 178, 1) Driver
        participant Driver as CacheOrchestration
    end
    box rgba(255, 249, 196, 1) Runtime
        participant Q as SharedDeliveryQueue
    end
    box rgba(200, 230, 201, 1) Orchestrator
        participant Orch as Your orchestrator
    end
    box rgba(187, 222, 251, 1) Source
        participant Src as Source changeset
    end
    box rgba(225, 190, 231, 1) Inner
        participant Inn as Inner observable
    end

    Sub->>Driver: Subscribe
    Driver->>Q: new SharedDeliveryQueue
    Driver->>Orch: factory(context, emitter)
    Driver->>Src: SynchronizeSafe(queue).SubscribeSafe

    rect rgba(245, 245, 245, 1)
        note over Q: drain cycle
        Src->>Q: OnNext(changeset)
        Q->>Orch: OnSourceChangeSet(changeset)
        Orch->>Driver: ctx.Track(key, innerObs)
        Driver->>Inn: SynchronizeSafe(queue).SubscribeSafe
        Inn->>Q: OnNext(value)
        Q->>Orch: OnInner(value, key)
        Q->>Orch: OnDrainComplete(isFinal, wasReentrant)
        Orch->>Q: emitter.OnNext(result)
        Q->>Sub: result
    end
```

## The Orchestrate family: five entry points

Two base entry points (typed factory and lambda overload) plus three specializations that bake in common output shapes.

### Base: `Orchestrate`

Same `CacheOrchestration` driver underneath. Use the typed factory when state and helpers want to live in a class. Use the lambda overload when closures suffice.

```csharp
public static IObservable<TResult> Orchestrate<TSource, TKey, TInner, TResult, TOrch>(
    this IObservable<IChangeSet<TSource, TKey>> source,
    Func<ICacheOrchestratorContext<TKey, TInner>, IObserver<TResult>, TOrch> factory)
    where TOrch : ICacheOrchestrator<TSource, TKey, TInner, TResult>;

public static IObservable<TResult> Orchestrate<TSource, TKey, TInner, TResult>(
    this IObservable<IChangeSet<TSource, TKey>> source,
    Action<IChangeSet<TSource, TKey>, ICacheOrchestratorContext<TKey, TInner>> onSourceChangeSet,
    Action<TInner, TKey, IObserver<TResult>> onInner,
    Action<IObserver<TResult>>? onDrainComplete = null);
```

### `OrchestrateChangeSets`: cache changeset output via `ChangeAwareCache`

For filter-like or per-key transform shapes: source is a keyed changeset, output is a keyed changeset, each source key contributes 0 or 1 items to the output. The runtime gives you a `ChangeAwareCache` that you mutate from both source and inner events. It captures and emits once per drain cycle. Used by `FilterOnObservable` and `TransformOnObservable`.

```csharp
public static IObservable<IChangeSet<TOutput, TKey>> OrchestrateChangeSets<TSource, TKey, TInner, TOutput>(
    this IObservable<IChangeSet<TSource, TKey>> source,
    Func<TSource, TKey, IObservable<TInner>> innerFactory,
    Action<ChangeAwareCache<TOutput, TKey>, Change<TSource, TKey>> onSourceChange,
    Action<ChangeAwareCache<TOutput, TKey>, TKey, TSource, TInner> onInner);
```

### `OrchestrateManyChangeSets` (cache): merge child cache changesets

The shape `MergeManyChangeSets` produces: per-source-key inner observables that are themselves cache changesets, merged into one cache changeset output. The runtime gives you a `ChangeSetMergeTracker` that handles add/update/remove/refresh and conflict resolution via the supplied comparers. Used by both `MergeManyChangeSets` overloads (vanilla and source-priority).

```csharp
public static IObservable<IChangeSet<TDest, TDestKey>> OrchestrateManyChangeSets<TSource, TKey, TDest, TDestKey>(
    this IObservable<IChangeSet<TSource, TKey>> source,
    Func<TSource, TKey, IObservable<IChangeSet<TDest, TDestKey>>> changeSetSelector,
    IEqualityComparer<TDest>? equalityComparer = null,
    IComparer<TDest>? comparer = null,
    bool reevalOnRefresh = false);
```

### `OrchestrateManyChangeSets` (list): merge child list changesets

Same shape but for list-typed inner changesets. Disambiguated by parameter type (no destination key). Used by `MergeManyListChangeSets`.

```csharp
public static IObservable<IChangeSet<TDest>> OrchestrateManyChangeSets<TSource, TKey, TDest>(
    this IObservable<IChangeSet<TSource, TKey>> source,
    Func<TSource, TKey, IObservable<IChangeSet<TDest>>> changeSetSelector,
    IEqualityComparer<TDest>? equalityComparer = null);
```

## Picking an entry point

| Operator shape | Use |
|---|---|
| Stateful (class fits naturally), value or changeset output | `Orchestrate<TOrch>` typed factory, optionally with `CacheOrchestratorBase` for per-`ChangeReason` virtual hooks |
| Stateless, or state small enough for closures | `Orchestrate` lambda overload |
| Cache changeset output built by mutating a `ChangeAwareCache` | `OrchestrateChangeSets` |
| Union of child cache changesets | `OrchestrateManyChangeSets` (cache) |
| Union of child list changesets | `OrchestrateManyChangeSets` (list) |

### Worked example: `MergeMany`

```csharp
public IObservable<TDestination> Run() =>
    _source.Orchestrate<TObject, TKey, TDestination, TDestination>(
        onSourceChangeSet: (changes, context) =>
        {
            foreach (var change in changes.ToConcreteType())
            {
                if (change.Reason is ChangeReason.Add or ChangeReason.Update)
                    context.Track(change.Key, _observableSelector(change.Current, change.Key));
                else if (change.Reason is ChangeReason.Remove)
                    context.Untrack(change.Key);
            }
        },
        onInner: (value, _, emitter) => emitter.OnNext(value));
```

The whole operator is one method-call expression. The pre-orchestrator version subclassed `CacheParentSubscription` with its own fields, constructor, and manual `OnNext`/`Subscribe` plumbing.

## Operators migrated

| Operator | Entry point |
|---|---|
| `AutoRefresh` | `Orchestrate<TOrch>` + custom orchestrator (timer state, pending-refresh map) |
| `FilterOnObservable` | `OrchestrateChangeSets` |
| `GroupOnObservable` | `Orchestrate` lambda (closes over a `DynamicGrouper`) |
| `MergeMany` | `Orchestrate` lambda |
| `MergeManyItems` | `Orchestrate` lambda |
| `MergeManyChangeSets` (cache) | `OrchestrateManyChangeSets` (cache) |
| `MergeManyChangeSets` (source-compare) | `OrchestrateManyChangeSets` (cache) with per-item parent-child boxing |
| `MergeManyListChangeSets` | `OrchestrateManyChangeSets` (list) |
| `TransformManyAsync` | `Orchestrate<TOrch>` + custom orchestrator (per-key cache, merge tracker) |
| `TransformOnObservable` | `Orchestrate` lambda (closes over a `ChangeAwareCache`) |

## AutoRefresh correctness fixes

The rewrite fixes four pre-existing defects that had previously-skipped tests. All four tests are un-skipped here and pass:

- **`PropertyAccessorIsNull_ThrowsException`**: `propertyAccessor` is now validated upfront. Previously threw `NullReferenceException` on the first source notification.
- **`ReevaluatorIsNull_ThrowsException`**: `reEvaluator` is now validated upfront. Same prior failure mode.
- **`ReevaluatorEmitsImmediately_ItemDoesNotRefresh`** (#1099): a reevaluator that fires synchronously during item subscription no longer produces a redundant Refresh paired with the Add.
- **`ReevaluatorThrows_ExceptionPropagates`**: reevaluator exceptions now propagate to `OnError` instead of being silently swallowed (Rx contract §6.4).

## Other fixes

- **`AutoRefresh` drops a pending refresh when the source forwards its own Refresh.** A reevaluator-queued refresh from a prior drain was being emitted in addition to the source's Refresh for the same key. Symmetrical to the existing pending-drop logic on Update/Remove. Regression test added.
- **`isFinal` is computed from the subscription counter, not a latched flag.** A latched flag races when the orchestrator calls `Track` during the `isFinal` callback: counter goes back to 1 but the flag still says true, so `OnCompleted` fires while a live inner exists. The counter is the source of truth.
- **Construction error path.** The constructor's try/catch now covers every step from queue allocation through the source subscribe. Any throw during construction releases everything successfully allocated rather than leaking.
- **Deleted `MergeManyCacheChangeSets` and `MergeManyCacheChangeSetsSourceCompare` wrapper classes.** Inlined their bodies into the public `MergeManyChangeSets` overloads. The source-compare overload's supporting types now live as private nested generic types in the same partial-class file as the public API.
- **`Virtualise`, `GroupOn`, `GroupOnImmutable`: collapsed per-branch null/empty filters into single post-merge filters.** Matches the `Page` shape and removes duplicated filtering on every merge input.

## Tests

- `OrchestrateFixture` (16 tests): primitive contract. Source/inner serialization, per-drain coalesced emission, completion accounting (parent and all children), error propagation, cross-cache deadlock proofs, reentrant drain.
- `SharedDeliveryQueueFixture`, `DeliveryQueueFixture`: queue contract.
- Per-orchestrator isolation: `AutoRefreshOrchestratorFixture`, `ChangeSetOrchestratorFixture`, `LambdaCacheOrchestratorFixture`, `MergedOrchestratorFixture`, `MergedListOrchestratorFixture`, `TransformManyAsyncOrchestratorFixture`.
- Existing public-API operator fixtures continue to pass unchanged.

Full Cache + Internal suite: pass. `DeadlockTortureTest` fixture (added in #1097): pass.

## Public surface

No new or changed public APIs.

### Behavioral change: per-item observable errors now propagate

Several existing public operators previously **swallowed** errors from per-item observables (the offending item was silently unsubscribed and the merged stream continued). They now **forward** those errors, terminating the merged stream. This matches the contract of standard `Observable.Merge`.

**Cache side:**

| Operator | Old behavior | New behavior |
| --- | --- | --- |
| `MergeMany` | Child errors swallowed | Child errors terminate the output |
| `WhenAnyPropertyChanged` (routes through `MergeMany`) | Per-item PropertyChanged errors swallowed | Per-item errors terminate the output |
| `WhenPropertyChanged` (routes through `MergeMany`) | Per-item subscription errors swallowed | Per-item errors terminate the output |
| `WhenValueChanged` (routes through `MergeMany`) | Per-item subscription errors swallowed | Per-item errors terminate the output |
| `AutoRefreshOnObservable` | Per-item reevaluator errors swallowed | Per-item errors terminate the output |

**List side:**

| Operator | Old behavior | New behavior |
| --- | --- | --- |
| `MergeMany` | Child errors swallowed | Child errors terminate the output |
| `WhenAnyPropertyChanged` (routes through `MergeMany`) | Per-item PropertyChanged errors swallowed | Per-item errors terminate the output |
| `WhenPropertyChanged` (routes through `MergeMany`) | Per-item subscription errors swallowed | Per-item errors terminate the output |
| `WhenValueChanged` (routes through `MergeMany`) | Per-item subscription errors swallowed | Per-item errors terminate the output |
| `AutoRefreshOnObservable` | Per-item reevaluator errors swallowed | Per-item errors terminate the output |
| `FilterOnObservable` | Per-item filter errors swallowed | Per-item errors terminate the output |

XML docs on each operator now describe the new behavior and call out the prior swallowing.

If an application was relying on the old swallowing behavior, the workaround is to defensively `.Catch(...)` inside the per-item observable.


src/DynamicData/Cache/Internal/MergeMany.cs
```diff
@@ -30,32 +25,21 @@ public MergeMany(IObservable<IChangeSet<TObject, TKey>> source, Func<TObject, IO
...
-    }
+                        context.Track(change.Key, _observableSelector(change.Current, change.Key));
+                    }
+                    else if (change.Reason is ChangeReason.Remove)
+                    {
+                        context.Untrack(change.Key);
+                    }
+                }
+            },
+            onInner: (value, _, emitter) => emitter.OnNext(value));
```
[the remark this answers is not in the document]
dwcullop: Updated XML docs on MergeMany (and the four operators that route through it: WhenAnyPropertyChanged, WhenPropertyChanged, WhenValueChanged, AutoRefreshOnObservable) to describe the new error-propagation behavior. PR description now has a 'Behavioral change' section calling this out. Pushed in 9e45d5c1.

src/DynamicData/Cache/Internal/IntObservableCacheEx.OrchestrateManyChangeSets.cs
```diff
@@ -0,0 +1,192 @@
...
+                _tracker.RefreshItems(current.Value.Cache.Keys);
+            }
+        }
+
+        private void SubscribeChild(TSource item, TKey key)
+        {
+            var entry = new ChangeSetCache<TDest, TDestKey>(Context.Serialize(_changeSetSelector(item, key).IgnoreSameReferenceUpdate()));
+            _cache.AddOrUpdate(entry, key);
+            Context.Track(key, entry.Source);
+        }
```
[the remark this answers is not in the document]
dwcullop: Good catch. Dropped the explicit Context.Serialize(...) on both call sites in IntObservableCacheEx.OrchestrateManyChangeSets.cs (cache + list) and the same pattern in TransformManyAsync.cs. Track already wraps with SynchronizeSafe, so there's now a single sub-queue per key. Pushed in 9e45d5c1.

src/DynamicData.Tests/Internal/TransformManyAsyncOrchestratorFixture.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        // The deferred transform subscribes when we drive the tracked observable; subscribe now and
+        // settle the asynchronous defer to surface the error to the handler.
+        var trackedObs = context.Tracked[1];
+        using var sub = trackedObs.Subscribe();
+        await Task.Delay(20);
+
+        capturedError.Should().NotBeNull();
+        capturedError!.Exception!.Message.Should().Be("transformer-broke");
+    }
```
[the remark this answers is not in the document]
dwcullop: Replaced Task.Delay with await trackedObs.DefaultIfEmpty().LastOrDefaultAsync() in commit 0c69976b. Deterministic completion.

src/DynamicData.Tests/Internal/OrchestrateFixture.cs
```diff
@@ -0,0 +1,576 @@
...
+        lock (orchestrator.ChildCalls)
+        {
+            orchestrator.ChildCalls.Count.Should().Be(totalEmissions,
+                "every concurrent inner emission must reach OnInner; reentrant drain during emit must not drop items");
+        }
+
+        foreach (var subj in innerSubjects.Values)
+        {
+            subj.Dispose();
+        }
```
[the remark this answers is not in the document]
dwcullop: Replaced Task.Delay with SpinWait.SpinUntil(predicate, TimeSpan.FromSeconds(5)) in commit 0c69976b. Deterministic count-based wait with bounded timeout.

src/DynamicData/Cache/Internal/ICacheOrchestrator.cs
```diff
@@ -0,0 +1,52 @@
...
+internal interface ICacheOrchestrator<TSource, TKey, TInner, TResult>
+    where TSource : notnull
+    where TKey : notnull
+    where TInner : notnull
+{
+    /// <summary>
+    /// Invoked for each source changeset.
+    /// </summary>
+    /// <param name="changes">The source changeset.</param>
+    void OnSourceChangeSet(IChangeSet<TSource, TKey> changes);
```
JakenVeina: "Event handler" methods should be named as verbs. I look at this and I say "On source changeset.... what? What happened that I am supposed to respond to?" A better name in my mind would be `OnSourceNext()`, as it clarifies very specifically what the method is responding to, with just the name alone: it's `source.OnNext()`. `OnSourceChangeSetReceived()` would fit as well.

src/DynamicData/Cache/Internal/ICacheOrchestrator.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// <summary>
+    /// Invoked for each value emitted by a tracked inner observable, paired with its source key.
+    /// </summary>
+    /// <param name="value">The value emitted by the inner observable.</param>
+    /// <param name="key">The source key the inner observable was registered against.</param>
+    void OnInner(TInner value, TKey key);
```
JakenVeina: Similarly as above, "On inner what?" What "inner" thing happened?

Also, similarly as above, I think using already-established terms is better. `OnItemSourceNext()` would be my pick. "Item" is well-understood to mean items within the source collection, and "item source" then implies "a source that came from an item". Plus "Next" to refer to `OnNext()`, obviously. `OnInnerNext()` would work as well, or `OnInnerSourceNext()`, or `OnChildNext()` or `OnChildSourceNext()`.

src/DynamicData/Cache/Internal/ICacheOrchestrator.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <c>OnCompleted</c> (the source changeset and every tracked inner observable have all
+    /// completed). Implementations holding deferred state (timer-armed buffers, debounced batches)
+    /// should flush synchronously when this is <see langword="true"/>.
+    /// </param>
+    /// <param name="wasReentrant">
+    /// <see langword="true"/> when a reentrant drain occurred during the prior delivery cycle
+    /// (an orchestrator emit triggered same-thread re-entry into the drain loop). Most
+    /// orchestrators ignore this.
+    /// </param>
+    void OnDrainComplete(bool isFinal, bool wasReentrant);
```
JakenVeina, line 51: Is there a scenario in the other operators that needs this, cause it still doesn't make sense to me. `.SynchronizeSafe()` already guarantees synchronization of all the notifications coming out of it. Seems to me the only thing this accomplishes is a slight amount of potentially-free batching, which seems not worth the extra complexity to me, because it would only be "dumb" batching.

For example:

`.AutoRefresh()` desires to do "batching" in that any events that are received from a child stream during the `.Subscribe()` call to that stream should be ignored. The operator definitely DOESN'T want to emit a `Refresh` for an item before its `Add`, and after its `Add` a `Refresh` is redundant. `OnDrainCompleted()` doesn't help accomplish this at all. If `.AutoRefresh()` were to use `OnDrainCompleted()` to accomplish this, it'd have to keep track of a lot more potential temporary state (I.E. which refreshes came from source, and need to be emitted, and which refreshes came from children, and need to be ignored), THEN it'd have to iterate over that state a second time, to generate the final changeset. If it just does its own re-entrancy tracking, it can skip that entire process and just put changes directly into the result changeset, or not, as they come in.

Seems to me pretty much EVERY other operator that would use this system would want the same thing: know about re-entrancy WHEN the incoming notification arrives, to decide how to process it, not just queue it up to get processed later.

src/DynamicData/Cache/Internal/ICacheOrchestratorContext.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// the same serialization as source and inner notifications.
+    /// </summary>
+    /// <remarks>
+    /// Unlike <see cref="Track"/>, a Serialize-wrapped subscription does NOT participate in
+    /// completion accounting: downstream can complete while it is still active.
+    /// </remarks>
+    /// <typeparam name="T">Value type of the observable being serialized.</typeparam>
+    /// <param name="observable">The observable to wrap.</param>
+    /// <returns>A new observable whose <c>OnNext</c> delivery runs under the orchestrator's queue lock.</returns>
+    IObservable<T> Serialize<T>(IObservable<T> observable);
```
JakenVeina, line 47: Is there any disadvantage to just exposing the entire `DrainQueue`? " "Open for extension, but closed for modification" and all that.

src/DynamicData/Cache/Internal/CacheOrchestration.cs
```diff
@@ -0,0 +1,177 @@
...
+            container.Disposable = observable!
+                .SynchronizeSafe(_queue)
+                .Finally(DecrementSubscriptionCount)
+                .SubscribeSafe(
+                    onNext: value => OnInner(value, key),
+                    onError: _emitter.OnError,
+                    onCompleted: () => _innerSubscriptions.Remove(key));
+        }
+
+        public void Untrack(TKey key) => _innerSubscriptions.Remove(key);
```
JakenVeina, line 118: You've forgotten to unsubscribe.

src/DynamicData/Cache/Internal/CacheOrchestration.cs
```diff
@@ -0,0 +1,177 @@
...
+    where TInner : notnull
+    where TOrch : ICacheOrchestrator<TSource, TKey, TInner, TResult>
+{
+    public IObservable<TResult> Run() => Observable.Create<TResult>(observer => new OrchestratorContext(source, observer, factory));
+
+    private sealed class OrchestratorContext : ICacheOrchestratorContext<TKey, TInner>, IDisposable
+    {
+        private readonly KeyedDisposable<TKey> _innerSubscriptions = new();
+        private readonly SingleAssignmentDisposable _sourceSubscription = new();
+        private readonly SharedDeliveryQueue _queue;
```
JakenVeina: `_upstreamQueue` would be a tad more descriptive.

src/DynamicData/Cache/Internal/CacheOrchestration.cs
```diff
@@ -0,0 +1,177 @@
...
+        private readonly DeliverySubQueue<TResult> _emitter;
```
JakenVeina: Similarly, `_innerQueue` or `_itemSourceQueue` would work better, or whatever matches the naming used for the child observables.

src/DynamicData/Cache/Internal/CacheOrchestration.cs
```diff
@@ -0,0 +1,177 @@
...
+            _queue = new SharedDeliveryQueue(onDrainComplete: OnDrainComplete);
+
+            try
+            {
+                // Emitter sub-queue is allocated first so it drains last (LIFO); source-triggered
+                // sync inner emissions land on later-indexed queues and deliver before it.
+                _emitter = _queue.CreateQueue(observer);
+
+                _orchestrator = factory(this, _emitter);
+                Debug.Assert(_orchestrator is not null, "Factory must not return null");
```
JakenVeina: This assert seems pointless if these interfaces are only `internal` and we already have nullable-type checking enabled, to prevent this from happening.

src/DynamicData/Cache/Internal/AutoRefresh.cs
```diff
@@ -1,40 +1,135 @@
...
+        private readonly SerialDisposable _timerSubscription = new();
+
+        public void Dispose() => _timerSubscription.Dispose();
+
+        public override void OnSourceChangeSet(IChangeSet<TObject, TKey> changes)
+        {
+            base.OnSourceChangeSet(changes);
+            if (changes.Count > 0)
+            {
+                Emitter.OnNext(changes);
```
JakenVeina, line 53: Wait.... what? We're using the same drain queue for both upstream and downstream?

src/DynamicData/Cache/Internal/CacheOrchestratorBase.cs
```diff
@@ -0,0 +1,71 @@
...
+{
+    protected ICacheOrchestratorContext<TKey, TInner> Context => context;
+
+    protected IObserver<TResult> Emitter => emitter;
+
+    public virtual void OnSourceChangeSet(IChangeSet<TSource, TKey> changes)
+    {
+        foreach (var change in changes.ToConcreteType())
+        {
+            switch (change.Reason)
```
JakenVeina, line 31: You're missing some important logic here. After each change, this loop needs to check for an error condition, and exit if there is one. Otherwise, it will continue to create and subscribe to child streams, and since the teardown logic for the operator will have already run, they'll never get disposed.

This should actually get baked into `ICacheOrchestrator` directly, since this is logic that EVERY implementer will need, and it's QUITE bad if any of them forget, or get it wrong. I.E. `ICacheOrchestrator` needs to expose a handler per-change, not per-changeset. It should then expose an "After" handler for the whole changeset, for operators that might want to post-process or forward the original changeset.

src/DynamicData/Cache/Internal/ChangeSetCache.cs
```diff
@@ -2,23 +2,21 @@
...
 
 /// <summary>
-/// Wraps an Observable ChangeSet while maintaining a copy of the aggregated changes.
+/// Holds an observable changeset alongside an aggregated mirror cache.
+/// <see cref="Process"/> applies a changeset to <see cref="Cache"/>.
 /// </summary>
 /// <typeparam name="TObject">ChangeSet Object Type.</typeparam>
 /// <typeparam name="TKey">ChangeSet Key Type.</typeparam>
-internal sealed class ChangeSetCache<TObject, TKey>
+internal sealed class ChangeSetCache<TObject, TKey>(IObservable<IChangeSet<TObject, TKey>> source)
```
JakenVeina: What's the purpose of this class? Its name certainly doesn't tell me, the name suggests to me that we're building a cache of individual changesets. All this seems to be is a tuple of `Cache` and `Source` with a pointless alias for `.Clone()` on it.

src/DynamicData/Cache/Internal/IntObservableCacheEx.Orchestrate.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// </summary>
+    /// <typeparam name="TSource">Type of items in the source changeset.</typeparam>
+    /// <typeparam name="TKey">Type of the source changeset key.</typeparam>
+    /// <typeparam name="TInner">Type of values emitted by the per-key inner observables.</typeparam>
+    /// <typeparam name="TResult">Type delivered downstream.</typeparam>
+    /// <typeparam name="TOrch">Concrete orchestrator type returned by the factory. Generic-typed so dispatch sites devirtualize.</typeparam>
+    /// <param name="source">The keyed source changeset stream.</param>
+    /// <param name="factory">Builds the per-subscription orchestrator from its runtime context and emitter.</param>
+    /// <returns>An observable that orchestrates source and inner activity into a single result stream.</returns>
+    public static IObservable<TResult> Orchestrate<TSource, TKey, TInner, TResult, TOrch>(
```
JakenVeina: I like either `.OrchestrateSubscriptions()` or `.SubscribeManyBase()`. Either way, I think the clarification that this operator is all about dealing with subscriptions is valuable.

JakenVeina (changes requested):
Overall, I'm still pretty vehemently against the dual-interface implementation. To re-summarize:
* It's more noisy (code-wise) than just an abstract base class. You've got, what, 5 files? Versus just one for a base class.
* The biggest downside of an abstract base class is all the virtual method calls, which only occur for the abstract/virtual methods on the class. With two interfaces EVERY call is virtual. The performance is WORSE.
* It's not an established pattern that newcomers to the codebase are going to recognize.

src/DynamicData/Cache/Internal/ICacheOrchestrator.cs
dwcullop: Fair question, and I think the name is obscuring the answer. This hook is not really about reentrancy, it is the batch boundary. Three concrete things depend on it today.

**1. Coalescing into a single changeset.** `ChangeSetOrchestrator.OnDrainComplete` calls `_cache.CaptureChanges()` and emits once. Move that to per-notification and every inner value becomes its own changeset. `AutoRefreshFixture.Base.ChangeSetBufferIsGiven_PropertyChangedNotificationsAreBufferedOnScheduler` pins the current behaviour: two property change notifications raised in separate buffer windows arrive as one changeset with `Count == 2` and `Refreshes == 2`. Emitting per notification turns that into two changesets of one, and the `Count().Should().Be(1, "a buffer window expired")` assertion no longer holds. The same shape drives `OrchestrateManyChangeSets` (both orchestrators) and `TransformManyAsync`, which all call `_tracker.EmitChanges(Emitter)` at the boundary.

**2. `isFinal` cannot be derived at arrival time.** It comes from `_subscriptionCounter == 0` once the drain settles. AutoRefresh needs it to flush pending refreshes on the last drain, because stream termination cancels the buffer timer and anything still pending would never be delivered. No per-notification hook can know it is looking at the final one.

**3. The de-dup window needs a place to reset.** AutoRefresh uses `_sourceTouched` to suppress a refresh for any key the source itself already touched in the same batch. `OnDrainComplete` is where that set clears. Without a boundary there is nothing to scope it to.

On `wasReentrant` you have a point. I went through the implementations and none of them actually read it, so I am happy to drop that parameter and keep `isFinal` if you agree.

src/DynamicData/Cache/Internal/ICacheOrchestratorContext.cs
dwcullop: I would like to keep this surface narrow. Leaving the queue implicit means the orchestrators never take a direct dependency on it, so if we want to change or remove the queue later the blast radius stays in this file rather than spreading across every implementation.

Open for extension is definitely the goal, I would just rather the extension points be the orchestrator hooks themselves than the plumbing underneath them.

src/DynamicData/Cache/Internal/ChangeSetCache.cs
dwcullop: Fair, the name was doing it no favours. Renamed it to `ChangeSetMirror`.

The purpose is to pair a child changeset stream with a live mirror of what that child currently holds. The merge trackers keep one per child, and `LookupBestValue` / `UpdateToBestValue` walk them to resolve a key that more than one child publishes. Without the mirror there is nowhere to answer "what does child N hold right now?" when a key is removed from one child and has to fall back to the value in another.

I have put that explanation into the summary so it is not tribal knowledge. On `Process`, you are right that it is thin. I have kept it for now because it makes the call sites read as "apply this changeset to the mirror", but I am happy to drop it and let them call `Cache.Clone` directly if you would prefer.

src/DynamicData/Cache/Internal/AutoRefresh.cs
dwcullop: Not quite in the sense it reads. There is one `SharedDeliveryQueue`, and the emitter is a sub-queue created from it by `_queue.CreateQueue(observer)`. Upstream work (the source subscription, the per-key item sources, and this timer) goes through the gate, while downstream delivery goes through the sub-queue.

Sharing the parent buys two things. First, a single drain thread, so upstream processing cannot race downstream delivery while both are touching orchestrator state such as the `ChangeAwareCache` or `_pendingRefreshes`. Second, ordering: the emitter sub-queue is allocated first so it drains last, which is what makes "process all upstream work, then emit" hold. That ordering is what `OnDrainComplete` keys off, and two independent queues could not express the relationship.

For the timer specifically, routing it through the gate is what stops `FlushPending` running concurrently with a source notification that is mutating `_pendingRefreshes`.

src/DynamicData/Cache/Internal/ICacheOrchestrator.cs
dwcullop: Following up on the `wasReentrant` half of this: I am going to keep the parameter after all. Nothing reads it today, but it is cheap to carry and it is the kind of thing an operator may well want once we start adding more of them, so I would rather it be there than have to thread it back through later.

src/DynamicData/Cache/Internal/CacheOrchestration.cs
dwcullop: I think this anchor has drifted, because the line it is sitting on now is `Untrack`:

```csharp
public void Untrack(TKey key) => _innerSubscriptions.Remove(key);
```

`_innerSubscriptions` is a `KeyedDisposable<TKey>`, and its `Remove` pulls the entry out of the dictionary under the gate and then calls `toDispose.Dispose()` on the way out, so the unsubscribe does happen. The counter stays balanced too, since the subscription carries `Finally(DecrementSubscriptionCount)`.

If you were looking at something else when you left this, point me at it and I will take another run at it.

src/DynamicData/Cache/Internal/CacheOrchestratorBase.cs
dwcullop: I chased the leak and I do not think it can happen, though it took me a minute to convince myself.

If the emitter faults partway through a changeset and teardown runs, the remaining iterations still call `Context.Track`. But `Track` does `_innerSubscriptions.Add(key, new SingleAssignmentDisposable())`, and once `KeyedDisposable` is disposed its `Add` takes the else branch and disposes the incoming container immediately rather than storing it. `Track` then assigns the real subscription into that already-disposed `SingleAssignmentDisposable`, which disposes it on assignment. So the container always gets disposed and may or may not get assigned, and either way there is nothing left subscribed. `Finally(DecrementSubscriptionCount)` keeps the counter balanced.

Wasted work on a path that is already failing, but not a leak. If you can see a route where the container survives, I would rather find out now.

Your second point is the more interesting one and I agree with it. Making `ICacheOrchestrator` expose a per-change handler plus an "after the changeset" hook is a better shape than handing every implementer a raw changeset and trusting them to loop over it correctly. `CacheOrchestratorBase` already offers that shape and every orchestrator in the PR uses it, so the interface is arguably just describing the wrong contract. I would rather do that as a follow-up than grow this PR further, but I am happy to be talked into doing it here.

src/DynamicData/Cache/Internal/ICacheOrchestrator.cs
dwcullop: Agreed, OnSourceNext is clearer about what it is responding to. Renamed.

src/DynamicData/Cache/Internal/ICacheOrchestrator.cs
dwcullop: Gone with OnItemSourceNext. The "item source" framing reads well and it lines up with the queue naming below.

src/DynamicData/Cache/Internal/CacheOrchestration.cs
dwcullop: Renamed to `_upstreamQueue`.

src/DynamicData/Cache/Internal/CacheOrchestration.cs
dwcullop: Renamed to `_innerQueue`. Worth flagging that this field is the sub-queue that delivers downstream to the observer rather than a queue for the item sources, so if the name still reads wrong to you once you have seen the reply on AutoRefresh.cs, I am open to another pass at it.

src/DynamicData/Cache/Internal/CacheOrchestration.cs
dwcullop: Fair, nullable analysis already covers that one. Removed. I have left the other two asserts in place since they guard runtime invariants rather than type invariants.

src/DynamicData/Cache/Internal/IntObservableCacheEx.Orchestrate.cs
dwcullop: Gone with `OrchestrateSubscriptions`. You are right that being explicit about the subscription management is worth the extra characters.

src/DynamicData/Cache/Internal/ICacheOrchestrator.cs
JakenVeina: TL;DR: I'm not sure what form the final DQ implementation will take, but I feel pretty strongly that it won't contain the `OnDrainComplete` API.

In my book, This level of de-duping is not the responsibility of the DQ, and not really desirable anyway.

We had a discussion a long time ago (I think when the re-entrancy thing came up for `.MergeManyChangeSets()`) of whether we wanted to implement, let's call it "aggressive batching". I.E. operators should generally publish a single notification "per-frame" of received notifications, where a frame equates roughly to a thread acquiring a lock. If that same thread publishes additional notifications during that lock window (I.E. via nested subscriptions), that's the same frame. The "aggressive batching" approach would have us also batch up additional frames (I.E. notifications that came in during the lock window, but on a different thread) by holding onto the batch if we detect that another thread is waiting by the time we're done, and letting that thread build on it.

We elected to not go for this, and I think that was the right call. Blurring the lines between changesets might have a small performance impact, but it would be VERY minimal, as we're not actually generating or processing fewer changes, just putting the same changes together into fewer collections. It also makes debugging more difficult, I think, when it's less obvious to trace where changes in a changeset came from, as they propagated from upstream.

Yet that's what the "de-duping" here is. It's merging together multiple upstream "frames" into a single downstream "frame". `.AutoRefresh()` I suppose does represent a unique exception to the above, as we CAN safely reduce the total number of changes, but this is ONLY a thing that can be done for `Refresh` changes. Really, that's an argument that `.AutoRefresh()` is SPECIAL and SHOULDN'T use the shared orchestrator logic.

And regardless of the merits of "aggressive batching", I see it as DEFINITELY not the responsibility of the DQ. The DQ's goal and responsibility is synchronization of notifications (without the chance of deadlocking), as made clear by the fact that the operator you wrote for it is `.SynchronizeSafe()`. It's made even clearer by the `AsyncLock` class you pointed out, which accomplishes that goal exactly, with a DEAD simple API.

src/DynamicData/Cache/Internal/CacheOrchestration.cs
JakenVeina: No drift, I just missed the fact that the collection self-disposes items, on removal. I don't really like it, TBH. It obscures the disposal action that's taking place, and it seems like a whole lot of logic for something that'd just be a few additional lines in the consumer. But there's nothing overtly wrong with it.

src/DynamicData/Cache/Internal/AutoRefresh.cs
JakenVeina: I've genuinely got no idea what this comment was about. There's no drain queue in sight here across the entire file. I guess there WAS some kind of permalink drift? Which is nonsense, cause that defeats the definition of a "permalink". Did you end up ripping the queue references out inot `CacheOrchestration`?

> the emitter sub-queue is allocated first so it drains last, which is what makes "process all upstream work, then emit" hold

If that's important, then we definitely need to make it explicit, which this doesn't. The fact that something gets processed first just because it happens to be allocated first, compared to other things, is an implementation detail.

> two independent queues could not express the relationship

Having a single queue doesn't express the relationship either.

src/DynamicData/Cache/Internal/CacheOrchestratorBase.cs
JakenVeina: As above, I missed the nature of `_innerSubscriptions` as an auto-disposing collection. In that context, I think you're right, there's no actual leak here, just redundant work on a failure path.

dwcullop:
> I feel pretty strongly that it won't contain the `OnDrainComplete` API. In my book, this level of de-duping is not the responsibility of the DQ

Agreed on the layering, and I have taken a run at it in #1163. `SharedDeliveryQueue` no longer has the callback at all: `CacheParentSubscription` tracks its own delivery frame and emits when the depth returns to zero, so the queue is back to only serializing.

Worth flagging what I found on the way, because it changes the shape of the argument a bit.

It is load-bearing. `EmitChanges` is abstract and that callback was its only caller, so removing it without a replacement means the six operators built on `CacheParentSubscription` never emit. I tried the obvious alternative of emitting per notification and 34 tests fail, all on changeset granularity. `ClearingParentEmitsSingleChangeSet` gives 15 where it expects 2. `OrderOfChangesIsPreserved` gives 11 where it expects 2, and that one is more than a count: one `Edit` doing a `Clear()` then an `AddOrUpdate()` has to arrive as a single changeset, otherwise a consumer observes the collection emptied, which is a state that never existed upstream.

I also think it is not quite the thing you rejected. By your own description a frame is one thread's lock window including anything nested subscriptions publish inside it, and you said operators generally should publish once per frame. That is what the callback does in the common case. Where you are right is that it goes further: `DrainPending` drains whatever is queued, so work another thread enqueued mid-drain lands in the same emission. That is the cross-thread batching you said was rejected, and it should not be there.

The frame-depth version in #1163 keeps the first and drops the second. A second thread's work now gets its own frame.

Separately, #1162 fixes the queue delivering out of receipt order. Those two are independent, but the second one turns out to have been hiding the first: with everything in a drain cycle folded into one changeset, the wrong order mostly washed out before anything downstream saw it. Once emission is per frame, delivery order becomes changeset order.

That also clears up the `AutoRefresh.cs` comment you could not place. The queue references had moved into `CacheOrchestration`, so the comment was stranded, and I will fix it here. On the related point, you are right that "allocated first so it drains last" should not be load bearing. #1162 removes the mechanism that made it work at all, since drain order becomes arrival order rather than index order, so this branch needs to state that relationship explicitly instead of relying on it. I will sort that out once those settle, as they would otherwise collide silently.

PR #1118, opened 2026-06-23
Feat: sum benchmarks

giard-alexandre:
Adds some benchmarks to the Sum aggregate operator.

This is basically a preamble to the full rewrite of the Sum operator, as described in https://github.com/reactivemarbles/DynamicData/discussions/1031#discussioncomment-15777487

### DIFF
--- a/src/DynamicData.Benchmarks/Cache/Sum_Cache.cs
+++ b/src/DynamicData.Benchmarks/Cache/Sum_Cache.cs
@@ -12,18 +12,22 @@ namespace DynamicData.Benchmarks.Cache;
 [MarkdownExporterAttribute.GitHub]
 public class Sum_Cache
 {
-    private readonly IReadOnlyList<IChangeSet<Item, int>> _addChangeSets;
-    private readonly IReadOnlyList<IChangeSet<Item, int>> _replaceChangeSets;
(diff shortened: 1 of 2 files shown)


src/DynamicData.Benchmarks/Cache/Sum_Cache.cs
```diff
@@ -0,0 +1,108 @@
...
+            var replacement = new Item()
+            {
+                Id = id,
+                Value = id * 2
+            };
+            items[id] = replacement;
+            source.AddOrUpdate(replacement, key: id);
+            replaceChangeSets.Add(source.CaptureChanges());
+        }
+        _replaceChangeSets = replaceChangeSets;
```
JakenVeina, line 60: You've generated an invalid sequence of changes for all of these, except for `_addChangeSets`. `source.CaptureChanges()` also clears out the changes buffer, so when you get around to sending `_replaceChangeSets` to your operator, what you've got is nothing but `Replace` changes that refer to items that never had an `Add`. Apparently, the existing implementation of `.Sum()` doesn't care, but your new one might.

JakenVeina (changes requested):

JakenVeina (approved):

giard-alexandre:
That was quick! Thanks @JakenVeina !

Issue #1119, opened 2026-06-24
[Bug]: RemoveKey with AutoRefresh and Filter throws


jcummings2:
### Describe the bug 🐞

Upgrading to 9.4.31 caused code with the pattern mentioned in the "Steps to Reproduce" section below to fail in some situations.

It can result in the following Exception:

> System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')

The bug was introduced with the rewrite of the static list filter logic in #1063. The issue ultimately stems from what seems to a be long standing feature/bug of `RemoveKeyEnumerator` that sets the `CurrentIndex` to -1 when a list is not passed into the constructor (or technically when the item isn't found when a list is passed but no callers actually pass a list). Pull request #1120  with unit tests that demonstrate the problem and a fix.

### Step to reproduce

```c#
        ReadOnlyObservableCollection<Person> collection;
        using var source = new SourceCache<Person, string>(p => p.Key);
        using var subscription = source.Connect()
                .AutoRefresh(x => x.Age)
                .RemoveKey()
                .Filter(x => x.Age > 50)
                .Bind(out collection)
                .Subscribe();
```

### Reproduction repository

https://github.com/jcummings2/DynamicData

### Expected behavior

List should be properly updated with filtered list of items from cache.

### DynamicData Version

9.4.31

### Additional information ℹ️

Worked fine in 9.4.1

jcummings2:
As noted in the PR, moving the `RemoveKey()` call to just before `Bind()` avoids the exception. This detail means code following this pattern works:

```c#
        ReadOnlyObservableCollection<Person> collection;
        using var source = new SourceCache<Person, string>(p => p.Key);
        using var subscription = source.Connect()
                .AutoRefresh(x => x.Age)
                .Filter(x => x.Age > 50)
                .RemoveKey()
                .Bind(out collection)
                .Subscribe();
```

JakenVeina:
Arguably, this should be the recommended solution. If I had my way, we'd deprecate and remove list-domain filter operators entirely. They're a massive pain to implement properly and/or performantly.

But as I mentioned over in the PR, there's probably quite a few scenarios littered around where list indexes can get stripped out of changesets. If we want to get rid of that behavior, thay really needs to be its own effort. I'm fine with this targeted fix until then.

JakenVeina:
[...]

PR #1120, opened 2026-06-24
Fix Exception in list static filter index assumptions

jcummings2:
See #1119


src/DynamicData.Tests/Cache/RemoveKeyFixture.cs
```diff
@@ -0,0 +1,119 @@
...
+    public RemoveKeyFixture()
+    {
+        _source = new SourceCache<Person, string>(p => p.Key);
+        _cleanup.Add(_source);
+    }
+
+    public void Dispose() => _cleanup.Dispose();
+
+    [Fact]
+    public void Add()
```
JakenVeina: Test names should be a little more descriptive. Ideally, the pattern is "UnitUnderTest_Scenario_ExpectedOutcome", but that's certainly not followed everywhere, so at minimum, we just want to see some description of either "Scenario" or "ExpectedOutcome" in the name. Just saying "Add", "Refresh", etc. doesn't tell me anything about what we're intending to test here. Could be as simple as "KeyIsRemoved".

src/DynamicData.Tests/Cache/RemoveKeyFixture.cs
```diff
@@ -0,0 +1,119 @@
...
+    public void Filter()
+    {
+        var people = _generator.Take(100).ToArray();
+        var average = people.Average(x => x.Age);
+
+        ReadOnlyObservableCollection<Person> collection;
+        _cleanup.Add(
+            _source.Connect()
+                .RemoveKey()
+                .Filter(x => x.Age < average)
```
JakenVeina, line 63: In the test fixture for "RemoveKey", the only thing that should be getting tested is `.RemoveKey()`, What you're REALLY trying to test for here is behavior in `.Filter()`, specifically, `.Filter()`'s ability to support missing indexes. That belongs in the appropriate "Filter" fixture.

src/DynamicData/List/Internal/Filter.Static.cs
```diff
@@ -212,13 +211,24 @@ public static IObservable<IChangeSet<T>> Create(
                                 {
                                     var isIncluded = predicate.Invoke(change.Item.Current);
 
-                                    var itemState = upstreamItemsStates[change.Item.CurrentIndex];
-                                    upstreamItemsStates[change.Item.CurrentIndex] = (
+                                    var currentIndex = change.Item.CurrentIndex;
+                                    // A Replace might have a negative CurrentIndex from a Refresh in RemoveKeyEnumerator
+                                    if (currentIndex < 0)
+                                    {
+                                        // this code mimics the previous logic
```
JakenVeina: This comment doesn't belong here. Useful for the PR, sure, but just pulling up and reading this comment, as-is begs the question "What previous logic?". The comment above clarifying that `CurrentIndex` can be negative does plenty to tell us what's going on here: we don't have the index, so we've got to look it up.

src/DynamicData/List/Internal/Filter.Static.cs
```diff
@@ -212,13 +211,24 @@ public static IObservable<IChangeSet<T>> Create(
                                 {
                                     var isIncluded = predicate.Invoke(change.Item.Current);
 
-                                    var itemState = upstreamItemsStates[change.Item.CurrentIndex];
-                                    upstreamItemsStates[change.Item.CurrentIndex] = (
+                                    var currentIndex = change.Item.CurrentIndex;
+                                    // A Replace might have a negative CurrentIndex from a Refresh in RemoveKeyEnumerator
```
JakenVeina, line 215: Probably not even the only scenario. Boy do I hate working inside the old "list" code.

JakenVeina (changes requested):

src/DynamicData.Tests/Cache/RemoveKeyFixture.cs
jcummings2: What I was actually trying to demonstrate was what series of calls would lead to the error. This test is really just a step indicating it wasn't the problem. `Filter()` itself is tested more thoroughly in all the `Filter*` classes.

src/DynamicData.Tests/Cache/RemoveKeyFixture.cs
jcummings2: I couldn't discern the patten used in this code base so I went very basic. I can improve the name(s).  I'm not sure what you would want for "UnitUnderTest" in the pattern.

src/DynamicData.Tests/Cache/RemoveKeyFixture.cs
jcummings2: Are you trying to say you want this test removed?

jcummings2:
I updated test names and removed the extraneous comment per your feedback. Let me know if those names need further tweaking.

I also added another test that moves the `RemoveKey()` call to just before the bind. This new test passes even with the old code as it turns out.

src/DynamicData.Tests/Cache/RemoveKeyFixture.cs
JakenVeina: No, I'm saying move it over to the Filter fixture. `.Filter()` is the operator here that has defective behavior. If other operators are capable of generating changesets with missing list indexes, `.Filter()` needs to be able to process them. At least, that's how you've written the fix. With that being a behavior that `.Filter()` needs, there should ve a test in the Filter fixture that feeds missing indexes into the operator, and verifies that it works.

jcummings2: There are about 7 files making up the Filter fixture. I moved it to `FilterFixture.Static.cs`

JakenVeina: Yup. The "static" version of `.Filter()` (that is, the one that does static filtering based on a predicate that doesn't change and doesn't take any parameters except the filter subject) is the one that file corresponds with.

JakenVeina (approved):

PR #1127, opened 2026-07-02
Restore OnError routing for NPC-sourced WhenPropertyChanged emissions

dwcullop:
## Summary

PR #1111 fixed a TOCTOU race in `WhenPropertyChanged` / `WhenValueChanged`, but a follow-up commit in that PR stopped routing downstream exceptions through `OnError`. Because these emissions originate from an `INotifyPropertyChanged` callback, an unhandled `OnNext` throw escaped **raw out of the property setter** that raised `PropertyChanged`: setting a property could throw because some unrelated observer misbehaved.

This restores the error routing on both the shallow (`x => x.Prop`) and deep-chain (`x => x.A.B.Prop`) paths.

## What changed

- `SinglePropertySubscription.EmitCurrent` and `DeepChainSubscription.ProcessSignal` again wrap the accessor read, chain walk, and downstream `OnNext` in try/catch, routing any exception through `OnError`. A subscriber that supplied an error handler now observes the failure instead of it escaping into the setter.
- Per @JakenVeina's review point on #1111, we deliberately do **not** swallow a throw from `OnError` itself. An unhandled downstream error (or a rethrowing error handler) still propagates, matching `Subject<T>.OnNext` semantics. This is the middle ground between the original swallow-everything behavior and the raw-escape behavior that shipped.

## Behavior

| Scenario | Current main | After |
| --- | --- | --- |
| Subscriber `OnNext` throws, has error handler | escapes raw to setter, handler bypassed | routed to `OnError`, setter safe |
| Subscriber `OnNext` throws, no error handler | escapes to setter | routed to `OnError` (no-op after detach), setter safe |
| Accessor throws, error handler rethrows | propagates to setter | propagates to setter (unchanged) |

## Tests

`WhenPropertyChangedBehaviorFixture` gains 4 tests (shallow + deep) covering the `OnNext`-routing restore (RED on current main) and the non-swallow decision. Full `Binding` suite: 156 passed, 2 pre-existing skips.

### DIFF
--- a/src/DynamicData.Tests/Binding/WhenPropertyChangedBehaviorFixture.cs
+++ b/src/DynamicData.Tests/Binding/WhenPropertyChangedBehaviorFixture.cs
@@ -154,6 +154,109 @@ public sealed class WhenPropertyChangedBehaviorFixture
         emissions.Should().Equal(new[] { 10, 20, 30 }, "leaf event on detached subtree is ignored");
     }
 
+    [Fact]
+    public void Shallow_ObserverThrowsInOnNext_RoutedToOnError_DoesNotEscapeSetter()
+    {
+        // Emissions originate from an INotifyPropertyChanged callback. A subscriber whose OnNext
+        // throws must not surface that exception raw out of the property setter; it is routed
+        // through OnError so a subscriber that supplied an error handler can observe it, and the
+        // setter that raised PropertyChanged stays safe.
+        var model = new TestModel { Value = 10 };
+
+        using var sub = model.WhenPropertyChanged(m => m.Value, notifyOnInitialValue: false)
+            .Subscribe(_ => throw new InvalidOperationException("boom"));
+
+        var setter = () => model.Value = 20;
+        var laterSetter = () => model.Value = 30;
+
+        setter.Should().NotThrow("a PropertyChanged-sourced OnNext throw is routed to OnError, not escaped to the setter");
+        laterSetter.Should().NotThrow("the terminated subscription must not resurface the exception on a later setter");
+    }
+
+    [Fact]
+    public void Shallow_AccessorThrows_ErrorHandlerRethrows_PropagatesToSetter()
+    {
+        // We route accessor/OnNext failures to OnError but do NOT swallow a throw from OnError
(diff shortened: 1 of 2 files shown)


JakenVeina:
This will only make the described scenario worse. Instead of getting thrown out of the property setter, the exception will now infinite loop through the stream chain, and hang the entire thread. Throwing exceptions upstream is something that only happens **when there is no downstream exception handler**, or when there is a defect within an operator.

If a consumer can't tolerate exceptions bubbling out of a property setter, they need to introduce proper handling for it, either downstream, or just below the `.WhenPropertyChanged()` subscription. We can introduce an overload of `.WhenPropertyChanged()` that allows an error handler to be supplied, and/or we can add an operator for catching downstream errors. Maybe even suggest such an operator to dotnet/reactive, I don't think there's an existing one.

dwcullop:
Closing this. After tracing the delivery path and probing the actual behavior, `main` is already correct from an Rx standpoint and this PR's change is not.

The contract distinction:
- Producer-side faults (the accessor `x => x.Prop`, the chain walk, notifier factories) route to `OnError` -- which is what `main` already does, and what a `Select` projection would do.
- A subscriber's own `OnNext` throw propagates back to the producer (the `PropertyChanged` setter), exactly like `Subject<T>.OnNext`. `main` does this too.

This PR wraps the downstream `OnNext` and routes it to `OnError`. Empirically that does **not** let a supplied error handler observe the failure: Rx's `AutoDetachObserver` detaches the moment `OnNext` throws, so the routed `OnError` lands on a stopped observer and no-ops. The net effect is a silent swallow with no diagnostics, i.e. the behavior we moved away from in #1111.

Verified with probes:
- Against this PR: subscriber `OnNext` throws with a swallowing `onError` supplied -> `onErrorCalls=0` (handler never fires).
- Against `main`: a terminal `OnNext` throw propagates to the setter; an accessor throw is observed by the `onError` handler with the setter left safe.

@JakenVeina's conclusion was right: if it isn't handled downstream, that exception belongs escaping to the setter. Consumers who need to shield an INPC setter should handle it downstream (`.Catch`), hop threads (`.ObserveOn`), or via an explicit opt-in overload. No source change needed.

PR #1132, opened 2026-07-04
Fixed #1131 - Race condition between `SuspendNotifications()` and `.ResumeNotifications()`

### LINKED ISSUE #1131: [Bug]: Intermittent test failure for in `SuspendNotificationsFixture`
Specifically, the newly-added test `ConcurrentSuspendDuringResumeDoesNotCorrupt`.

Given that this test seeks to force two methods to run concurrently, I think the main problem is that it's being allowed to run in parallel. That is, when running the test by itself, it fails nearly every time, but when running as part of the full suite, it tends to often pass. Which makes sense, because if the `ThreadPool` is being saturated by all the other tests running in parallel, it drastically reduces the likelihood that the two parallel `Task`s that get scheduled don't actually get scheduled concurrently. The `Barrier` being used here only increases the likelihood that the meaningful units under test get scheduled concurrently, by making sure they wait for each other to be read, but it doesn't actually guarantee concurrent scheduling.

The biggest thing we can do is move this test into a separate fixture and disable parallelism. Which we should probably also do on a LOT of other concurrency tests.

Also, this does seem to be a legit defect in `ResumeNotifications()`. `SuspendNotification()`s `lock`s on all the state its modifying, but `ResumeNotifications()` doesn't.

JakenVeina:
Fixed that `ObservableCache.SuspendNotifications()` uses a `lock` to synchronize updates to internal state for tracking notification suspensions, but `.ResumeNotifications()` did not use any synchronization when updating that same state, allowing for corruption of state if the two methods happen to overlap, concurrently.

Also fixed that the test for this behavior was intermittent, specifically that it tended to falsely pass when running in parallel with many other tests. The test has now been moved to its own `IntegrationTests` fixture, to guarantee it is never parallelized with other tests.

Resolves #1131.

@dwcullop I suspect this may not be a proper fix, in the context of `DeliveryQueue<>`. I.E. this solution involves using a basic `lock()` when perhaps the proper solution should be using the locking mechanisms upon `DeliveryQueue<>`. But I don't have the familiarity with it. For now, I just duplicated the existing `lock(_locker)` upon `.SuspendNotifications()` over to `.ResumeNotifications()`. If you'd rather take a crack at this yourself instead, go for it.

The test changes we should definitely keep, though.


dwcullop:
The lock that you're using is the same lock the DQ uses.  However, doing the Resume while holding the lock will cause the changes to be emitted while holding the lock.  Let me see if I can suggest an alternative.

src/DynamicData.Tests/Cache/SuspendNotificationsFixture.IntegrationTests.cs
```diff
@@ -13,6 +13,124 @@ public static partial class SuspendNotificationsFixture
     public sealed class IntegrationTests
         : IntegrationTestFixtureBase
     {
+        [Fact]
```
JakenVeina, line 18: Any particular reason to move this?

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -358,11 +352,18 @@ private void ResumeNotifications()
...
+        // accumulated changes are delivered first. Re-check the suspend count
+        // under the lock: a concurrent SuspendNotifications() may have run in the
+        // window since the count was decremented, and its state must win. Emitting
+        // the signal only when still unsuspended keeps _notifySuspendCount and the
+        // suspended-notification subject from diverging.
         if (emitResume)
         {
             using var readLock = _notifications.AcquireReadLock();
-            _suspensionTracker.Value.EmitResumeNotification();
+            if (!_suspensionTracker.Value.AreNotificationsSuspended)
```
JakenVeina: So, we're basically saying "state corruption is cool, if we just guard against the impact"? I don't think this is sound the way it's currently implemented. In particular, `.SuspendNotifications()` does not acquire any locks from `_notifications`, so even with this lock here, it's possible that a concurrent `.SuspendNotifications()` can swap `_suspensionTracker.Value.AreNotificationsSuspended` from `false` to `true` in the time between this check, and the call to `.EmitResumeNotification()`.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -482,10 +483,16 @@ public void EnqueueChanges(IEnumerable<Change<TObject, TKey>> changes)
...
         {
-            if (++_notifySuspendCount == 1)
+            ++_notifySuspendCount;
+
+            // Signal suspension based on the subject's own state rather than the count.
+            // ResumeNotifications() decrements the count and emits its resume signal in
+            // separate steps, so the count can briefly reach zero before the 'false' signal
+            // is emitted. Gating on the subject keeps the signal monotonic and avoids a
+            // redundant 'true' when a suspend interleaves in that window.
+            if (!_areNotificationsSuspended.IsDisposed && !_areNotificationsSuspended.Value)
```
JakenVeina: Would `Interlocked.Increment()` and `Interlocked.Decrement()` not make more sense for tracking `_notifySuspendCount`?

Otherwise, does `_notifySuspendCount` even have a purpose anymore?

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -352,11 +352,18 @@ private void ResumeNotifications()
...
+        // window since the count was decremented, and its state must win. Emitting
+        // the signal only when still unsuspended keeps _notifySuspendCount and the
+        // suspended-notification subject from diverging.
         if (emitResume)
         {
             using var readLock = _notifications.AcquireReadLock();
-            _suspensionTracker.Value.EmitResumeNotification();
+            if (!_suspensionTracker.Value.AreNotificationsSuspended)
+            {
+                _suspensionTracker.Value.EmitResumeNotification();
```
dwcullop: **Should fix: this can throw `ObjectDisposedException`, and the invariant is being enforced from outside the class that owns it.**

**The bug:** the re-check tests `AreNotificationsSuspended` but not `_areNotificationsSuspended.IsDisposed`. `SuspensionTracker.Dispose()` is called from `CacheUpdateObserver.OnCompleted` / `OnError`, which run **outside** `_locker` during delivery. So a terminal notification draining on another thread can dispose the subject in the window between this scope being entered and `EmitResumeNotification()` running, and `OnNext(false)` on a disposed `BehaviorSubject` throws. Note that `SuspensionTracker.ResumeNotifications()` already guards for exactly this with `!_areNotificationsSuspended.IsDisposed`; this new path does not.

**The design issue:** `ObservableCache` reaching into `_suspensionTracker.Value.AreNotificationsSuspended` to decide whether calling `EmitResumeNotification()` is legal is Tell-Don't-Ask. The suspend count, the subject, and the relationship between them are all `SuspensionTracker`'s private business, and the rule now lives in two places.

Both problems go away by folding the guard into the tracker, which also makes `EmitResumeNotification()` safe to call unconditionally:

```csharp
// SuspensionTracker
public void EmitResumeNotification()
{
    // A concurrent SuspendNotifications() may have run since the count was
    // decremented, and its state must win. Emitting only when still unsuspended
    // keeps _notifySuspendCount and the subject from diverging.
    if ((_notifySuspendCount == 0) && !_areNotificationsSuspended.IsDisposed)
    {
        _areNotificationsSuspended.OnNext(false);
    }
}
```

The caller then reverts to its shape on `main`:

```csharp
if (emitResume)
{
    using var readLock = _notifications.AcquireReadLock();
    _suspensionTracker.Value.EmitResumeNotification();
}
```

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -476,10 +483,16 @@ public void EnqueueChanges(IEnumerable<Change<TObject, TKey>> changes)
...
         {
-            if (++_notifySuspendCount == 1)
+            ++_notifySuspendCount;
+
+            // Signal suspension based on the subject's own state rather than the count.
+            // ResumeNotifications() decrements the count and emits its resume signal in
+            // separate steps, so the count can briefly reach zero before the 'false' signal
+            // is emitted. Gating on the subject keeps the signal monotonic and avoids a
+            // redundant 'true' when a suspend interleaves in that window.
+            if (!_areNotificationsSuspended.IsDisposed && !_areNotificationsSuspended.Value)
```
dwcullop: Minor, but worth doing while you are in here: `_areNotificationsSuspended.Value` and `.IsDisposed` are two separate reads with no atomicity between them, so this has the same TOCTOU as the resume path (a concurrent `Dispose()` from the off-lock delivery callback can slip between them and make `.Value` throw). `BehaviorSubject<T>.Value` also takes the subject's own internal lock on every call.

Since **every** mutation of this state already happens under `_locker`, a plain field is cheaper, race-free, and makes the intent explicit:

```csharp
private bool _isSuspendSignalled;

public void SuspendNotifications()
{
    ++_notifySuspendCount;

    // ResumeNotifications() decrements the count and emits its resume signal in
    // separate steps, so the count can briefly reach zero before the 'false' signal
    // is emitted. Gating on the signalled state keeps the signal monotonic and avoids
    // a redundant 'true' when a suspend interleaves in that window.
    if (!_isSuspendSignalled && !_areNotificationsSuspended.IsDisposed)
    {
        Debug.Assert(_pendingChanges.Count == 0, "Shouldn't be any pending values if suspend was just started");
        _isSuspendSignalled = true;
        _areNotificationsSuspended.OnNext(true);
    }
}
```

That reduces the subject to a pure output channel, which is all any consumer actually uses it for.

(FWIW I verified the `Debug.Assert` you kept is still sound under the new gate: `_pendingChanges` is only appended to from `EnqueueChanges` while the count is above zero, and the count above zero now implies the subject is already `true`, so the assert cannot be reached with a non-empty list.)

src/DynamicData.Tests/Cache/SuspendNotificationsFixture.IntegrationTests.cs
```diff
@@ -0,0 +1,182 @@
...
+            //
+            // Dedicated threads are used (rather than the thread pool) so the test is immune
+            // to pool starvation when run alongside other test collections.
+            using var cache = new SourceCache<int, int>(static x => x);
+            using var deliveryStarted = new ManualResetEventSlim(false);
+            using var releaseDelivery = new ManualResetEventSlim(false);
+            using var concurrentOpDone = new ManualResetEventSlim(false);
+
+            // An active subscriber (connected before suspension) that blocks on its first delivery.
+            var blockOnce = true;
```
dwcullop, line 35: These two integration tests share `blockOnce` across the resume thread and the delivering thread with no barrier or `volatile`. It works in practice because the surrounding `ManualResetEventSlim.Set()` / `.Wait()` pair provides the necessary fences, but that is incidental rather than stated. A `volatile bool`, or `Interlocked.Exchange(ref _blockOnce, 0) == 1`, would make the intent explicit and survive future edits that move the event signalling around.

Same applies to the copy on line 93.

src/DynamicData.Tests/Cache/SuspendNotificationsFixture.UnitTests.cs
```diff
@@ -0,0 +1,464 @@
...
+using System.Linq;
+using System.Reactive.Linq;
+using System.Threading.Tasks;
+
+using FluentAssertions;
+using Xunit;
+
+using DynamicData.Kernel;
+
+namespace DynamicData.Tests.Cache
```
dwcullop: Two style deviations in this file:

1. **Block-scoped namespace.** Every other file in the repo uses file-scoped, including `SuspendNotificationsFixture.IntegrationTests.cs` added in this same PR and the `SuspendNotificationsFixture.cs` this was split out of. Beyond consistency, the block scoping re-indents all 464 lines, so `git diff` cannot detect this as a move and the change is effectively unreviewable. Switching to `namespace DynamicData.Tests.Cache;` and dropping one indent level should make the diff collapse to near zero.
2. **Missing UTF-8 BOM.** The repo convention is UTF-8 with BOM; this file is written without one.

Also, `public sealed partial class UnitTests` on line 16 is marked `partial` with no second part.

dwcullop (commented):
## Review summary

**Verdict: approve with minor changes.** The invariant restored here is correct, and the test isolation does what it claims.

Note for context: findings 1 through 3 below are against head commit `6911074d`, not @JakenVeina's original commit. The approach in the head commit is a better fix than the plain `lock()` duplication originally proposed.

### Validation performed

- Full `net9.0` suite on the PR branch: **2439 passed, 0 failed, 38s**.
- `SuspendNotificationsFixture` in isolation: **24 passed in 170ms**, and still 170ms when run concurrently with a second full test suite saturating every core. The move to the non-parallel `IntegrationTests` collection achieves its goal.
- Merges cleanly with #1133 (no textual conflicts) and is semantically compatible with it.

### Correctness analysis

I traced the invariant `_notifySuspendCount > 0` <=> `_areNotificationsSuspended.Value == true` through every interleaving. It holds:

- `false` can never be emitted while the count is above zero, because the re-check and the `OnNext` are both inside the same `_locker` acquisition and no other thread can mutate the count without that lock.
- `false` can never be lost, because whichever resumer observes a zero count last will emit it.
- `true` can never be emitted twice, because the gate reads the subject's own value under `_locker`.
- A duplicate `false` **is** possible when two resumers race in the window after the enqueue scope is released. It is harmless behind `Where(b => !b).Take(1)`, but worth knowing.

---

### 4. Please call out the behavioral change

`ResumeSignalUnderLockPreventsStaleSnapshotFromReSuspend` was removed, and it asserted the **opposite** of the new behavior: that a deferred `Connect()` activates on a resume signal that is immediately followed by a re-suspend, snapshotting only batch 1. Under this PR the subscriber correctly stays deferred and receives batch 1 and batch 2 as a single changeset on the real resume.

I think the new behavior is more correct: a `Connect()` issued during a suspension should deliver nothing until the suspension truly ends. But per the repo's SemVer guidance, behavioral changes are breaking even when the signature is unchanged, so this belongs in the PR description and release notes rather than only in a deleted test.

### 5. Style nits

- `SuspendNotificationsFixture.UnitTests.cs` has **no UTF-8 BOM** and uses a **block-scoped namespace**. Every other file in the repo uses file-scoped, including the sibling `SuspendNotificationsFixture.IntegrationTests.cs` added in this same PR, and the `SuspendNotificationsFixture.cs` it was split out of. The block scoping also re-indents all 464 lines, which hides the fact that this is a pure move and makes the diff unreviewable.
- `SuspendNotificationsFixture.IntegrationTests.cs` also lacks the BOM (compare `FilterFixture.DynamicPredicate.IntegrationTests.cs`, which has one).
- `public sealed partial class UnitTests` is declared `partial` with no second part.
- `SuspensionsAreThreadSafe` stays in the parallelized `UnitTests`. It is benign (no interleaving requirement, so it cannot go intermittent the way the moved test did), but the PR's stated rationale about other concurrency tests would apply.

dwcullop:
You're right, they shouldn't diverge. I went back and looked at this again and I think I was patching the symptom rather than the cause.

The actual problem is that `ResumeNotifications()` takes the lock twice. The `using` block grabs it to decrement the count and queue up the pending changes, releases it, and then the resume signal grabs it *again* in a second acquisition. Everything is under a lock, so nothing looks wrong at a glance, but the gap between those two acquisitions is the race window. A suspend that lands in there sees a zero count and skips emitting `true`, and then the resume comes along and emits its `false` on top of it. My re-check just made the second acquisition notice it had been beaten, which is exactly the "state corruption is cool if we guard against it" thing you called out.

So rather than re-checking, don't have a second acquisition:

```diff
     private void ResumeNotifications()
     {
-        bool emitResume;
+        using var notifications = _notifications.AcquireLock();

-        using (var notifications = _notifications.AcquireLock())
-        {
-            Debug.Assert(_suspensionTracker.IsValueCreated, "Should not be Resuming Notifications without Suspend Notifications instance");
+        Debug.Assert(_suspensionTracker.IsValueCreated, "Should not be Resuming Notifications without Suspend Notifications instance");

-            (var changes, emitResume) = _suspensionTracker.Value.ResumeNotifications();
-            if (changes is not null)
-            {
-                notifications.EnqueueNext(new CacheUpdate(changes, _readerWriter.Count, ++_currentVersion));
-            }
+        var (changes, emitResume) = _suspensionTracker.Value.ResumeNotifications();
+        if (changes is not null)
+        {
+            notifications.EnqueueNext(new CacheUpdate(changes, _readerWriter.Count, ++_currentVersion));
         }

-        // Emit the resume signal after releasing the delivery scope so that
-        // accumulated changes are delivered first. Re-check the suspend count
-        // under the lock: a concurrent SuspendNotifications() may have run in the
-        // window since the count was decremented, and its state must win. Emitting
-        // the signal only when still unsuspended keeps _notifySuspendCount and the
-        // suspended-notification subject from diverging.
+        // Emit the resume signal in the same lock acquisition that cleared the suspend
+        // count, so the count and the signal can never disagree: there is no window for a
+        // concurrent SuspendNotifications() to observe one without the other. The accumulated
+        // changes are still delivered off the lock, when this scope exits. A deferred
+        // subscriber that activates on this signal snapshots the current state and skips the
+        // still-pending delivery via the version guard in CreateConnectObservable.
         if (emitResume)
         {
-            using var readLock = _notifications.AcquireReadLock();
-            if (!_suspensionTracker.Value.AreNotificationsSuspended)
-            {
-                _suspensionTracker.Value.EmitResumeNotification();
-            }
+            _suspensionTracker.Value.EmitResumeNotification();
         }
     }
```

The thing that makes this work, and the reason I originally thought we couldn't do it, is that `ScopedAccess.Dispose()` calls `ExitLockAndDeliver()`, which releases the lock *and then* drains the queue. So the enqueue and the signal both happen in one critical section, and the accumulated changes still go out with nothing held. `ResumeDeliversPendingChangesWithoutHoldingTheLock` still passes, so we're not trading the race for the deadlock risk that your original `lock(cache._locker)` version would have introduced.

One consequence worth noting: the signal now fires *before* the pending changes are delivered instead of after. That turns out to be fine, because a deferred subscriber that activates on the signal takes its snapshot from `GetInitialUpdates()` and then skips the queued delivery it already has via the `HasPending` / `_currentDeliveryVersion` guard in `CreateConnectObservable`. That guard is there for precisely this case.

Nice side effect: with the window actually closed instead of papered over, `ResumeSignalUnderLockPreventsStaleSnapshotFromReSuspend` can come back. I dropped it back into the integration fixture and it passes right alongside your `StaleResumeSignalIsSuppressedByConcurrentReSuspend`. I'd much rather have both than delete one that was asserting real behavior. 2443 passing locally on net9.0, and I ran the suspend fixture 8 times back to back to make sure the integration tests weren't flaky, all green at 5s.

Two small things folded into the diff above:

- `EmitResumeNotification()` needs an `IsDisposed` check. `SuspensionTracker.Dispose()` runs out of the `CacheUpdateObserver` completion/error callbacks, and those execute off the lock during delivery, so the subject can get torn down underneath a resume on another thread. `ResumeNotifications()` already guards for this, that path didn't.
- Went back to `++_notifySuspendCount == 1` for the suspend gate. Once the count and the signal can't disagree, reading the subject's `.Value` isn't buying anything (and it takes the subject's own internal lock on every call), plus it lets the `SuspendSubject should be false for the first suspend call` assert come back.

On `Interlocked` for the counter, I don't think it helps here. Every mutation of it already happens inside `_locker`, so the increment itself was never the problem. The problem was the count and the subject getting updated in two different critical sections, and making the counter atomic on its own wouldn't have closed that.

Let me know what you think.

JakenVeina:
> `false` can never be emitted while the count is above zero, because the re-check and the `OnNext` are both inside the same `_locker` acquisition and no other thread can mutate the count without that lock.

That's just blatantly untrue. `++_notifySuspendCount` on line 486 and `--_notifySuspendCount` on line 508 happen underneath two separate and unsynchronized locks. The increment is inside `lock(_locker)` from line 166, and the decrement is inside `_notifications.AcquireLock()` from line 343. This was basically my point in my [prior comment](https://github.com/reactivemarbles/DynamicData/pull/1132#discussion_r3563588650).

> SuspendNotificationsFixture.UnitTests.cs ... uses a block-scoped namespace. Every other file in the repo uses file-scoped, including the sibling SuspendNotificationsFixture.IntegrationTests.cs added in this same PR, and the SuspendNotificationsFixture.cs it was split out of.

Fixed.

> SuspensionsAreThreadSafe stays in the parallelized UnitTests. It is benign (no interleaving requirement, so it cannot go intermittent the way the moved test did), but the PR's stated rationale about other concurrency tests would apply.

Fixed.

And your new comment just popped up, while I was working this, so I'll just leave it at these two fixes.

dwcullop:
Pushed this as `bf07da5`, since you said go ahead.

Kept it to the two files: the `ResumeNotifications()` / `SuspendNotifications()` / `EmitResumeNotification()` change described above, plus dropping `ResumeSignalUnderLockPreventsStaleSnapshotFromReSuspend` back into the integration fixture. I left the rest of your nits alone since you'd already handled the styling and moved `SuspensionsAreThreadSafe` over.

2444 passing on net9.0, and I ran the suspend fixture 5 more times against the real branch to be sure the integration tests are solid: 26 tests, 5s, green every time.

dwcullop:
They're the same lock, though. Both `ObservableCache` constructors hand `_locker` to the `DeliveryQueue` as its gate:

```csharp
_notifications = new DeliveryQueue<CacheUpdate>(_locker, new CacheUpdateObserver(this));
```

So `lock (_locker)` on 166 and `_notifications.AcquireLock()` on 343 are taking the same object. That's what I was getting at in my first comment on this PR about the DQ using the same lock, which I should have been clearer about at the time.

Doesn't really change anything about the fix, though, and your instinct about it was right regardless. Two acquisitions of the same lock still leave a gap between them, and that gap is where this bit us: resume drops the count to zero and exits the scope, a suspend grabs the lock and bumps it back to one and emits a second `true`, then resume re-enters and puts `false` on top of it. Same lock, still a window.

JakenVeina:
> The thing that makes this work, and the reason I originally thought we couldn't do it, is that ScopedAccess.Dispose() calls ExitLockAndDeliver(), which releases the lock and then drains the queue. So the enqueue and the signal both happen in one critical section, and the accumulated changes still go out with nothing held. 

I find this pretty misleading then, so I'd definitely like to improve that aspect of the `DeliveryQueue` at some point. Even if it's just as simple as putting "ReleaseLockAnd" in the name of that resume method. But I can live with it, for now.

> Let me know what you think.

I think the entire concept of buffering/batching/queueing notifications within `SourceCache<>` needs a re-design. We have 2 competing systems trying to do the same thing, and stepping on each others' toes: the `SuspendNotifications()` system and the `.Edit()` system. But that's definitely out-of-scope. This implementation, while not ideal, looks sound.

> They're the same lock, though. Both `ObservableCache` constructors hand `_locker` to the `DeliveryQueue` as its gate:

I actually had the thought to wonder if that was the case, and looked for it. Somehow, I missed it, anyway.

> On `Interlocked` for the counter, I don't think it helps here. Every mutation of it already happens inside _locker, so the increment itself was never the problem.

I meant, like, in terms of the possibility that we could implement the entire suspension system without any locking at all (except for the actual delivery). If the whole system hinges on that one counter, it should be possible, in theory.

dwcullop:
Won't let me approve because I pushed a commit, but I'd approve if I could.

JakenVeina:
> Won't let me approve because I pushed a commit, but I'd approve if I could.

Oh. Fun. A deadlock.

Well, I was able to adjust the branch protection rules to allow Maintainers to bypass them, for now. It requires an explicit additional confirmation to do so, so I'm good with leaving it that way, as long as you and @RolandPheasant are as well.

PR #1133, opened 2026-07-04
Fixed that `.Edit()` operations would not defer notifications for new subscriptions occurring during the edit.

### LINKED ISSUE #1129: [Bug]: `ArgumentException: An item with the same key was already added` with Filter on 9.4.31+
### Describe the bug 🐞

I ran into the following exception in my application:
```
System.ArgumentException: An item with the same key has already been added. Key: GroupId { Value = 2047 }
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at DynamicData.ChangeAwareCache`2.Add(TObject item, TKey key) in /_/src/DynamicData/Cache/ChangeAwareCache.cs:line 75
   at DynamicData.Cache.Internal.Filter.Static`2.<>c__DisplayClass0_1.<Create>b__1(IChangeSet`2 upstreamChanges) in /_/src/DynamicData/Cache/Internal/Filter.Static.cs:line 36
   at System.Reactive.Linq.ObservableImpl.Select`2.Selector._.OnNext(TSource value)
```

After investigation it appears in DynamicData 9.4.31 and newer and only seems to happen when multiple `SourceCache`s, multiple subscriptions and multiple schedulers are applied.

### Step to reproduce

```csharp
using System.Reactive.Linq;
using DynamicData;
using Microsoft.Reactive.Testing;

namespace FilterTests;

public class FilterTests
{
    private static TestScheduler scheduler = new TestScheduler();

    [Fact]
    public async Task Filter_Throws_ArgumentException()
    {
        var viewModels = new SourceCache<GroupViewModel, int>(x => x.Group.Id);
        var groups = new SourceCache<Group, int>(x => x.Id);

        viewModels.Connect().Subscribe();

        groups.Connect()
            .Subscribe(changes => {
                viewModels.Edit(inner => {
                    foreach (var change in changes) {
                        switch (change.Reason)
                        {
                            case ChangeReason.Add:
                                inner.AddOrUpdate(new GroupViewModel(change.Current, viewModels));
                                break;
                        }
                    }
                });
            });

        groups.Edit(inner => inner.AddOrUpdate([
                new (1, 0),
                new (0, null)
            ]));

        scheduler.AdvanceBy(10);
    }

    private record Group(int Id, int? Parent);

    private class GroupViewModel
    {
        public Group Group { get; }

        public GroupViewModel(Group group, SourceCache<GroupViewModel, int> masterList)
        {
            Group = group;

            masterList
                .Connect()
                .Filter(n => n.Group.Parent == group.Id)
                .ObserveOn(scheduler)
                .Subscribe();
        }
    }
}
```

### Reproduction repository

See above.

### Expected behavior

Should not throw ArgumentException.

### DynamicData Version

9.4.31

### Additional information ℹ️

Took me over 8 hours to get this reproduced, I hope you guys fix this once and for all.

JakenVeina:
Fixed that subscriptions occurring during an `.Edit()` operation upon a `SourceCache<>` or `SourceList<>` would still publish an initial notification to the subscriber. This resulted in the potential for notifications to be duplicated, for items added or manipulated during the edit.

Instead, such subscriptions now have their notifications deferred until the `.Edit()` is complete.

Resolves #1129.

### DIFF
--- a/src/DynamicData.Benchmarks/Miscellaneous/LockImplementations.cs
+++ b/src/DynamicData.Benchmarks/Miscellaneous/LockImplementations.cs
@@ -1,0 +1,42 @@
+using System.Threading;
+using BenchmarkDotNet.Attributes;
+
+namespace DynamicData.Benchmarks.Miscellaneous;
+
+[MemoryDiagnoser]
+[MarkdownExporterAttribute.GitHub]
+public class LockImplementations
+{
+    public LockImplementations()
+    {
+        _objectGate     = new();
+        _threadingGate  = new();
+    }
+
+    [Benchmark(Baseline = true)]
+    public int NoLock()
+    {
+        return 0;
+    }
+
+    [Benchmark]
+    public int ObjectLock()
+    {
+        lock (_objectGate)
+        {
+            return 0;
+        }
+    }
+
+    [Benchmark]
+    public int ThreadingLock()
+    {
+        lock (_threadingGate)
+        {
+            return 0;
+        }
+    }
+
+    private readonly object _objectGate;
+    private readonly Lock   _threadingGate;
+}
--- a/src/DynamicData.Tests/Cache/SourceCacheFixture.cs
+++ b/src/DynamicData.Tests/Cache/SourceCacheFixture.cs
@@ -163,8 +163,6 @@ public class SourceCacheFixture : IDisposable
         change!.Count.Should().Be(0);
     }
 
-
-
     [Fact]
     public void StaticFilterRemove()
     {
@@ -191,7 +189,6 @@ public class SourceCacheFixture : IDisposable
     }
 
     public record class SomeObject(int Id, int Value);
-
 
     [Fact]
     public async Task MultiCacheFanInDoesNotDeadlock()
@@ -403,6 +400,23 @@ public class SourceCacheFixture : IDisposable
             "all items in the source should have propagated downstream");
 
         results.HasCompleted.Should().BeFalse("the source has not yet completed");
+    }
+
+    [Fact]
+    public void ConnectContinuesToWorkNormallyAfterAFailedEdit()
+    {
+        using var source = new SourceCache<int, int>(static item => item);
+    
+        source.AddOrUpdate(1);
+
+        source.Invoking(source => source.Edit(_ => throw new Exception("Test")))
+            .Should().Throw<Exception>()
+            .WithMessage("Test");
+
+        using var subscription = source.Connect().RecordCacheItems(out var results);
+
+        results.Error.Should().BeNull("new subscribers should not receive previous errors");
+        results.RecordedChangeSets.Should().ContainSingle("an initial changeset should have been published.");
     }
 
     private sealed record TestItem(string Key, string Value);
--- a/src/DynamicData.Tests/Cache/SourceCacheFixture.cs
+++ b/src/DynamicData.Tests/Cache/SourceCacheFixture.cs
@@ -354,7 +354,7 @@ public class SourceCacheFixture : IDisposable
 
     // Covers https://github.com/reactivemarbles/DynamicData/issues/1129
     [Fact]
-    public void ConnectDuringEditDoesNotDuplicate()
+    public void ConnectDuringEditsDoesNotDuplicate()
     {
         using var items = new SourceCache<int, int>(static item => item);
         
@@ -385,17 +385,20 @@ public class SourceCacheFixture : IDisposable
 
             results.Error.Should().BeNull("no errors should have occurred");
             results.RecordedChangeSets.Should().BeEmpty("no changes should be published in the middle of an edit");
+
+            // Explicitly doing a nested edit, as that system is closely intertwined with the edit-tracking system that
+            // .Connect() uses.
+            items.Remove(item: 1);
+
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.RecordedChangeSets.Should().BeEmpty("no changes should be published in the middle of an edit");
         });
         
         results.Should().NotBeNull("the edit delegate should have been invoked");
         results.Error.Should().BeNull("no errors should have occurred");
         results.RecordedChangeSets.Should().ContainSingle("subscribers should only receive a single initial changeset");
         results.RecordedItemsByKey.Should().BeEquivalentTo(
-            new Dictionary<int, int>()
-            {
-                [1] = 1,
-                [2] = 2
-            },
+            new Dictionary<int, int>() { [2] = 2 },
             options => options.WithoutStrictOrdering(),
             "all items in the source should have propagated downstream");
 
--- a/src/DynamicData.Tests/Cache/SourceCacheFixture.cs
+++ b/src/DynamicData.Tests/Cache/SourceCacheFixture.cs
@@ -1,12 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 
 using DynamicData.Tests.Domain;
-
+using DynamicData.Tests.Utilities;
 using FluentAssertions;
 
 using Xunit;
@@ -352,6 +353,56 @@ public class SourceCacheFixture : IDisposable
         // Each key should appear exactly once in the new subscriber's view
         addCounts.GetValueOrDefault("k1").Should().Be(1, "k1 should appear once (snapshot only)");
         addCounts.GetValueOrDefault("k2").Should().Be(1, "k2 should appear once, not duplicated from snapshot + queued delivery");
+    }
+
+    // Covers https://github.com/reactivemarbles/DynamicData/issues/1129
+    [Fact]
+    public void ConnectDuringEditDoesNotDuplicate()
+    {
+        using var items = new SourceCache<int, int>(static item => item);
+        
+        using var subscriptions = new CompositeDisposable();
+        
+        // An initial subscription is required to initiate internal buffering of changes, during the upcoming .Edit().
+        // That is, we want there to be changes buffered, internally, when the mid-edit subscription comes in, to
+        // ensure that they don't get duplicated. This is the scenario that came in up #1129. 
+        subscriptions.Add(items
+            .Connect()
+            .Subscribe());
+            
+        CacheItemRecordingObserver<int, int>? results = null;
+            
+        items.Edit(inner =>
+        {
+            inner.AddOrUpdate(1);
+
+            subscriptions.Add(items
+                .Connect()
+                .ValidateChangeSets(static item => item)
+                .RecordCacheItems(out results));
+        
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.RecordedChangeSets.Should().BeEmpty("no changes should be published in the middle of an edit");
+        
+            inner.AddOrUpdate(2);
+
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.RecordedChangeSets.Should().BeEmpty("no changes should be published in the middle of an edit");
+        });
+        
+        results.Should().NotBeNull("the edit delegate should have been invoked");
+        results.Error.Should().BeNull("no errors should have occurred");
+        results.RecordedChangeSets.Should().ContainSingle("subscribers should only receive a single initial changeset");
+        results.RecordedItemsByKey.Should().BeEquivalentTo(
+            new Dictionary<int, int>()
+            {
+                [1] = 1,
+                [2] = 2
+            },
+            options => options.WithoutStrictOrdering(),
+            "all items in the source should have propagated downstream");
+
+        results.HasCompleted.Should().BeFalse("the source has not yet completed");
     }
 
     private sealed record TestItem(string Key, string Value);
--- a/src/DynamicData/Cache/ObservableCache.cs
+++ b/src/DynamicData/Cache/ObservableCache.cs
@@ -198,24 +198,29 @@ internal sealed class ObservableCache<TObject, TKey> : IObservableCache<TObject,
 
         _editLevel++;
         _isEditInProgress.OnNext(_editLevel is not 0);
-        if (_editLevel == 1)
+        try
         {
-            var previewHandler = _changesPreview.HasObservers ? (Action<ChangeSet<TObject, TKey>>)InvokePreview : null;
-            changes = _readerWriter.Write(updateAction, previewHandler, _changes.HasObservers);
+            if (_editLevel == 1)
+            {
+                var previewHandler = _changesPreview.HasObservers ? (Action<ChangeSet<TObject, TKey>>)InvokePreview : null;
+                changes = _readerWriter.Write(updateAction, previewHandler, _changes.HasObservers);
+            }
+            else
+            {
+                _readerWriter.WriteNested(updateAction);
+            }
         }
-        else
+        finally
         {
-            _readerWriter.WriteNested(updateAction);
+            _editLevel--;
+
+            if (changes is not null && _editLevel == 0)
+            {
+                notifications.EnqueueNext(new CacheUpdate(changes, _readerWriter.Count, ++_currentVersion));
+            }
+
+            _isEditInProgress.OnNext(_editLevel is not 0);
         }
-
-        _editLevel--;
-
-        if (changes is not null && _editLevel == 0)
-        {
-            notifications.EnqueueNext(new CacheUpdate(changes, _readerWriter.Count, ++_currentVersion));
-        }
-
-        _isEditInProgress.OnNext(_editLevel is not 0);
     }
 
     internal void UpdateFromSource(Action<ISourceUpdater<TObject, TKey>> updateAction)
@@ -228,24 +233,29 @@ internal sealed class ObservableCache<TObject, TKey> : IObservableCache<TObject,
 
         _editLevel++;
         _isEditInProgress.OnNext(_editLevel is not 0);
-        if (_editLevel == 1)
+        try
         {
-            var previewHandler = _changesPreview.HasObservers ? (Action<ChangeSet<TObject, TKey>>)InvokePreview : null;
-            changes = _readerWriter.Write(updateAction, previewHandler, _changes.HasObservers);
+            if (_editLevel == 1)
+            {
+                var previewHandler = _changesPreview.HasObservers ? (Action<ChangeSet<TObject, TKey>>)InvokePreview : null;
+                changes = _readerWriter.Write(updateAction, previewHandler, _changes.HasObservers);
+            }
+            else
+            {
+                _readerWriter.WriteNested(updateAction);
+            }
         }
-        else
+        finally
         {
-            _readerWriter.WriteNested(updateAction);
+            _editLevel--;
+
+            if (changes is not null && _editLevel == 0)
+            {
+                notifications.EnqueueNext(new CacheUpdate(changes, _readerWriter.Count, ++_currentVersion));
+            }
+
+            _isEditInProgress.OnNext(_editLevel is not 0);
         }
-
-        _editLevel--;
-
-        if (changes is not null && _editLevel == 0)
-        {
-            notifications.EnqueueNext(new CacheUpdate(changes, _readerWriter.Count, ++_currentVersion));
-        }
-
-        _isEditInProgress.OnNext(_editLevel is not 0);
     }
 
     private IObservable<IChangeSet<TObject, TKey>> CreateConnectObservable(Func<TObject, bool>? predicate, bool suppressEmptyChangeSets) =>
--- a/src/DynamicData/Cache/ObservableCache.cs
+++ b/src/DynamicData/Cache/ObservableCache.cs
@@ -119,22 +119,41 @@ internal sealed class ObservableCache<TObject, TKey> : IObservableCache<TObject,
         {
             lock (_locker)
             {
-                var observable = ((!_suspensionTracker.IsValueCreated || !_suspensionTracker.Value.AreNotificationsSuspended)
-                        && (!_isEditInProgress.Value))
+                var observable = (
+                        _suspensionTracker.IsValueCreated,
+                        _isEditInProgress.Value)
+                    switch
+                    {
+                        // No suspensions or edits active, create the connection immediately.
+                        (false, false) => CreateConnectObservable(predicate, suppressEmptyChangeSets),
 
-                    // Create the Connection Observable
-                    ? CreateConnectObservable(predicate, suppressEmptyChangeSets)
+                        // Edit in progress, but suspension system is inactive
+                        // Need to avoid activating the suspension system if we don't absolutely need to, as it adds
+                        // locking overhead to edit operations. But then when the edit is done, we need to check if a
+                        // suspension came in. If so, do a followup wait with the full logic for both systems. It
+                        // needs to be the full logic, in case an edit comes in during the suspension, and so on.
+                        (false, true) => _isEditInProgress
+                            .Where(static shouldConnectionBeDeferred => !shouldConnectionBeDeferred)
+                            .Take(1)
+                            .Select(_ => CreateFullyDeferredConnection())
+                            .Switch(),
 
-                    // Defer until notifications are no longer suspended
-                    : Observable.CombineLatest(
+                        // If the suspension system is already active, we can monitor both systems simultaneously, and
+                        // make the connection as soon as both are idle at the same time.
+                        _ => CreateFullyDeferredConnection()
+                    };
+
+                return observable.SubscribeSafe(observer);
+
+                IObservable<IChangeSet<TObject, TKey>> CreateFullyDeferredConnection()
+                    => Observable.CombineLatest(
                             _suspensionTracker.Value.NotificationsSuspendedObservable,
                             _isEditInProgress,
                             static (areNotificationsSuspended, isEditInProgress) => areNotificationsSuspended || isEditInProgress)
+                        .Do(static _ => { }, observer.OnCompleted)
                         .Where(static shouldConnectionBeDeferred => !shouldConnectionBeDeferred)
                         .Take(1)
                         .SelectMany(_ => CreateConnectObservable(predicate, suppressEmptyChangeSets));
-
-                return observable.SubscribeSafe(observer);
             }
         });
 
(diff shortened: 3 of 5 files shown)


src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -197,6 +218,8 @@ internal void UpdateFromIntermediate(Action<ICacheUpdater<TObject, TKey>> update
         {
             notifications.EnqueueNext(new CacheUpdate(changes, _readerWriter.Count, ++_currentVersion));
         }
+
+        _isEditInProgress.OnNext(_editLevel is not 0);
```
dwcullop: If the update action throws, we never get back to line 222, so `_isEditInProgress` is left stuck on `true` for the life of the cache. Every `Connect()` after that takes the deferral path and waits for a `false` that is never coming, so the subscriber just sits there silently. No exception, no completion, it simply never emits.

A stuck `_editLevel` was survivable before this change. Now it takes out the primary API of the type:

```csharp
using var cache = new SourceCache<int, int>(static x => x);
using var warmup = cache.Connect().Subscribe();

try
{
    cache.Edit(inner => { inner.AddOrUpdate(1); throw new InvalidOperationException(); });
}
catch (InvalidOperationException) { }

var received = 0;
using var sub = cache.Connect().Subscribe(cs => received += cs.Adds);
// received is 0 here, was 1 before this change
```

`UpdateFromSource` has the same shape, and so does `SourceList.Edit` (where `InvokeNext` can also throw if a subscriber throws). The decrement and the reset both want to be in a `finally`. Worth a regression test on each, since the whole existing suite passes with this bug in place.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -115,14 +119,22 @@ public IObservable<IChangeSet<TObject, TKey>> Connect(Func<TObject, bool>? predi
...
-                        .Where(static b => !b).Take(1).Select(_ => CreateConnectObservable(predicate, suppressEmptyChangeSets)).Switch();
+                    : Observable.CombineLatest(
+                            _suspensionTracker.Value.NotificationsSuspendedObservable,
+                            _isEditInProgress,
+                            static (areNotificationsSuspended, isEditInProgress) => areNotificationsSuspended || isEditInProgress)
+                        .Do(static _ => { }, observer.OnCompleted)
+                        .Where(static shouldConnectionBeDeferred => !shouldConnectionBeDeferred)
+                        .Take(1)
+                        .Select(_ => CreateConnectObservable(predicate, suppressEmptyChangeSets))
+                        .Switch();
```
dwcullop: `_suspensionTracker` is `Lazy<>` on purpose, and the condition on 122 is careful to check `IsValueCreated` before touching `.Value` so it stays unrealised on caches that never suspend. Line 130 gives that up: reaching `NotificationsSuspendedObservable` forces the tracker into existence, and the deferral branch is now reachable from an ordinary edit rather than only from a real suspension.

The allocation isn't really the problem. It's that `IsValueCreated` latches true and never goes back, and both `EmitChanges` and `EmitCount` take `lock (_locker)` whenever it's set. So one mid-edit subscribe on a cache that never suspends buys a lock acquisition on every delivery for the rest of that cache's life.

If the edit deferral is driven by something that doesn't involve the tracker, the lazy stays lazy and the two concerns stay independent.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -181,6 +201,7 @@ internal void UpdateFromIntermediate(Action<ICacheUpdater<TObject, TKey>> update
         ChangeSet<TObject, TKey>? changes = null;
 
         _editLevel++;
+        _isEditInProgress.OnNext(_editLevel is not 0);
```
dwcullop: Two things on the hot path here. `_isEditInProgress` is allocated in both constructors whether or not anything ever subscribes mid-edit, and every `Edit()` now pays two `BehaviorSubject.OnNext` calls, each taking the subject's internal lock and snapshotting observers even when there are none. Every `AddOrUpdate` and `Remove` goes through this, so it's the busiest path in the library, and the case it's paying for (someone subscribing while an edit is open) is vanishingly rare.

`_countChanged` and `_suspensionTracker` are both `Lazy<>` in this same file for exactly this reason.

Worth noting that `Connect()` already holds `_locker`, and `_editLevel` is mutated under that same lock, so the "is an edit in progress" test doesn't need a subject at all: it can read `_editLevel` directly for free. The signalling is only needed to wake parked subscribers, which only matters when there are any. That would keep the common path to a null check and make the allocation conditional on someone actually subscribing mid-edit.

src/DynamicData/Cache/ObservableCache.cs
dwcullop: A subscriber that connects while an edit is in progress never receives `OnCompleted` when the source is later disposed. It activates fine and receives every changeset, it just never terminates, so anything using `Finally` for cleanup or awaiting completion waits forever. `SourceList` behaves the same way.

This isn't yours: the suspension deferral on `main` already loses completion the same way, and this reuses that shape. The difference is the trigger. Connecting during a suspension is unusual, connecting during an edit is not, so the PR takes a corner case and makes it routine.

Whatever the fix is, it probably belongs on both paths at once, and it might be worth splitting out separately rather than growing this PR.

dwcullop:
Filed this as #1136 so it doesn't get lost if this PR lands first.

src/DynamicData/Cache/ObservableCache.cs
dwcullop: Nested edits emit on every level, so a 3-deep nest sends `true` five times before the single `false`. Only the outermost 0 to 1 transition actually changes anything, and `BehaviorSubject` won't dedupe them for you. `if (_editLevel == 1)` on the way in and `if (_editLevel == 0)` on the way out would cut it to one each.

src/DynamicData/List/SourceList.cs
```diff
@@ -78,40 +83,32 @@ public SourceList(IObservable<IChangeSet<T>>? source = null)
...
+                return observable.SubscribeSafe(observer);
+            }
+        });
 
     /// <inheritdoc />
     public void Dispose()
     {
         _cleanUp.Dispose();
         _changesPreview.Dispose();
+        _isEditInProgress.OnCompleted();
```
dwcullop: `_cleanUp.Dispose()` already runs the private `OnCompleted()`, which completes `_isEditInProgress` under `_locker`. Line 111 does it again, this time without the lock. It's harmless because the second call is a no-op, but it reads as though the first one didn't happen, and it's the only place this subject is touched outside `_locker`.

src/DynamicData.Tests/Cache/SourceCacheFixture.cs
```diff
@@ -354,5 +355,55 @@ public void ConnectDuringDeliveryDoesNotDuplicate()
         addCounts.GetValueOrDefault("k2").Should().Be(1, "k2 should appear once, not duplicated from snapshot + queued delivery");
     }
 
+    // Covers https://github.com/reactivemarbles/DynamicData/issues/1129
+    [Fact]
+    public void ConnectDuringEditDoesNotDuplicate()
```
dwcullop: Good test, and it does pin the reported bug. Three gaps worth closing while this is fresh: `Watch()` got the same deferral treatment and has no coverage at all; nested edits aren't exercised, and they're the case most likely to go wrong; and both tests only add items, so the snapshot assertion would still pass if a removal leaked through. An edit that adds and removes would prove the deferred subscriber sees the true post-edit state rather than an intermediate one.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: Fixed.

But it's also worth noting that your version of the test actually represents a defect, IMO. On reflection, I think it's the right call to have errors during an edit lock out future subscribers. The collection itself still serves as a perfectly good, functional, collection. I probably need to make sure this behavior is implemented in VNext as well. But it doesn't make sense for a partial edit to be "committed". If the consumer's intention was for multiple, related, items to be added, but only some of them got added before the exception, then that could represent corrupt state in a business layer. I think it's something we should look into fixing, eventually.

The test I added simplifies the scenario: subscribers that follow a failed edit should still work as normal.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: The point about lock acquisition is somewhat compelling. I did assume it was all just about the allocation. Although I will point out that acquiring an uncontested lock isn't exactly an expensive operation. A little under 20ns, according to a basic benchmark. 

I think I've got it fixed, though. I also went ahead and checked in that benchmark.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: An `.OnNext()` call on a `BehaviorSubject<T>` with a single subscriber that doesn't send notifications through a large downstream query, is worth 17ns.

With some benchmark trickery, the approximate cost of a single `.AddOrUpdate()` (without contention, concurrency, internal re-allocations) is 242ns with these `.OnNext()` calls, and 202ns without them. A difference of 40ns, which does pretty closely align with the 17ns/each benchmark. So, roughly a 17% speed penalty.

I'd say we're prematurely micro-optimizing, but your point about `.Edit()` basically being the hottest path in the library is fair. It's certainly the most-used user-facing API. I'll see about lazifying the system, but it's gonna make that other fix above even MORE convoluted.

src/DynamicData/Cache/ObservableCache.cs
dwcullop: Also filed #1140. The same deferral converts a source failure into a successful completion, not just a lost completion: a deferred subscriber gets `OnCompleted` where an ordinary one gets `OnError`.

Relevant to this PR specifically: routing `_isEditInProgress.OnError` through the same `Do(..., observer.OnCompleted)` means an edit-deferred subscriber inherits that behaviour, which widens the trigger from suspensions to ordinary edits.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: Fixed.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: Let's take that one first then, and I'll make sure it gets properly incorporated here, when I update the branch.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: I went ahead and fixed this, on account of it being simple and straightforward, and we already now have a conditional on each one of these calls. But for the record, it shaves about 10ns off of 400ns operations, a 3% speed improvement. I'd definitely put this in the category of micro-optimization.

src/DynamicData/List/SourceList.cs
JakenVeina: I ended up with this after it was initially `_isEditInProgress.Dispose()` and that resulted in a bunch of `ObjectDisposedException`s across the test suite.

Swapped it for a clarification comment instead.

src/DynamicData.Tests/Cache/SourceCacheFixture.cs
JakenVeina: I can confirm I don't see any test coverage for `.Watch()` either. I'm gonna call that out-of-scope for this Issue/PR, and something we should get around to eventually, alongside the other probably 50% of the codebase that is inadequately tested.

Fair point for the other two suggestions. I've implemented them.

src/DynamicData/List/SourceList.cs
```diff
@@ -124,21 +124,28 @@ public void Edit(Action<IExtendedList<T>> updateAction)
...
+            finally
             {
-                _readerWriter.WriteNested(updateAction);
-            }
+                _editLevel--;
 
-            _editLevel--;
+                if (changes is not null && _editLevel == 0)
+                {
+                    InvokeNext(changes);
```
dwcullop, line 148: The `finally` is here now, but `InvokeNext` came inside it and sits ahead of the reset, so the hole never actually closed. `_changes.OnNext` delivers inline, so an existing subscriber that throws takes line 148 out on the way past and `_isEditInProgress` is stuck on `true` from then on. Every `Connect()` after that parks waiting for a `false` that isn't coming.

```csharp
using var list = new SourceList<int>();
using var thrower = list.Connect().Subscribe(static _ => throw new InvalidOperationException());

try { list.Add(1); } catch (InvalidOperationException) { }

var received = 0;
using var after = list.Connect().Subscribe(cs => received += cs.Adds);
// received is 0, and stays 0
```

The cache dodges this, and it's worth knowing why because it isn't obvious from the shape: `EnqueueNext` only queues, and delivery happens when the notification scope is disposed, which is after the `finally` has already run. Here the delivery is the last thing the `finally` does before the reset, which is the one ordering that doesn't survive a throw.

Moving the reset above `InvokeNext` is probably the whole fix. `_editLevel` is back to zero by that point, and all the `false` does is tell parked subscribers it's safe to look, so they'd be looking at the same state either way.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -256,6 +252,63 @@ private IObservable<IChangeSet<TObject, TKey>> CreateConnectObservable(Func<TObj
...
+
+                        // Edit system is active, suspension system isn't
+                        // Need to avoid activating the suspension system if we don't absolutely need to, as it adds
+                        // locking overhead to edit operations. But then when the edit is done, we need to check if a
+                        // suspension came in. If so, do a followup wait with the full logic for both systems. It
+                        // needs to be the full logic, in case an edit comes in during the suspension, and so on.
+                        (false, true) => _isEditInProgress.Value
+                            .Where(static isEditInProgress => !isEditInProgress)
+                            .Take(1)
+                            .Select(_ => CreateFullyDeferredConnection())
```
dwcullop: The comment above this branch has it exactly right, the code just doesn't do the second half of it. "When the edit is done, we need to check if a suspension came in. If so, do a followup wait" is the correct shape, but there's no check anywhere: `CreateFullyDeferredConnection` runs unconditionally, and the first thing it reaches for is `_suspensionTracker.Value`.

So on a cache that never suspends, one mid-edit `Connect()` still builds the tracker the moment the edit finishes, and `EmitChanges` and `EmitCount` go back to taking `_locker` on every delivery for the rest of that instance's life. Same thing mirrored in the branch below it, which realises `_isEditInProgress` and puts the `OnNext` pair back on every edit.

Re-testing the two `IsValueCreated` flags inside the `Select` should close it, falling through to `factory.Invoke()` when the other system still hasn't been touched.

To be clear, the path your 17% was actually measuring is fixed. This is just the tail case still leaking through.

src/DynamicData/List/SourceList.cs
```diff
@@ -124,21 +124,28 @@ public void Edit(Action<IExtendedList<T>> updateAction)
             IChangeSet<T>? changes = null;
 
             _editLevel++;
-
-            if (_editLevel == 1)
+            _isEditInProgress.OnNext(_editLevel is not 0);
```
dwcullop: The cache got the `Lazy<>` and the transition guards, the list got neither. `_isEditInProgress` is still built in the constructor for every `SourceList` anyone creates, and both `OnNext` calls still fire at every level, so the three-deep nest still sends five `true`s ahead of the one `false`.

I'm not going to pretend the numbers are big here, you were right about that the first time. It's more that the two types now behave differently on the same hot path for no reason a reader could work out from the code, and whoever is in here next gets to rediscover why. The `Lazy<>` and the `_editLevel == 1` / `== 0` guards should port across close to as-is, with `Connect()` reading `_editLevel` directly since it's already holding `_locker`.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -256,6 +252,63 @@ private IObservable<IChangeSet<TObject, TKey>> CreateConnectObservable(Func<TObj
...
+                            .Switch(),
+
+                        // If both systems are already active, we can monitor both systems simultaneously, and make the
+                        // connection as soon as both are idle at the same time.
+                        _ => CreateFullyDeferredConnection()
+                    };
+
+                return observable.SubscribeSafe(observer);
+
+                IObservable<T> CreateFullyDeferredConnection()
```
dwcullop, line 301: Went back through this against #1140 now that the structure has changed, and it holds up better than I expected. The edit path is fine: `CacheUpdateObserver.OnError` faults `_isEditInProgress` before it disposes the tracker, so `CombineLatest` sees the error first, the `Do` never gets a completion to forward, and a subscriber deferred by an edit ends up with `OnError` like everyone else.

The one still broken is the suspension-only branch above, where `_isEditInProgress` may not exist yet and the tracker's `Dispose` completes the subject instead of faulting it. That's #1140 unchanged rather than anything this PR made worse, so disregard my earlier comment about it widening the trigger, that was written against the old shape.

Worth carrying forward when #1140 lands though: the fix wants to be on the suspension subject itself, not on the deferral shape, since the deferral shape here is already doing the right thing.

src/DynamicData/List/SourceList.cs
JakenVeina: > so the hole never actually closed

I'd call this an entirely different hole: subscribers throwing is a bit different from directly-injected user logic throwing, but it amounts to the same thing. Fixed.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: I had the logic in there at one point, but it got lost somewhere in shuffling different fixes around. Fixed.

src/DynamicData/List/SourceList.cs
JakenVeina: Having implementation parity between `SourceCache` and `SourceList` is a lost cause, IMO, until we want to sit down and make a concentrated effort to overhaul both classes. Which we should do, eventually. But it's a fair point for the `Lazy<>` approach. I don't really have an argument to not do it now, given it's a pretty trivial change. I simply didn't because `SourceList` didn't already have any of that going on.

Fixed.

src/DynamicData/List/SourceList.cs
```diff
@@ -124,21 +124,29 @@ public void Edit(Action<IExtendedList<T>> updateAction)
...
+                }
+                else
+                {
+                    _readerWriter.WriteNested(updateAction);
+                }
 
-            if (changes is not null && _editLevel == 0)
+                if (changes is not null && (_editLevel is 1))
+                {
+                    InvokeNext(changes);
```
dwcullop, line 148: Moving `InvokeNext` inside the `try` means it now fires while `_editLevel` is still 1. If a subscriber edits the list in response, that edit takes the `WriteNested` path and throws, because the outer `Write` has already finished.

This works on main and throws here:

```csharp
list.Connect().Subscribe(_ => list.Edit(u => u.Add(999)));
list.Add(1);
// InvalidOperationException: WriteNested can only be used if another write is already in progress.
```

The flag flip does need to stay after `InvokeNext`, otherwise the deferred subscriber gets the snapshot and the changeset, which is #1129 all over again. So it's only the decrement that needs to move:

```csharp
try
{
    try
    {
        // ...Write or WriteNested...
    }
    finally
    {
        _editLevel--;
    }

    if (changes is not null && (_editLevel is 0))
    {
        InvokeNext(changes);
    }
}
finally
{
    if (_isEditInProgress.IsValueCreated && (_editLevel is 0))
        _isEditInProgress.Value.OnNext(false);
}
```

`ObservableCache` isn't affected, since `EnqueueNext` stayed after the decrement there.

src/DynamicData/List/SourceList.cs
```diff
@@ -78,40 +83,35 @@ public SourceList(IObservable<IChangeSet<T>>? source = null)
...
-                            });
-                    }
+                var observable = _isEditInProgress.IsValueCreated || (_editLevel is not 0)
 
-                    var source = _changes.Finally(observer.OnCompleted);
+                    // Defer connection until there is no longer an in-progress edit.
+                    ? _isEditInProgress.Value
+                        .Where(static isEditInProgress => !isEditInProgress)
+                        .Take(1)
+                        .Select(_ => CreateConnectObservable(predicate))
```
dwcullop: `CreateConnectObservable` returns `IObservable<IChangeSet<T>>`, so this `.Switch()` picks up `ObservableListEx.Switch` rather than the Rx one. That's the shadowing gotcha from #1136, and the effect is that a connection deferred by an edit never completes:

```csharp
list.Edit(_ => list.Connect().Subscribe(..., () => completed = true));
list.Dispose();  // completed is still false
```

`Take(1)` means there's only ever one inner sequence, so `SelectMany` does the same job and carries the terminal event through:

```csharp
.SelectMany(_ => CreateConnectObservable(predicate))
```

The cache path only escapes this because `CreateDeferredNotificationObservable<T>` is generic, so `T` can't bind to `IChangeSet`.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -256,6 +252,67 @@ private IObservable<IChangeSet<TObject, TKey>> CreateConnectObservable(Func<TObj
...
+                    => Observable.CombineLatest(
+                            _suspensionTracker.Value.NotificationsSuspendedObservable,
+                            _isEditInProgress.Value,
+                            static (areNotificationsSuspended, isEditInProgress) => areNotificationsSuspended || isEditInProgress)
+                        .Do(static _ => { }, observer.OnCompleted)
```
dwcullop, line 306: This is the shape #1141 replaced in `Connect`/`Watch`, and it's the bit that reports a failure as a successful completion. Since each of these is behind a `Take(1)`, `SelectMany` covers all three sites and the `.Do` can go:

```csharp
.Where(static shouldConnectionBeDeferred => !shouldConnectionBeDeferred)
.Take(1)
.SelectMany(_ => factory.Invoke());
```

Worth doing as part of the merge so #1141 doesn't get undone.

src/DynamicData.Benchmarks/Miscellaneous/LockImplementations.cs
```diff
@@ -0,0 +1,42 @@
+using System.Threading;
```
dwcullop, line 1: Was this meant to go in with this PR? Looks like it might be left over from checking the lock overhead.

dwcullop (changes requested):
Thanks for working through all of that, it's in much better shape. The lazy seeding with `_editLevel is not 0` is a neat touch, it keeps the mid-edit case correct without paying for the subject up front. Nested edits and the failed-edit recovery both have coverage now too.

Two things I'd like to sort out first, both in `SourceList`. Details inline.

One heads-up as well: this branch predates #1141, which reworked the same deferral in `Connect`/`Watch`. Merging main in will conflict there, so worth a careful resolution.

src/DynamicData/List/SourceList.cs
dwcullop: Still there on the current head. I reverted `SourceList.cs` and `ObservableCache.cs` to main and it passes, so it is coming from this branch rather than something that arrived with the merge.

Also, the fix I suggested in the older thread on this same line was wrong, sorry about that. Moving the reset above `InvokeNext` would have the deferred subscriber connect, take its snapshot, and then receive the changeset that snapshot already contains, which is #1129 again. Your ordering is the right one. It is only the decrement that needs to move:

```csharp
try
{
    try
    {
        // ...Write or WriteNested...
    }
    finally
    {
        _editLevel--;
    }

    if (changes is not null && (_editLevel is 0))
    {
        InvokeNext(changes);
    }
}
finally
{
    if (_isEditInProgress.IsValueCreated && (_editLevel is 0))
        _isEditInProgress.Value.OnNext(false);
}
```

Whole suite passes with that applied.

src/DynamicData/List/SourceList.cs
dwcullop: Withdrawing this one. #1139 went in a few hours ago and taught the list `Switch` to carry terminal events, so a connection deferred by an edit does complete now. I re-ran the case against this head and it passes.

Still resolves to `ObservableListEx.Switch` rather than the Rx one, which is worth a glance at some point, but there is no bug behind it any more and it is not worth holding this PR for.

src/DynamicData/Cache/ObservableCache.cs
dwcullop: Scratch this one, I got it wrong. I checked the actual behaviour and both deferred paths hand `OnError` to the subscriber and do not complete, so there is nothing to fix here.

What threw me was my own test: the warm-up subscription was a bare `.Subscribe()` with no error handler, so Rx rethrew and it looked like the failure was going missing. Once it had a handler the behaviour was correct. The real cause of #1140 was on the producing side and your `Fault()` from #1141 already deals with it, so `.Do(..., observer.OnCompleted)` is doing the right thing for both cases.

src/DynamicData.Benchmarks/Miscellaneous/LockImplementations.cs
dwcullop: Saw your note above that this is the benchmark behind the 17ns and 40ns figures. Makes sense, ignore the question.

dwcullop (changes requested):
Went back over the four from last time against the current head, now that the merge has brought in #1137, #1139 and #1141.

Three of them are gone. Two were mine to withdraw: the list `Switch` one is fixed by #1139 landing, and the deferral one I simply had wrong, which I have owned in the thread. The benchmark question you had already answered.

The re-entrancy one in `SourceList.Edit` is still live and I think it is the last thing standing. A subscriber that edits the list from a change handler throws, because `InvokeNext` now runs while `_editLevel` is still 1 and the nested edit takes the `WriteNested` path after the outer write has finished. It works on main, so it is coming from this branch. Suggested shape is in the thread, and the suite is green with it.

Everything else from the first pass looks good. The lazy seeding, the transition guards, `SourceList` getting the same treatment as the cache, and the extra coverage for nested edits and removals all read well.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: I think this is another permalink that's drifted, but what you describe should be good now.

src/DynamicData/List/SourceList.cs
JakenVeina: Ugh. I was trying to avoid having this much noise, but I also had a sneaking feeling that moving the `_editLevel` manipulation down wasn't going to be right. I don't see any other cleaner implementation.

Fixed.

src/DynamicData/List/SourceList.cs
JakenVeina: Fixed anyway. The `.SelectMany()` implementation is honestly just better.

src/DynamicData.Benchmarks/Miscellaneous/LockImplementations.cs
JakenVeina: Figured it was worth leaving the benchmark in, just for reference.

dwcullop (approved):

PR #1135, opened 2026-07-13
Feature: Cancellation support for TransformAsync

glen-nicol:
Following [this discussion](https://github.com/reactivemarbles/DynamicData/discussions/1126#discussioncomment-17551854) I decided to revert back to just an additional overload of TransformAsync instead of using TransferAsyncCancel. I removed the ambiguous overload. 

I decided it was best to extend the existing pattern of using the overload with the most parameters and modified existing operators to use the new overload that has the cancellation token parameter and just discarded that argument for the overloads with fewer parameters.

I added support to the List variant too and modified the internals on that one to avoid a cancelation of one from releasing the async lock of another in the same stream.

### DIFF
--- a/src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet9_0.verified.txt
+++ b/src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet9_0.verified.txt
@@ -1964,6 +1964,14 @@ namespace DynamicData
             where TDestination :  notnull
             where TSource :  notnull
             where TKey :  notnull { }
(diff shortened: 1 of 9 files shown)


JakenVeina (changes requested):
Looks alright. Not sure if I completely follow all the exception logic in the List version, but I suppose it makes sense to swallow exceptions related to cancellation. Kinda makes me think `Observable.FromAsync()` ought to be the one doing that.

I do need to see some tests added for new functionality, though, even if they're mostly performative. If you can add some, great, otherwise I'll get around to it eventually, assuming I've got write access to your branch.

glen-nicol:
k, I can appreciate that thought on cancellation. I think the Object disposed handler has to stay though since there is no guarantee that the transform function will return immediately and complete before the outer subscription of the Observable.Using disposes of the async lock.

I added a couple simple tests to ensure cancellation is connected to disposal.

JakenVeina (approved):

PR #1137, opened 2026-07-25
Fix that the cache Switch operator never completes

### LINKED ISSUE #1136: [Bug]: Deferred Connect() never receives OnCompleted when the source is disposed
### Describe the bug 🐞

A `Connect()` that gets deferred never receives `OnCompleted` when the source is later disposed.

`ObservableCache.Connect()` defers when notifications are suspended, rather than subscribing immediately:

```csharp
: _suspensionTracker.Value.NotificationsSuspendedObservable.Do(static _ => { }, observer.OnCompleted)
    .Where(static b => !b).Take(1).Select(_ => CreateConnectObservable(predicate, suppressEmptyChangeSets)).Switch();
```

Once the suspension lifts, the subscriber activates and works normally: it gets its initial changeset and every subsequent one. But it never terminates. Disposing the cache completes an ordinary subscriber and does nothing for a deferred one.

That leaks. Anything relying on `Finally` for cleanup, or awaiting completion, waits forever. `Watch()` has the same shape.

### Step to reproduce

```csharp
var log = "";
var observer = Observer.Create<IChangeSet<int, int>>(
    _ => log += "OnNext;",
    _ => log += "OnError;",
    () => log += "OnCompleted;");

var cache = new SourceCache<int, int>(static x => x);
using (cache.Connect().Subscribe()) { }

var suspend = cache.SuspendNotifications();
cache.AddOrUpdate(1);
cache.Connect().SubscribeSafe(observer);   // deferred by the suspension
suspend.Dispose();
cache.AddOrUpdate(2);
cache.Dispose();

// log is "OnNext;OnNext;" - the OnCompleted never arrives.
// Connecting outside a suspension gives "OnNext;OnCompleted;"
```

### Expected behavior

A deferred subscriber should be indistinguishable from an ordinary one once it activates, including receiving `OnCompleted` (or `OnError`) when the source terminates.

### DynamicData Version

main

### Additional information ℹ️

Found while reviewing #1133, which reuses this deferral shape to defer subscriptions made during an `Edit()`. That PR is not the cause, but it does turn this from a corner case into a routine one: connecting during a suspension is unusual, connecting during an edit is not. `SourceList` picks up the same behavior there.

Errors are unaffected, `OnError` propagates correctly on both paths. It is specifically completion that is lost.

Probably worth fixing on the suspension path here so #1133 inherits the fix rather than duplicating the gap.

dwcullop:
Fixes #1136.

The cache `Switch` operator could never complete, could retain data from a source it had already switched away from, and routed delivery through a lock.

### Never completing

It relayed changes through a private `LockFreeObservableCache`:

```csharp
var destination = new LockFreeObservableCache<TObject, TKey>();
...
.PopulateInto(destination);

destination.Connect().Merge(errors).SubscribeSafe(observer)
```

A relay only ends when it is disposed, so the terminal event of the source had nowhere to go. Errors were carried across by hand through the merged `errors` subject; completion had no equivalent path and was silently dropped. A consumer of this operator never received `OnCompleted`, ever.

### The lock

The obvious fix is to let `Observable.Switch` sit between the sources and the observer and propagate terminal events on its own. That is wrong here, and it took a measurement to see why.

`Observable.Switch` holds its gate for the whole of the downstream `OnNext` call. A pipeline that crosses into another cache runs that work under the gate, and a producer on another thread blocks behind it. That is precisely the cross cache deadlock shape the delivery queue exists to avoid, and the operator on `main` avoids it today by keeping the observer on the far side of the relay cache.

Measured with a subscriber that blocks inside `OnNext`, writing to the source from another thread:

| | time to return |
|---|---|
| `main` | 0 ms |
| via `Observable.Switch` | 748 ms |
| this PR | 1 ms |

So the operator is now written by hand. A `SerialDisposable` holds the current inner subscription, and every subscription carries an identity, so a superseded source still delivering concurrently is dropped rather than applied on top of the state its replacement has already established. That is the second defect above, fixed by construction rather than by timing. All state changes and all delivery go through the queue, which releases its lock before handing anything downstream, so a second producer enqueues and returns instead of waiting.

The reset that clears the previous source's contribution is enqueued under the same lock acquisition that takes the new identity, which keeps it ordered against the changes that follow it.

### Terminal semantics

Completes once the sources and the current inner have both completed. Fails as soon as either does. That matches `Observable.Switch`, which is what callers reasonably expect from something called Switch.

### Tests

`Cache/SwitchFixture`:

- `CompletesWhenSourcesAndInnerComplete`
- `DoesNotCompleteWhileInnerIsStillRunning`
- `DoesNotCompleteWhenOnlyASupersededInnerCompletes`
- `CompletesWhenSourcesAndInnerCompleteSynchronously`
- `DeliversChangesEmittedBeforeSynchronousCompletion`
- `PropagatesInnerErrorsRaisedSynchronously`
- `IgnoresChangesFromASupersededSource`
- `DoesNotHoldALockWhileDeliveringDownstream`

That last one blocks a subscriber inside `OnNext` and asserts that writing to the source from another thread still returns. It fails against an `Observable.Switch` implementation and passes against this one, so the lock behaviour cannot quietly come back.

`SuspendNotificationsFixture`, covering #1136 directly:

- `OnCompletedFiresIfCacheDisposedAfterConnectingWhileSuspended`
- `OnErrorFiresIfCacheFailsAfterConnectingWhileSuspended`
- `OnCompletedFiresIfCacheDisposedAfterWatchingWhileSuspended`

### Notes

Rebased onto current `main` and squashed. The branch previously carried a workaround in `ObservableCache.Connect()` that was later reverted in favour of fixing the operator, and replaying that history over #1132 conflicted repeatedly for no benefit. Nothing outside `Switch` and its tests is touched now.

The list `Switch` has the same three defects and is fixed the same way in #1139. The two are independent.

Behavioural change to a public operator: sequences that previously never terminated now terminate when their sources do. `main` is on `10.0-preview`.

Unrelated, and not addressed here: `AggregationEx.InvalidateWhen` still uses Rx's `Switch` and therefore still holds a gate during delivery. It operates on plain value streams rather than changesets, so the blast radius is smaller, but it is the same pattern if anyone is worried about it.

### DIFF
--- a/src/DynamicData.Tests/Cache/SuspendNotificationsFixture.UnitTests.cs
+++ b/src/DynamicData.Tests/Cache/SuspendNotificationsFixture.UnitTests.cs
@@ -1,7 +1,8 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reactive.Linq;
+using System.Reactive.Subjects;
 using System.Threading.Tasks;
 
 using FluentAssertions;
@@ -451,6 +452,67 @@ public static partial class SuspendNotificationsFixture
             results.Data.Count.Should().Be(allData.Count, $"{allData.Count} items in final state");
             results.Error.Should().BeNull();
             results.IsCompleted.Should().BeFalse();
+        }
+
+        [Fact]
+        public void OnCompletedFiresIfCacheDisposedAfterConnectingWhileSuspended()
+        {
+            // A connection made while suspended is deferred until the suspension lifts. Once it
+            // activates it must behave exactly like any other subscriber, including terminating
+            // when the source does. Previously the deferral dropped the completion, leaving the
+            // subscriber alive forever.
+            var suspend = _source.SuspendNotifications();
+            using var results = _source.Connect().AsAggregator();
+            Enumerable.Range(101, 37).ForEach(_source.AddOrUpdate);
+
+            // Act
+            suspend.Dispose();
+            _source.AddOrUpdate(1000);
+            _source.Dispose();
+
+            // Assert
+            results.IsCompleted.Should().BeTrue("a connection deferred by a suspension should still complete when the source does");
+            results.Error.Should().BeNull("no error should have occurred");
+            results.Data.Count.Should().Be(38, "all data written before disposal should have arrived");
+        }
+
+        [Fact]
+        public void OnErrorFiresIfCacheFailsAfterConnectingWhileSuspended()
+        {
+            // The same applies to the error case: a deferred connection that has activated must
+            // still see a failure of the source.
+            using var source = new Subject<IChangeSet<int, int>>();
+            using var cache = new ObservableCache<int, int>(source);
+
+            var suspend = cache.SuspendNotifications();
+            using var results = cache.Connect().AsAggregator();
+            source.OnNext(new ChangeSet<int, int> { new(ChangeReason.Add, 1, 1) });
+
+            // Act
+            suspend.Dispose();
+            var expectedError = new Exception("Test Exception");
+            source.OnError(expectedError);
+
+            // Assert
+            results.Error.Should().Be(expectedError, "a connection deferred by a suspension should still see the source fail");
+            results.Data.Count.Should().Be(1, "the data written before the failure should have arrived");
+        }
+
+        [Fact]
+        public void OnCompletedFiresIfCacheDisposedAfterWatchingWhileSuspended()
+        {
+            // Watch() defers the same way Connect() does, and has the same obligation.
+            var suspend = _source.SuspendNotifications();
+            var isCompleted = false;
+            using var subscription = _source.Watch(1).Subscribe(static _ => { }, () => isCompleted = true);
+            _source.AddOrUpdate(1);
+
+            // Act
+            suspend.Dispose();
+            _source.Dispose();
+
+            // Assert
+            isCompleted.Should().BeTrue("a watch deferred by a suspension should still complete when the source does");
         }
 
         public void Dispose()
--- a/src/DynamicData.Tests/Cache/SuspendNotificationsFixture.UnitTests.cs
+++ b/src/DynamicData.Tests/Cache/SuspendNotificationsFixture.UnitTests.cs
@@ -455,28 +455,6 @@ public static partial class SuspendNotificationsFixture
         }
 
         [Fact]
-        public void OnCompletedFiresIfCacheDisposedAfterConnectingWhileSuspended()
-        {
-            // A connection made while suspended is deferred until the suspension lifts. Once it
-            // activates it must behave exactly like any other subscriber, including terminating
-            // when the source does. Previously the deferral dropped the completion, leaving the
-            // subscriber alive forever.
-            var suspend = _source.SuspendNotifications();
-            using var results = _source.Connect().AsAggregator();
-            Enumerable.Range(101, 37).ForEach(_source.AddOrUpdate);
-
-            // Act
-            suspend.Dispose();
-            _source.AddOrUpdate(1000);
-            _source.Dispose();
-
-            // Assert
-            results.IsCompleted.Should().BeTrue("a connection deferred by a suspension should still complete when the source does");
-            results.Error.Should().BeNull("no error should have occurred");
-            results.Data.Count.Should().Be(38, "all data written before disposal should have arrived");
-        }
-
-        [Fact]
         public void OnErrorFiresIfCacheFailsAfterConnectingWhileSuspended()
         {
             // The same applies to the error case: a deferred connection that has activated must
--- a/src/DynamicData/Cache/Internal/Switch.cs
+++ b/src/DynamicData/Cache/Internal/Switch.cs
@@ -70,7 +70,17 @@ internal sealed class Switch<TObject, TKey>(IObservable<IObservable<IChangeSet<T
                                     scope.EnqueueNext(changes);
                                 }
                             },
-                            queue.OnError,
+                            error =>
+                            {
+                                using var scope = queue.AcquireLock();
+
+                                if (id != active)
+                                {
+                                    return;
+                                }
+
+                                scope.EnqueueError(error);
+                            },
                             () =>
                             {
                                 using var scope = queue.AcquireLock();
--- a/src/DynamicData/Cache/Internal/Switch.cs
+++ b/src/DynamicData/Cache/Internal/Switch.cs
@@ -28,18 +28,18 @@ internal sealed class Switch<TObject, TKey>(IObservable<IObservable<IChangeSet<T
 
                 // Identifies the current source. A superseded one may still be mid-delivery, and anything
                 // it produces after this point belongs to a source that has already been switched away from.
-                var active = 0;
+                var activeSourceId = 0;
                 var isSourceRunning = false;
                 var areSourcesComplete = false;
 
-                var outer = _sources.Subscribe(
+                var outer = _sources.SubscribeSafe(
                     source =>
                     {
-                        int id;
+                        int sourceId;
 
                         using (var scope = queue.AcquireLock())
                         {
-                            id = ++active;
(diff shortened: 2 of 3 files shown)


src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -21,25 +22,32 @@ public IObservable<IChangeSet<TObject, TKey>> Run() => Observable.Create<IChange
 
                 var destination = new LockFreeObservableCache<TObject, TKey>();
 
-                var errors = new Subject<IChangeSet<TObject, TKey>>();
+                // The destination cache only ends when disposed, so terminal events
+                // from the source have to be carried across.
+                var terminal = new Subject<Unit>();
```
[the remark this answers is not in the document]
dwcullop: This was a real bug. `PopulateInto` subscribes before the output does, so a source that terminates synchronously during that subscription signalled completion with nothing listening. A plain `Subject` replays a terminal error to late subscribers but not an `OnNext`, which is why the error case worked and completion did not: `TakeUntil` saw only `OnCompleted`, which does not stop it.

Switched to `AsyncSubject<Unit>` as suggested. Added `CompletesWhenSourcesAndInnerCompleteSynchronously` and `PropagatesInnerErrorsRaisedSynchronously`, both against `Observable.Return(Observable.Empty/Throw(...))` so everything terminates during subscription. The completion one fails with the plain `Subject`.

Also reordered the `CompositeDisposable` so the terminal subject is disposed after the producer and the subscriber, and added a `Catch` after the terminal subject is notified so the error never reaches `PopulateInto`, which subscribes without an error handler.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -17,30 +16,103 @@ internal sealed class Switch<TObject, TKey>(IObservable<IObservable<IChangeSet<T
...
+                // crossing into another cache cannot deadlock against it.
+                var queue = new DeliveryQueue<IChangeSet<TObject, TKey>>(observer);
+
+                // What the current source has contributed, so that switching away can take it back out.
+                var current = new Cache<TObject, TKey>();
+                var subscription = new SerialDisposable();
+
+                // Identifies the current source. A superseded one may still be mid-delivery, and anything
+                // it produces after this point belongs to a source that has already been switched away from.
+                var active = 0;
```
JakenVeina: Recommend a clearer name. `activeSourceId` maybe, with `id` renamed to `sourceId` to match.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -17,30 +16,103 @@ internal sealed class Switch<TObject, TKey>(IObservable<IObservable<IChangeSet<T
...
+                                scope.EnqueueNext(new ChangeSet<TObject, TKey>(
+                                    current.KeyValues.Select(static pair => new Change<TObject, TKey>(ChangeReason.Remove, pair.Key, pair.Value))));
+
+                                current.Clear();
+                            }
+                        }
+
+                        // Subscribed outside the lock. The source may deliver synchronously, and that
+                        // delivery takes the lock for itself.
+                        subscription.Disposable = source.Subscribe(
```
JakenVeina: All `.Subscribe()`s need to be `.SubscribeSafe()`.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -17,30 +16,103 @@ internal sealed class Switch<TObject, TKey>(IObservable<IObservable<IChangeSet<T
...
+
+                        using (var scope = queue.AcquireLock())
+                        {
+                            id = ++active;
+                            isSourceRunning = true;
+
+                            if (current.Count != 0)
+                            {
+                                scope.EnqueueNext(new ChangeSet<TObject, TKey>(
+                                    current.KeyValues.Select(static pair => new Change<TObject, TKey>(ChangeReason.Remove, pair.Key, pair.Value))));
```
JakenVeina, line 48: I think this is, slightly, the wrong approach. That is, enqueueing a `Clear` separately from the subscription to the new source. It breaks our general rule of "one frame, one changeset". If subscribing to the new source causes an immediate publication of the items already in that source (which it should), those should get appended to the same changeset as the removals from the old source.

It looks like the old implementation has this same defect, so I'm good with splitting that out into a separate issue.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -17,30 +16,103 @@ internal sealed class Switch<TObject, TKey>(IObservable<IObservable<IChangeSet<T
...
+
+                        // The current source may still be running, and the result ends only once both have.
+                        if (!isSourceRunning)
+                        {
+                            scope.EnqueueCompleted();
+                        }
+                    });
+
+                // Queue first, so that delivery is finished before the subscriptions are torn down.
+                return new CompositeDisposable(queue, outer, subscription);
```
JakenVeina: If disposal order is important, be explicit about it. Use a `Disposable.Create()` closure instead. `CompositeDisposable` doesn't guarantee disposal order. In fact, it can dispose items in an order other than the order they were added, if enough items get added to make it transition to a `Dictionary<>` instead of a `List<>` for internal storage.

src/DynamicData.Tests/Cache/SuspendNotificationsFixture.UnitTests.cs
```diff
@@ -453,6 +454,67 @@ public void ReSuspendThenResumeDeliversAllInSingleBatch()
             results.IsCompleted.Should().BeFalse();
         }
 
+        [Fact]
+        public void OnCompletedFiresIfCacheDisposedAfterConnectingWhileSuspended()
```
JakenVeina: This becomes redundant if #1145 gets merged first. I think the test over there is more thorough.

JakenVeina (changes requested):

src/DynamicData/Cache/Internal/Switch.cs
dwcullop: It won't transition to Dictionary here because there are only 3, so the order will be what I give it.  I guess it's not ideal to assume such a detail, but we do it in a lot of places.

src/DynamicData/Cache/Internal/Switch.cs
dwcullop: Done, both renamed. Reads much better.

src/DynamicData/Cache/Internal/Switch.cs
dwcullop: Good catch, fixed. Both the outer and the inner subscription use it now, so a source that throws out of its subscribe call gets reported through the queue instead of escaping to whoever happened to be subscribing at the time.

src/DynamicData/Cache/Internal/Switch.cs
dwcullop: Fair point, and relying on unspecified ordering was sloppy of me given the comment right above it says the order matters. Swapped for a Disposable.Create closure so it is stated rather than assumed.

src/DynamicData/Cache/Internal/Switch.cs
dwcullop: Agreed, and the flicker is the visible part of it: the collection empties and then refills instead of moving in one step. Filed as #1150, covering both cache and list, and left the two changesets alone here.

Confirmed it predates this PR. The version on main does destination.Clear() in a Do on the outer sequence before PopulateInto sees anything from the new source, so it splits the frame the same way.

src/DynamicData.Tests/Cache/SuspendNotificationsFixture.UnitTests.cs
dwcullop: Dropped it. That test does cover the resume path better.

Kept the other two, since they cover ground #1145 does not: one is the Watch path rather than Connect, and the other is the failure path rather than completion. If either turns out to overlap with what lands in #1141 they can go too.

JakenVeina (approved):

Issue #1140, opened 2026-07-26
[Bug]: A source failure is reported as successful completion to subscriptions deferred by SuspendNotifications


dwcullop:
### Describe the bug 🐞

If a cache's source fails while notifications are suspended, any `Connect()` or `Watch()` subscription that was deferred by that suspension is told the sequence **completed successfully**. The error is never delivered.

`ObservableCache.Connect()` defers when notifications are suspended, waiting for the suspension to lift:

```csharp
: _suspensionTracker.Value.NotificationsSuspendedObservable.Do(static _ => { }, observer.OnCompleted)
    .Where(static b => !b).Take(1).Select(_ => CreateConnectObservable(predicate, suppressEmptyChangeSets)).Switch();
```

The second argument to `Do` maps completion of the suspension subject directly onto `observer.OnCompleted`. When the source errors, `CacheUpdateObserver.OnError` calls `_suspensionTracker.Value.Dispose()`, and that does:

```csharp
public void Dispose()
{
    _areNotificationsSuspended.OnCompleted();
    _areNotificationsSuspended.Dispose();
}
```

So the failure path completes the subject, and every deferred subscriber reads that as a successful end of stream.

`Watch()` has the same shape and the same problem.

This is worse than losing the error. The subscriber's error handling never runs, `Finally`/`Catch` see a normal completion, and the consumer is left believing it holds a complete, valid dataset when the source actually failed.

### Step to reproduce

```csharp
var source = new Subject<IChangeSet<int, int>>();
using var cache = new IntermediateCache<int, int>(source);

var normal = "";
using var normalSub = cache.Connect().Subscribe(
    _ => normal += "OnNext;", _ => normal += "OnError;", () => normal += "OnCompleted;");

var deferred = "";
using (cache.SuspendNotifications())
{
    cache.Connect().Subscribe(
        _ => deferred += "OnNext;", _ => deferred += "OnError;", () => deferred += "OnCompleted;");

    source.OnError(new InvalidOperationException("source failed"));
}

// normal   == "OnError;"
// deferred == "OnCompleted;"   <-- the failure was reported as success
```

Same source, same failure, opposite terminal event, decided purely by whether the subscriber happened to connect during a suspension.

`Watch()` behaves identically:

```csharp
using (cache.SuspendNotifications())
{
    cache.Watch(1).Subscribe(_ => { }, _ => saw = "OnError", () => saw = "OnCompleted");
[...]

PR #1145, opened 2026-07-27
Fixed incorrect usage of `ObservableCacheEx.Switch()` instead of `Observable.Switch()`

### LINKED ISSUE #1136: [Bug]: Deferred Connect() never receives OnCompleted when the source is disposed

JakenVeina:
...within ObservableCache<TObject, TKey>.Connect()`, which was causing completion notifications to be dropped, and hurting performance.

Fixes #1136.

@dwcullop it looks like you already correctly identified that the root issue is in `ObservableCacheEx.Switch()`, but that operator wasn't the intention to be used here in the first place, so it's still worth swapping to the native `Observable.Switch()`.

### DIFF
--- a/src/DynamicData.Tests/Cache/SuspendNotificationsFixture.UnitTests.cs
+++ b/src/DynamicData.Tests/Cache/SuspendNotificationsFixture.UnitTests.cs
@@ -9,6 +9,7 @@ using FluentAssertions;
 using Xunit;
 
 using DynamicData.Kernel;
+using DynamicData.Tests.Utilities;
 
 namespace DynamicData.Tests.Cache;
 
@@ -44,6 +45,41 @@ public static partial class SuspendNotificationsFixture
             _results.Messages.Count.Should().Be(0, "Should have no item updates");
             _results.Data.Count.Should().Be(0, "Should not receive data after suspend");
             _results.IsCompleted.Should().BeFalse("IsCompleted should not have fired");
+        }
+
+        // https://github.com/reactivemarbles/DynamicData/issues/1136
+        [Fact]
+        public void SuspendedConnectCompletesCorrectly()
+        {
+            using var source = new SourceCache<int, int>(static item => item);
+        
+            using var suspension = source.SuspendNotifications();
+        
+            source.AddOrUpdate(1);
+        
+            using var subscription = source.Connect()
+                .RecordCacheItems(out var results);
+
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.RecordedChangeSets.Should().BeEmpty("notifications should have been suspended");
+
+            suspension.Dispose();
+
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.RecordedChangeSets.Should().ContainSingle("notifications should have been resumed");
+            results.RecordedItemsByKey.Should().BeEquivalentTo(source.KeyValues, "all changes should have propagated to the new subscriber");
+
+            source.AddOrUpdate(2);
+        
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.RecordedChangeSets.Skip(1).Count().Should().Be(1, "a single additional source operation was performed");
+            results.RecordedItemsByKey.Should().BeEquivalentTo(source.KeyValues, "all changes should have propagated to the new subscriber");
+
+            source.Dispose();
+        
+            results.Error.Should().BeNull("no errors should have occurred");
+            results.RecordedChangeSets.Skip(2).Should().BeEmpty("no additional source operations were performed");
+            results.HasCompleted.Should().BeTrue("the source has been disposed");
         }
 
         [Fact]
@@ -510,22 +546,6 @@ public static partial class SuspendNotificationsFixture
 
             results.Error.Should().Be(expectedError, "an activated connection should still see the source fail");
             results.Data.Count.Should().Be(1, "the data written before the failure should have arrived");
-        }
-
-        [Fact]
-        public void OnCompletedFiresIfCacheDisposedAfterResumingWhileWatchWasSuspended()
-        {
-            // The tests above cover failure. Completion has to reach an activated watch too, and
-            // this covers the path through the watch itself rather than through the suspension gate.
-            var suspend = _source.SuspendNotifications();
-            var isCompleted = false;
-            using var subscription = _source.Watch(1).Subscribe(static _ => { }, () => isCompleted = true);
-            _source.AddOrUpdate(1);
-
-            suspend.Dispose();
-            _source.Dispose();
-
-            isCompleted.Should().BeTrue("a watch deferred by a suspension should still complete when the source does");
         }
 
         public void Dispose()
--- a/src/DynamicData.Tests/Cache/SuspendNotificationsFixture.cs
+++ b/src/DynamicData.Tests/Cache/SuspendNotificationsFixture.cs
@@ -1,4 +1,3 @@
 ﻿namespace DynamicData.Tests.Cache;
 
 public static partial class SuspendNotificationsFixture;
-


dwcullop (commented):
Diagnosis looks right, and swapping to the native operator here is worth doing regardless of what happens to the operator itself. Two things about the diff, and then a note about ordering, because this collides with two open PRs.

**Watch() still ends in `.Switch()`**

`Connect()` gets the explicit `Observable.Switch(...)` and the comment explaining why, but `Watch()` keeps the fluent `.Switch()` a few lines down. That one is actually fine: its element type is `IObservable<Change<TObject, TKey>>`, and `Change<TObject, TKey>` is not a changeset, so it matches neither of the DynamicData overloads and already resolves to Rx's. So there is no bug there.

The problem is that the diff now has one call written explicitly and an adjacent one written fluently, with a comment on the first saying the explicit form is necessary. That reads like the second was missed. Either make both explicit, or say in the comment that `Watch` is unaffected because its element type is not a changeset. Otherwise someone will come along and "finish the job", or worse, assume the fluent form is safe elsewhere.

**`Do(static _ => { }, observer.OnCompleted)` is still there**

That is the other half of the problem, and it is #1140. The two argument `Do` has no error handler, and the gate is completed rather than faulted when the source fails: `CacheUpdateObserver.OnError` disposes the suspension tracker, which completes `NotificationsSuspendedObservable` normally. A subscriber that connected while notifications were suspended is therefore told the sequence succeeded when it actually failed. Swapping the `Switch` call does not change that.

**Ordering**

This overlaps two open PRs, both of which touch these exact lines.

#1141 fixes #1140 by dropping the `Do` and the `Switch` together:

```csharp
.Where(static areNotificationsSuspended => !areNotificationsSuspended)
.Take(1)
.SelectMany(_ => CreateConnectObservable(predicate, suppressEmptyChangeSets));
```

`Take(1)` means there is only ever one inner sequence, so `SelectMany` behaves identically to `Switch` but carries the gate's terminal event through on its own. No manual `observer.OnCompleted` call from inside a `Do`, and no `Switch` call to resolve to the wrong overload in the first place. It is structurally immune rather than carefully correct, and it fixes the error path at the same time.

#1137 fixes the `Switch` operator itself, which is a separate defect: any caller doing `.Switch()` on a changeset stream never receives `OnCompleted`, not just `ObservableCache`.

All three are worth having, but they need an order. Simplest is to take #1141 first, since its version of these lines supersedes the change here, and then this PR reduces to the regression test, which is the part worth keeping either way. `SuspendedConnectCompletesCorrectly` covers the resume path more thoroughly than anything in #1141.

Last thing: both this and #1137 say they fix #1136. Whichever merges first will close it, and the operator bug should stay tracked, since it affects anyone calling `.Switch()` on a changeset stream. Worth splitting that into its own issue so #1137 does not end up looking orphaned.

dwcullop:
Actually, I don't think we should be using Observable.Switch at all because it holds a lock.  Not sure if there's a real deadlock risk though.  Just thinking...

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -121,8 +121,15 @@ public IObservable<IChangeSet<TObject, TKey>> Connect(Func<TObject, bool>? predi
                     ? CreateConnectObservable(predicate, suppressEmptyChangeSets)
 
                     // Defer until notifications are no longer suspended
-                    : _suspensionTracker.Value.NotificationsSuspendedObservable.Do(static _ => { }, observer.OnCompleted)
-                        .Where(static b => !b).Take(1).Select(_ => CreateConnectObservable(predicate, suppressEmptyChangeSets)).Switch();
+                    // Direct call to Observable.Switch() is necessary, to avoid resolving to ObservableCacheEx.Switch()
+                    // which is not the intention, and would be overkill. It includes a lot of logic for calculating
+                    // item diffs between different subscriptions, which we don't need here as we're only dealing with
+                    // a single subscription.
+                    : Observable.Switch(_suspensionTracker.Value.NotificationsSuspendedObservable
```
dwcullop: Nice find on the overload resolution. That `ObservableCacheEx.Switch` wins over `Observable.Switch` for this shape isn't something you'd catch reading the line, and it explains the behaviour in #1136 neatly.

I think we only need one of this and #1141 though, and #1141 is probably the one to take. It rewrites this expression to use `SelectMany` instead, so `Switch` drops out of the call site completely and there's no overload left to resolve. It also drops the `Do(..., observer.OnCompleted)` on the next line, which is #1140: a source failure currently reaches a deferred subscriber as a successful completion, so their error handling never runs. Taking #1141 covers both, whereas this one fixes the completion path and leaves that second problem in place.

Worth adding that #1137 repairs `ObservableCacheEx.Switch` itself, so #1136 is handled for other callers regardless of what happens here.

Your diagnosis is still worth keeping, so I've filed it separately as #1148. `ObservableListEx.Switch` has the same signature, so it's a trap on the list side too, and it seemed a shame to lose that once this expression changes.

JakenVeina:
> The problem is that the diff now has one call written explicitly and an adjacent one written fluently, with a comment on the first saying the explicit form is necessary. That reads like the second was missed. Either make both explicit, or say in the comment that Watch is unaffected because its element type is not a changeset. Otherwise someone will come along and "finish the job", or worse, assume the fluent form is safe elsewhere.

I get the point you're making, but I disagree that it's an issue worth writing extraneous code for. The fact is, 99% of the time, the `ObservableCacheEx.Switch()` implementation is the one that SHOULD be used. `.Connect()` represents a singularly-unique exception that deserves explanation. `.Watch()` doesn't.

It's also a moot point, because #1133 extracts the shared logic between `.Connect()` and `.Watch()` and eliminates the issue entirely.

> That is the other half of the problem, and it is https://github.com/reactivemarbles/DynamicData/issues/1140.

If there's already another issue for this, then we'll deal with it there. Re-stating the issue here just ends up as noise.

> Simplest is to take https://github.com/reactivemarbles/DynamicData/pull/1141 first, since its version of these lines supersedes the change here, and then this PR reduces to the regression test, which is the part worth keeping either way. SuspendedConnectCompletesCorrectly covers the resume path more thoroughly than anything in https://github.com/reactivemarbles/DynamicData/pull/1141.

Works for me.

dwcullop (approved):
LGTM!

PR #1151, opened 2026-07-28
Move to project-level usings

dwcullop:
All three projects now declare their common namespaces once in the csproj instead of repeating them at the top of every file, and the settings shared by all three sit in `Directory.Build.props`.

The rule I applied: anything imported in more than a handful of files goes global. What is left in source files is only the four opt-in DynamicData namespaces:

| namespace | files |
| --- | --- |
| `DynamicData.Aggregation` | 7 |
| `DynamicData.PLinq` | 2 |
| `DynamicData.Experimental` | 1 |
| `DynamicData.Alias` | 1 |

Those stay local deliberately. They change what the operators in scope mean, so a reader should be able to see at the top of the file that `Select` is DynamicData's alias rather than LINQ's. Globalising them would apply that everywhere and hide it. `DynamicData.Diagnostics` had no such conflict, so it went global with the rest.

## What moved to Directory.Build.props

`ImplicitUsings` was set identically in all three projects, and nine namespaces appeared in all three `Using` blocks: `DynamicData.Binding`, `DynamicData.Kernel`, `System.Collections.ObjectModel`, `System.ComponentModel`, `System.Reactive.Disposables`, `System.Reactive.Linq`, `System.Reactive.Subjects`, `System.Reflection`, and `System.Runtime.CompilerServices`. Both are now declared once.

Anything needed by only one or two projects stays in the csproj that needs it. That includes `DynamicData.Internal` and `DynamicData.Cache.Internal`, which the library and tests share but which have no business in the benchmarks project, since it has no `InternalsVisibleTo`.

A few redundant settings went away while consolidating: `GenerateDocumentationFile` and `IsPackable=true` in `DynamicData.csproj`, and `IsPackable=false` and `PlatformTarget=AnyCPU` in the benchmarks, all already covered by the props file or the SDK default.

Worth flagging separately: the benchmarks `NoWarn` was missing its `$(NoWarn)` prefix, so it replaced the inherited list rather than appending to it. It relisted most of what it dropped, which is why nothing ever surfaced, but it was losing `VSX1000` and `CA1510`. It appends now.

## Collisions that needed a source change

Three places needed a real source change, because pulling namespaces into global scope made a name ambiguous where it had not been before.

- **`Notification<T>`** exists in both `System.Reactive` and `DynamicData.Internal`. The three test utilities that want the Rx one now say so explicitly.
- **`ToHashSet()`** is offered by both `DynamicData.Kernel.EnumerableEx` and `System.Linq.Enumerable`. Six calls in `WhenPropertyChangedRaceFixture` now go through `Enumerable.ToHashSet` directly.
- **`Person`** ships with `Bogus` and collides with `Domain.Person`. Rather than leave `using Bogus;` file-local across 26 test files, the tests project pins `Person` to the domain type with a global alias, which also retired the two existing file-level `Person =` aliases.

## Checks

No behaviour changes. I built `main` clean and diffed the warnings against this branch: both produce the same 399 distinct warnings, nothing added or removed. The full solution builds on all seven target frameworks and the whole test suite passes, including the public API approval tests.

This touches nearly 700 files, so it will conflict with essentially anything else in flight. Happy to hold it until the queue is clearer, or rebase whenever suits.

## The convention is written down

`.github/copilot-instructions.md` gains a Usings and Namespaces section covering where the declarations live, how to add one, and why the four opt-in namespaces are the exception. The exception is spelled out per namespace rather than as a blanket rule, since the reasons differ. `PLinq` is the sharp one, providing `Filter`, `Transform`, `TransformSafe`, and `SubscribeMany` under the same names as the core operators but with parallel implementations. `Experimental` shadows nothing at all; it stays local because the API is unstable and a call site should have to opt in visibly.

The three collisions above are recorded there too, so the fix is to qualify the call site rather than pull the namespace back into the file.


src/Directory.Build.props
```diff
@@ -36,6 +37,19 @@
     <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
   </PropertyGroup>
 
+  <!-- Namespaces used by every project. Project specific ones live in the csproj that needs them. -->
```
JakenVeina, line 40: I'll stand by what I said in the Primitives PR: I only mildly dislike the idea of global usings in general, but I VERY much dislike the idea of moving code into config files. I would much rather see these as `global using` statements in `.cs` files, for each project.

src/DynamicData.Benchmarks/DynamicData.Benchmarks.csproj
```diff
@@ -1,13 +1,20 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net9.0</TargetFramework>
-    <PlatformTarget>AnyCPU</PlatformTarget>
-    <IsPackable>false</IsPackable>
-    <NoWarn>;1591;1701;1702;1705;CA1822;CA1001</NoWarn>
+    <NoWarn>$(NoWarn);CA1822;CA1001</NoWarn>
```
JakenVeina, line 6: I've got no idea why you would be futzing around with warning settings as part of this PR. Or packaging configs, for that matter. The editorconfig/warning settings in the solution as a whole DESPERATELY need their own conversation.

JakenVeina (changes requested):
I've been wanting to chat about this since it came up in the Primitives PR.

Personally, I don't like the idea of global imports, as a whole, but I'm not terribly strong on the opinion. It more just extends from my dislike of C#'s trend towards making more and more things behave like magic, instead of just requiring devs to write a couple extra lines of code.

Is this a feature you've used in your own projects, or at work? How does it feel to use? Or do you even notice it?

Your methodology sounds fine, as far as what namespaces you're including or not.

dwcullop:
> Is this a feature you've used in your own projects, or at work? How does it feel to use? Or do you even notice it?

We've converted everything to use project-level (and even repo-level) usings and it has been a huge improvement.  It reduces merge conflicts that happen when people edit the same file (not as big of a deal on this project).  But I feel like when things need to be same, having to repeat them everywhere is error prone.  

It also has other nice properties like being to transition implementations by just changing the namespace.

It keeps the clutter down to a minimum.  Let's readers focus on the meat instead of the dressing (so to speak).

PR #1153, opened 2026-07-28
Make every BatchIf overload shape resolve

### LINKED ISSUE #1152: BatchIf cannot be called without redundant arguments: overloads are ambiguous
### Describe the bug 🐞

`ObservableCacheEx.BatchIf` cannot be called in its simplest form. `source.BatchIf(pauseSelector)` does not compile:

```
error CS0121: The call is ambiguous between the following methods or properties:
'ObservableCacheEx.BatchIf<TObject, TKey>(IObservable<IChangeSet<TObject, TKey>>, IObservable<bool>, IScheduler?)' and
'ObservableCacheEx.BatchIf<TObject, TKey>(IObservable<IChangeSet<TObject, TKey>>, IObservable<bool>, bool, IScheduler?)'
```

`source.BatchIf(pauseSelector, initialPauseState)` fails the same way, between the `bool, IScheduler?` and `bool, TimeSpan?, IScheduler?` overloads.

### Why

There are five overloads, and the parameters that tell them apart all carry defaults:

```csharp
BatchIf(source, pause, IScheduler? scheduler = null)
BatchIf(source, pause, bool initialPauseState = false, IScheduler? scheduler = null)
BatchIf(source, pause, TimeSpan? timeOut = default, IScheduler? scheduler = null)
BatchIf(source, pause, bool initialPauseState = false, TimeSpan? timeOut = default, IScheduler? scheduler = null)
BatchIf(source, pause, bool initialPauseState = false, IObservable<Unit>? timer = null, IScheduler? scheduler = null)
```

Because `initialPauseState` and `timeOut` are optional, a two-argument call is applicable to several overloads at once and none is better, so resolution fails.

The list-side `ObservableListEx.BufferIf` has the same five shapes but no defaults on those parameters, and every call form resolves cleanly. So this is a cache-side divergence rather than an inherent problem with the shape.

### Step to reproduce

```csharp
using var source = new SourceCache<Person, string>(p => p.Name);
using var pause = new Subject<bool>();

source.Connect().BatchIf(pause);         // CS0121
source.Connect().BatchIf(pause, true);   // CS0121

source.Connect().BatchIf(pause, Scheduler.Immediate);            // fine
source.Connect().BatchIf(pause, TimeSpan.FromSeconds(1));        // fine
source.Connect().BatchIf(pause, true, TimeSpan.FromSeconds(1));  // fine
```

Named arguments do not help. `BatchIf(pause, initialPauseState: true)` and `BatchIf(pause, timeOut: someTimeSpan)` are both ambiguous as well.

### Expected behavior

Every documented call shape should resolve to exactly one overload, as `BufferIf` already does.

The fix is to drop the defaults from the parameters that distinguish the overloads, matching the list side. Worth noting this is not a breaking change in practice: every call form that relies on those defaults is ambiguous today, so no code that currently compiles can depend on them.

### Additional information ℹ️

The workaround is an explicit cast to pin the overload, and it is already in use in this repository:

```csharp
.BatchIf(pauseBatch, false, (TimeSpan?)null)
```

See `CrossCacheDeadlockStressTest.cs` and `DeadlockTortureTest.cs`. Both were written that way because the natural call does not compile.

### DynamicData Version

main

dwcullop:
Fixes #1152.

`source.BatchIf(pause)` and `source.BatchIf(pause, initialPauseState)` did not compile. All five overloads carried defaults on the very parameters that tell them apart, so a short call was applicable to several at once with no best match and resolution failed with CS0121.

## Change

Dropped the defaults from the distinguishing parameters, and added an explicit two-parameter overload so the shortest form binds to a method of its own.

```diff
+ BatchIf(source, pause)

- BatchIf(source, pause, bool initialPauseState = false, IScheduler? scheduler = null)
+ BatchIf(source, pause, bool initialPauseState,         IScheduler? scheduler = null)

- BatchIf(source, pause, TimeSpan? timeOut = default,    IScheduler? scheduler = null)
+ BatchIf(source, pause, TimeSpan? timeOut,              IScheduler? scheduler = null)
```

Same treatment for the two longer overloads. This matches the list side, where `BufferIf` already has these shapes without defaults and every call form resolves.

## Compatibility

Nothing that compiles today can break, since every form relying on those defaults is currently ambiguous. Named forms like `BatchIf(pause, timeOut: x)` now resolve as well. Binary compatibility is unaffected, as defaults are baked in at the call site.

The new overload is an API addition, so this wants a minor version bump rather than a patch.

## API snapshots

Regenerating these by hand had stripped the UTF-8 BOM, converted LF to CRLF, and appended a trailing newline, which made line 1 show as changed for no visible reason. Rebuilt from main so only the signature lines differ.


src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet8_0.verified.txt
```diff
@@ -1,4 +1,4 @@
-﻿[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("DynamicData.Profile")]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("DynamicData.Profile")]
```
JakenVeina: Why on earth is this flagged as changed? Furthermore, what the hell is this? Project assemblies that don't exist anymore? Should probably remove these, as a separate PR.

src/DynamicData.Tests/Cache/BatchIfFixture.cs
```diff
@@ -140,4 +141,38 @@ public void ResultsWillBeReceivedIfNotPaused()
         _scheduler.AdvanceBy(TimeSpan.FromMinutes(1).Ticks);
         _results.Messages.Count.Should().Be(1, "Should be 1 update");
     }
+
+    [Fact]
```
JakenVeina: I don't like the idea of having "tests" for our API shape. At least not in hard-coded form. We could explore adding tests that use the same system as `ApiApprovalTests` to programmatically scan for overloads that would have ambiguity, but this strikes me as smelly.

src/DynamicData/Cache/ObservableCacheEx.BatchIf.cs
```diff
@@ -33,13 +33,13 @@ public static IObservable<IChangeSet<TObject, TKey>> BatchIf<TObject, TKey>(this
 
     /// <inheritdoc cref="BatchIf{TObject, TKey}(IObservable{IChangeSet{TObject, TKey}}, IObservable{bool}, bool, TimeSpan?, IScheduler?)"/>
     /// <remarks>This overload delegates to the primary overload with default <c>initialPauseState: false</c>.</remarks>
-    public static IObservable<IChangeSet<TObject, TKey>> BatchIf<TObject, TKey>(this IObservable<IChangeSet<TObject, TKey>> source, IObservable<bool> pauseIfTrueSelector, bool initialPauseState = false, IScheduler? scheduler = null)
+    public static IObservable<IChangeSet<TObject, TKey>> BatchIf<TObject, TKey>(this IObservable<IChangeSet<TObject, TKey>> source, IObservable<bool> pauseIfTrueSelector, bool initialPauseState, IScheduler? scheduler = null)
```
JakenVeina, line 42: So, by my reckoning, the call signature mentioned in the source issue (the one that only takes `source` and `pauseIfTrueSelector`) is no longer ambiguous, it's an outright error. I think in addition to making the existing overloads all take three-parameters, minimum, we should also add an overload that preserves the two-parameter behavior that was originally intended. Having no `scheulder` and a default `initialPauseState` of `false` still seems like the reasonable default way to use this operator.

JakenVeina (changes requested):

src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet8_0.verified.txt
dwcullop: Good catch, and the answer is embarrassing. I regenerated these by hand and my editor stripped the UTF-8 BOM, so line 1 differed by three invisible bytes. It had also converted every LF to CRLF and added a trailing newline. Git normalised the line endings away in the diff, which is why only the BOM and the final `}` showed up. I have rebuilt both files from main and applied only the signature edits, so the diff is now just the four changed lines plus the new overload.

On the `InternalsVisibleTo` entries, agreed those look stale. `DynamicData.Profile` and `DynamicData.ReactiveUI` are not in this solution. Worth its own PR rather than riding along here, so I will open an issue for it.

src/DynamicData.Tests/Cache/BatchIfFixture.cs
dwcullop: Agreed, and removed. Asserting API shape by hand is the wrong tool, especially when the assertions were incidental and the real check was whether the file compiled. The approval snapshot already covers the surface, and it caught this change without any help.

The programmatic ambiguity scan is an interesting idea. It would want to be more than an overload check, since ambiguity depends on which argument lists are actually reachable, but that seems worth exploring separately.

I have kept `PauseSelectorOnlyStartsUnpaused`, since that one tests behaviour rather than shape: it pins that the shortest form starts unpaused instead of binding to something that starts paused.

src/DynamicData/Cache/ObservableCacheEx.BatchIf.cs
dwcullop: The two-parameter call does still resolve, since `BatchIf(source, pause, IScheduler? scheduler = null)` is applicable with the scheduler defaulted, and `PauseSelectorOnlyStartsUnpaused` exercises exactly that shape. So it is not an error today.

That said, I think you are right that it should be explicit rather than incidental, so I have added the overload. The shortest form now binds to a method of its own instead of depending on another overload's default, which also makes the intended default behaviour visible in the API surface rather than implied.

src/DynamicData.Tests/API/ApiApprovalTests.DynamicDataTests.DotNet8_0.verified.txt
dwcullop: Filed as #1158 for the `InternalsVisibleTo` cleanup.

dwcullop:
Updated for all three points:

- Added the explicit two-parameter overload.
- Removed the overload-shape test. `PauseSelectorOnlyStartsUnpaused` stays, since it covers behaviour rather than shape.
- Rebuilt the API snapshots from main. The phantom line 1 change was a stripped UTF-8 BOM, plus LF converted to CRLF and an added trailing newline, all from regenerating them by hand.

`InternalsVisibleTo` cleanup is filed separately as #1158. Description has been tidied up too.

JakenVeina (approved):
After marinating for a day, I really like the idea of having a test that uses `Reflection` or whatever to search for ambiguous operator overloads across the whole solution. I've created #1160 for this.

Otherwise, good to go.

PR #1155, opened 2026-07-29
Fixed that .WhenPropertyChanged() did not support a variety of types of implicit casts…

### LINKED ISSUE #1149: WhenValueChanged throws InvalidCastException when acted on a generic type that implements an interface
### Describe the bug 🐞

WhenValueChanged method, when acting on a generic type **T**, that implements an interface **Interface** that contains a property we are interested in, throws System.InvalidCastException : Object must implement IConvertible.

### Step to reproduce

You can reproduce the issue by running the test provided:

1. We create a class that implements an interface, that exposes the property we are interested in observing
2. We have a generic method that defines the generic type, implementing the interface
3. We create the subscription and expect it to modify the value accordingly on value changed
4. The subscription fails during creation

```csharp
[TestFixture]
public class WhenValueChangedGenericConstraintTests
{
    [Test]
    public void WhenValueChanged_ThroughInterfaceConstrainedGeneric_Throws()
    {
        var holder = new FlagHolder();
        bool? observed = null;

        // Expected (green): the subscription observes Flag; the change below sets observed = true.
        // Actual (bug): this line throws System.InvalidCastException "Object must implement IConvertible".
        using var subscription = ObserveFlagWithGenerics(holder)
            .Subscribe(value => observed = value);

        holder.Flag = true;

        Assert.That(observed, Is.True);
    }

    // Observing value change through a generic parameter constrained to the interface throws an error.
    // Calling holder.WhenValueChanged(x => x.Flag) directly, or on an IHasFlag-typed variable, does NOT reproduce it
    private static IObservable<bool> ObserveFlagWithGenerics<T>(T source) where T : IHasFlag
    {
        return source
            .WhenValueChanged(x => x.Flag);
    }
}

internal interface IHasFlag : INotifyPropertyChanged
{
    bool Flag { get; }
}

internal sealed class FlagHolder : IHasFlag
{
    private bool _flag;

    public bool Flag
    {
        get => _flag;
        set
        {
            if (_flag == value)
                return;

            _flag = value;
            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(Flag)));
        }
    }

    public event PropertyChangedEventHandler PropertyChanged;
}

```

### Reproduction repository

see code above

### Expected behavior

We'd expect the subscription to work, given that the generic type is implementing the interface and therefore we should be able to cast it accordingly and access the property.

The test provided should pass.

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

Windows

### Version

_No response_

### Device

_No response_

### DynamicData Version

9.4.33

### Additional information ℹ️

Project is on .NET 9.0

JakenVeina:
… within the target expression that it should have. In particular, this allows for more flexibility for the operation, within generic type contexts.

Resolves #1149.

Long story short, when I added support for implicit casts back in #1059, I misinterpreted what `Convert.ChangeType()` is supposed to do. Based on some other examples of code that parses cast expressions, I identified `Convert.ChangeType()` as the runtime's internal mechanism for doing a broad range of casts. It's actually only for doing casts between `IConvertible` types. Really, the runtime itself is already capable of doing all the casting work we need, given that we're operating exclusively in `object` land, and we hand off to the runtime to get us back in to generic-typed land. So, I just removed `Convert.ChangeType()` entirely, and that covers all the existing behavior we had tests for, as well as the new scenario from #1149.

### DIFF
--- a/src/DynamicData.Tests/Binding/WhenPropertyChangedBehaviorFixture.cs
+++ b/src/DynamicData.Tests/Binding/WhenPropertyChangedBehaviorFixture.cs
@@ -5,9 +5,9 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
-
+using System.Linq;
 using DynamicData.Binding;
-
+using DynamicData.Tests.Utilities;
 using FluentAssertions;
 
 using Xunit;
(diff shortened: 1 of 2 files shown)


dwcullop (approved):
LGTM

PR #1159, opened 2026-07-30
Removed extraneous internal access grants, from old, long-removed, projects.

### LINKED ISSUE #1158: Remove stale InternalsVisibleTo entries for assemblies not in this solution
`src/DynamicData/Attributes.cs` grants internals access to three assemblies:

```csharp
[assembly: InternalsVisibleTo("DynamicData.Tests")]
[assembly: InternalsVisibleTo("DynamicData.ReactiveUI")]
[assembly: InternalsVisibleTo("DynamicData.Profile")]
```

Only `DynamicData.Tests` is in this solution. The projects here are `DynamicData`, `DynamicData.Tests`, and `DynamicData.Benchmarks`. `DynamicData.ReactiveUI` and `DynamicData.Profile` are not built from this repository.

Two reasons to tidy this up:

- The grants are unconditional and the assembly is not signed, so anything compiled with a matching assembly name picks up access to our internals.
- They show up at the top of both API approval snapshots, where they read as part of the public surface and invite exactly the question of what those assemblies are.

Worth confirming whether `DynamicData.ReactiveUI` still consumes internals from a separate repository before removing that one. `DynamicData.Profile` looks like it predates `DynamicData.Benchmarks`.

Raised by @JakenVeina during review of #1153.

JakenVeina:
Resolves #1158.


dwcullop (approved):
Thanks for fixing this before I could get to it.  LGTM!

PR #1163, opened 2026-08-06
Move change batching out of the delivery queue

dwcullop:
`SharedDeliveryQueue` takes a callback that fires once per drain cycle, and `CacheParentSubscription` uses it as its only emit point. That is a batching policy living inside a class whose job is serialization. Raised on #1114.

It also batches more than intended. `DrainPending` drains whatever is queued, so work another thread enqueued mid-drain lands in the same emission.

## Change

`CacheParentSubscription` tracks its own delivery frame:

```csharp
private void EndFrame()
{
    if (--_frameDepth != 0)
    {
        return;
    }

    EmitChanges(_observer);
    ...
}
```

A child that emits synchronously during parent processing is delivered inline by the queue's reentrant path, so it nests inside the parent's frame instead of emitting separately. One upstream notification plus everything it triggers synchronously still produces one downstream changeset.

The callback then has no consumers, so the field, the overload taking it, and the invoke site all go. `SharedDeliveryQueue` only serializes now.

No lock around the depth counter. The queue has already serialized delivery, so only one thread is ever inside those methods, and the queue's lock is the barrier between drains.

## The batching is load-bearing

`EmitChanges` is abstract and that callback was its only caller, so the six operators built on this class stop emitting entirely without a replacement.

Emitting per notification instead fails 34 tests, all on changeset granularity. `ClearingParentEmitsSingleChangeSet` gives 15 where it expects 2. `OrderOfChangesIsPreserved` gives 11 where it expects 2, and that one is more than a count: a single `Edit` doing `Clear()` then `AddOrUpdate()` has to arrive as one changeset, or the collection is observably emptied, which is a state that never existed upstream.

## What actually changes

A second thread's work is no longer folded into the same emission. It gets its own frame.

## Related

#1162 fixes the queue delivering out of receipt order. Independent, but this change exposes that one: once emission is per frame, delivery order becomes changeset order. Same file, so whichever lands second needs a trivial merge.

### DIFF
--- a/src/DynamicData/Internal/CacheParentSubscription.cs
+++ b/src/DynamicData/Internal/CacheParentSubscription.cs
@@ -120,62 +120,45 @@ internal abstract class CacheParentSubscription<TParent, TKey, TChild, TObserver
 
     private void DeliverParent(IChangeSet<TParent, TKey> changes)
     {
-        ++_frameDepth;
-        try
-        {
-            ParentOnNext(changes);
-        }
-        finally
-        {
-            EndFrame();
-        }
+        using var frame = BeginFrame();
+        ParentOnNext(changes);
     }
 
     private void DeliverChild(TChild child, TKey parentKey)
     {
-        ++_frameDepth;
-        try
-        {
-            ChildOnNext(child, parentKey);
-        }
-        finally
-        {
-            EndFrame();
-        }
+        using var frame = BeginFrame();
+        ChildOnNext(child, parentKey);
     }
 
     private void CompleteParent()
     {
-        ++_frameDepth;
-        try
-        {
-            CheckCompleted();
-        }
-        finally
-        {
-            EndFrame();
-        }
+        using var frame = BeginFrame();
+        CheckCompleted();
     }
 
     private void CompleteChild(TKey parentKey)
     {
-        ++_frameDepth;
-        try
-        {
-            RemoveChildSubscription(parentKey);
-        }
-        finally
-        {
-            EndFrame();
-        }
+        using var frame = BeginFrame();
+        RemoveChildSubscription(parentKey);
     }
 
     /// <summary>
-    /// Closes the current delivery frame. Deliveries nested beneath this one, which the queue
-    /// runs inline on the same thread, close their own frame first and leave the emit to the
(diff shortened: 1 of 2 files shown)


src/DynamicData/Internal/CacheParentSubscription.cs
```diff
@@ -116,8 +118,72 @@ protected virtual void Dispose(bool disposing)
     protected IObservable<T> MakeChildObservable<T>(IObservable<T> observable) =>
         observable.SynchronizeSafe(_queue);
 
-    private void OnDrainComplete()
+    private void DeliverParent(IChangeSet<TParent, TKey> changes)
     {
+        ++_frameDepth;
```
JakenVeina, line 123: Would it be crazy to say that this should be a `StartFrame()` or `BeginFrame()` method, just for the clarity that it pairs with `EndFrame()`?

Would it be even crazier to wrap up this logic in a `private struct FrameTracker : IDisposable` that does the tracking, and lets you just collapse it all to a `using`?

dwcullop: No, I like that idea.  More clearly expresses intent.

dwcullop: Done in 677c9093 — went with both halves of the suggestion.

There's now an explicit `BeginFrame()` that pairs with `EndFrame()`, and it returns a `private readonly struct FrameTracker : IDisposable` that closes the frame on dispose. Each delivery collapses to:

```csharp
private void DeliverParent(IChangeSet<TParent, TKey> changes)
{
    using var frame = BeginFrame();
    ParentOnNext(changes);
}
```

That replaces four copies of `++_frameDepth; try { ... } finally { EndFrame(); }`, so the increment and decrement can't drift apart. Net −6 lines despite adding the struct and its docs.

The tracker is a `readonly struct`, so `using var` dispatches `Dispose()` directly with no boxing — the frame stays allocation-free, which matters given this sits on the delivery path of six operators.

Semantics are unchanged. Full suite green on both net9.0 and net10.0 (2476 passed / 0 failed), including the granularity tests this class exists to satisfy — `ClearingParentEmitsSingleChangeSet` and both `OrderOfChangesIsPreserved` cases.

JakenVeina (approved):

PR #1165, opened 2026-08-08
WIP Release 9.5

JakenVeina:
Backporting fixes and non-breaking enhancements from main to 9.x, for a 9.5 release.


src/DynamicData/Binding/ObservablePropertyFactory.cs
```diff
@@ -1,87 +1,275 @@
...
+        // propagates back to whoever invoked the PropertyChanged setter, matching what a plain
+        // Subject<T>.OnNext would do.
+        private void EmitCurrent()
+        {
+            lock (_notificationGate)
+            {
+                PropertyValue<TObject, TProperty> value;
+                try
+                {
+                    value = new PropertyValue<TObject, TProperty>(_source, _accessor(_source));
```
dwcullop, line 108: [P1] Avoid holding the notification gate while invoking model code

This introduces an ABBA deadlock during initial subscription. A setter can hold the model's lock and raise PropertyChanged, then wait for _notificationGate. Meanwhile, the subscription thread holds _notificationGate and waits for the same model lock inside the getter. The gate is also held across downstream OnNext callbacks.

The original #1111 implementation on main uses queued delivery and does not hold this gate while invoking user code. Replacing the queue with a monitor is not equivalent. Please preserve registration-before-initial-read without holding a lock across getters or downstream callbacks that event producers also need to acquire.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -17,29 +18,108 @@ internal sealed class Switch<TObject, TKey>(IObservable<IObservable<IChangeSet<T
...
+                                    isSourceRunning = false;
+
+                                    if (areSourcesComplete)
+                                    {
+                                        observer.OnCompleted();
+                                    }
+                                }
+                            });
+                    },
+                    observer.OnError,
```
dwcullop: [P1] Serialize outer errors with the inner notifications

The outer error handler calls observer.OnError directly, while inner OnNext, OnError, and OnCompleted use synchronizationGate. An outer failure on one thread can therefore enter downstream error handling while an inner OnNext is still running on another thread. Both input streams can individually obey the Rx contract, but this operator's output violates it.

The original #1137 implementation routes this through queue.OnError. Without that queue, the backport needs a consistent 9.x-compatible serialization mechanism for all output notifications, including outer errors. Please preserve the intended completion fix while correcting this path.

src/DynamicData/Cache/ObservableCache.cs
```diff
@@ -61,6 +61,16 @@ public ObservableCache(IObservable<IChangeSet<TObject, TKey>> source)
...
                 _changes.OnError(ex);
+
+                if (_countChanged.IsValueCreated)
+                {
+                    _countChanged.Value.OnError(ex);
+                }
+
+                if (_suspensionTracker.IsValueCreated)
+                {
+                    _suspensionTracker.Value.Fault(ex);
```
dwcullop, line 82: [P2] Make suspension faulting safe against disposal from an error callback

Once the suspension tracker has been created, a Connect subscriber with an error handler that calls cache.Dispose() can break this path. _changes.OnError(ex) invokes that handler and disposes the tracker. Control then returns here, and Fault(ex) calls OnError on the disposed BehaviorSubject, throwing ObjectDisposedException back out of the producer despite the original error having a handler.

The original #1141 implementation relies on main's queued terminal-event handling, which prevents this teardown from occurring in the middle of the error delivery. Please coordinate tracker termination with disposal so this is safe for both reentrant and concurrent disposal in the 9.x implementation.

src/DynamicData.Tests/Cache/SuspendNotificationsFixture.IntegrationTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            resumeThread.Start();
+            deliveryStarted.Wait(TimeSpan.FromSeconds(10)).Should().BeTrue("delivery should have started");
+
+            // The resume thread is parked after decrementing the count but before signalling
+            // resume. Re-suspend and connect a new subscriber while the count is transiently zero.
+            var suspend2 = cache.SuspendNotifications();
+            using var lateResults = cache.Connect().AsAggregator();
+
+            // Let the resume thread proceed to its now-stale resume signal.
+            releaseDelivery.Set();
```
dwcullop: [P3] Do not let this race test pass by waiting for its timeout

In 9.x, ResumeNotifications holds the cache's _locker while delivering the pending changeset. The subscriber waiting on releaseDelivery still owns that lock, so this SuspendNotifications call blocks before the test can reach releaseDelivery.Set(). Progress only occurs when the ignored 30-second wait times out.

The test then passes after a sequential re-suspension, without exercising the stale-resume window it claims to force. Main's queue-based delivery releases the cache lock before the callback, so the same assumption does not transfer to this #1132 backport. Please replace this with a deterministic interleaving that is reachable under the 9.x locking model, and assert coordination wait results rather than silently continuing after a timeout. The DQ-related problem here does not require reverting the entire test rewrite.

dwcullop (changes requested):
I'd like to keep these fixes in 9.5, but the backports need to work correctly without the 10.0 delivery-queue infrastructure. The four comments below cover three runtime regressions and a regression test whose intended interleaving is not reachable under the 9.x locking model.

Please address these with 9.x-compatible implementations. This is not a request to revert the original PRs from main, where the delivery-queue infrastructure is present.

src/DynamicData/Binding/ObservablePropertyFactory.cs
JakenVeina: If I'm reading this right, we have the following sequence:

1. Thread A creates a new `SinglePropertySubscription`, which subscribes to `_source.PropertyChanged` in the constructor.
2. Before Thread A enters `_notificationGate` to retrieve and emit the initial value, Thread B invokes `_source.PropertyChanged`, which can enter `_notificationGate` first.
3. Thread A catches up, tries to etner `_notificationGate`, and can't.
4. Nothing within DD code stops Thread B from leaving `_notificationGate`, so when it leaves, Thread A finally gets to take its lock and emit the "initial" notification. Now, this notification is no longer "initial" and is probably redundant, but there's no deadlock.

The ABBA scenario here is on the CONSUMER side, which is what the DQ will fix, when we get that finished. Until then, ABBA deadlock scenarios introduced by the consumer are not a defect, they're unsupported behavior.

src/DynamicData/Cache/ObservableCache.cs
JakenVeina: I ended up having to fix this already when cherry-picking #1133.

src/DynamicData.Tests/Cache/SuspendNotificationsFixture.IntegrationTests.cs
JakenVeina: I think this should cover it.

I'm not a fan of tests like these, where we're forcing scenarios based on knowledge of the internal workings of the UUT. It makes the test fragile, as this issue highlights. The end goal of the test hasn't changed from main: we want to make sure that internal state can't get corrupted when being manipulated across different threads at the same time. But because the internal implementation is different, the entire test has to be re-written.

That being said, the alternative is a non-deterministic test, which we have used elsewhere, but are risky. We should probably brainstorm on that, at some point.

For now, I've re-written the test to be as deterministic as possible. Because all these methods block each other now, it's not TECHNICALLY possible to be purely deterministic. I can't TRULY guarantee that, say, a `.SuspendNotifications()` call has started and then blocked, but I can signal both before and after the call, and give the `ThreadPool` opportunity to run before checking things.

src/DynamicData/Cache/Internal/Switch.cs
JakenVeina: I think I had intended to put a `.Synchronize()` upstream, and then decided against it. Fixed.

PR #1166, opened 2026-08-08
Fix Switch resolution for virtualized changesets

### LINKED ISSUE #1164: [Bug]: Switch operator lacks support of virtualized changeset
### Describe the bug 🐞

There is no Switch operator version for virtualized changesets `IChangeSet<TObject, TKey, VirtualContext<TObject>>`. 

In this case default reactive `Switch` operator is used, so the test fails.

```c#
 [Fact]
 public void SwitchSet_Switch_Virtualized()
 {
     BehaviorSubject<bool> enableSubj = new(true);
     SourceCache<int, int> cache = new(x => x);
     cache.Edit(updater => updater.AddOrUpdate(Enumerable.Range(0, 100)));
     SortExpressionComparer<int> comparer = SortExpressionComparer<int>.Ascending(x => x);

     int count = 10;

     IObservable<VirtualRequest> virtualReqStream
         = Observable.Return(new VirtualRequest(0, count)).Concat(Observable.Never<VirtualRequest>());

     IObservable<IChangeSet<int, int, VirtualContext<int>>> virtualizedChangeSet = cache.Connect()
         .SortAndVirtualize(comparer, virtualReqStream);

     IObservable<IChangeSet<int, int, VirtualContext<int>>> testStream = enableSubj
         .Select(enable =>
         {
             if (!enable)
                 return Observable.Empty<IChangeSet<int, int, VirtualContext<int>>>();

             return virtualizedChangeSet;
         })
         .Switch();

     TestScheduler scheduler = new();

     ITestableObserver<IChangeSet<int, int, VirtualContext<int>>> observer
         = scheduler.CreateObserver<IChangeSet<int, int, VirtualContext<int>>>();

     testStream.Subscribe(observer);

     enableSubj.OnNext(false);

     Assert.Collection(
         observer.Messages,
         msg =>
         {
             Assert.Equal(NotificationKind.OnNext, msg.Value.Kind);
             Assert.Equal(count, msg.Value.Value.Adds);
             Assert.Equal(0, msg.Value.Value.Removes);
         },
         msg =>
         {
             Assert.Equal(NotificationKind.OnNext, msg.Value.Kind);
             Assert.Equal(0, msg.Value.Value.Adds);
             Assert.Equal(count, msg.Value.Value.Removes);
         }
     );
 }
```

### DynamicData Version

9.4.33

Guflly:
Fixes #1164.

Adds a virtualized changeset overload so DynamicData's Switch is selected outside DynamicData namespaces. Includes coverage for switching to an empty source and between virtualized sources.

Tests:
- Switch, SortAndVirtualize, and API approval tests (41 passed)
- .NET 9 build

### DIFF
--- a/src/DynamicData.Tests/Cache/VirtualizedSwitchFixture.cs
+++ b/src/DynamicData.Tests/Cache/VirtualizedSwitchFixture.cs
@@ -4,7 +4,7 @@ using System.Linq;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using DynamicData;
-using DynamicData.Tests;
+using DynamicData.Tests.Utilities;
 using FluentAssertions;
 using Xunit;
 
@@ -23,16 +23,19 @@ public class VirtualizedSwitchFixture
         IObservable<IObservable<IChangeSet<int, int, VirtualContext<int>>>> sources = enabled.Select(
             isEnabled => isEnabled ? virtualized : Observable.Empty<IChangeSet<int, int, VirtualContext<int>>>());
         IObservable<IChangeSet<int, int>> switched = sources.Switch();
-        using ChangeSetAggregator<int, int> results = switched.AsAggregator();
+        using var subscription = switched
+            .ValidateChangeSets(static value => value)
+            .RecordCacheItems(out var results);
 
-        results.Messages.Should().ContainSingle();
-        results.Messages[0].Adds.Should().Be(10);
+        results.RecordedChangeSets.Should().ContainSingle();
+        results.RecordedChangeSets[0].Adds.Should().Be(10);
+        results.RecordedItemsByKey.Values.Should().BeEquivalentTo(Enumerable.Range(0, 10));
 
         enabled.OnNext(false);
 
-        results.Messages.Should().HaveCount(2);
-        results.Messages[1].Removes.Should().Be(10);
-        results.Data.Items.Should().BeEmpty();
+        results.RecordedChangeSets.Should().HaveCount(2);
+        results.RecordedChangeSets[1].Removes.Should().Be(10);
+        results.RecordedItemsByKey.Should().BeEmpty();
     }
 
     [Fact]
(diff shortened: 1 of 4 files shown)


src/DynamicData/Cache/ObservableCacheEx.Switch.cs
```diff
@@ -37,6 +37,17 @@ public static IObservable<IChangeSet<TObject, TKey>> Switch<TObject, TKey>(this
         return sources.Select(cache => cache.Connect()).Switch();
     }
 
+    /// <inheritdoc cref="Switch{TObject, TKey}(IObservable{IObservable{IChangeSet{TObject, TKey}}})"/>
+    /// <param name="sources">An observable that emits virtualized changeset streams.</param>
+    public static IObservable<IChangeSet<TObject, TKey>> Switch<TObject, TKey>(this IObservable<IObservable<IChangeSet<TObject, TKey, VirtualContext<TObject>>>> sources)
+        where TObject : notnull
+        where TKey : notnull
+    {
+        sources.ThrowArgumentNullExceptionIfNull(nameof(sources));
```
JakenVeina, line 46: The `.Switch()` overload you're aliasing already has a null-check, no point in repeating it.

Alternatively, do what the other `.Switch()` overload does and call `new Switch().Run()` directly.

src/DynamicData.Tests/Cache/VirtualizedSwitchFixture.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        using var enabled = new BehaviorSubject<bool>(true);
+        using var source = new SourceCache<int, int>(value => value);
+        source.AddOrUpdate(Enumerable.Range(0, 20));
+
+        IObservable<IChangeSet<int, int, VirtualContext<int>>> virtualized = CreateVirtualized(source, 10);
+        IObservable<IObservable<IChangeSet<int, int, VirtualContext<int>>>> sources = enabled.Select(
+            isEnabled => isEnabled ? virtualized : Observable.Empty<IChangeSet<int, int, VirtualContext<int>>>());
+        IObservable<IChangeSet<int, int>> switched = sources.Switch();
+        using ChangeSetAggregator<int, int> results = switched.AsAggregator();
```
JakenVeina, line 26: Throw in a `.ValidateChangeSets()` call, between the UUT and the results collector.

Also, recommend using `.RecordCacheItems()` over `.AsAggregator()`.

src/DynamicData.Tests/Cache/VirtualizedSwitchFixture.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        results.Messages.Should().ContainSingle();
+        results.Messages[0].Adds.Should().Be(10);
+
```
JakenVeina, line 30: You should also be asserting against the actual collected items, to make sure they match.

JakenVeina (changes requested):

Guflly:
Updated this: removed the redundant null check, added ValidateChangeSets with RecordCacheItems, and asserted the collected items. The 15 related Switch tests pass locally.

PR #1185, opened 2026-09-21
Validate legacy TransformSafeAsync factories before forwarding

### LINKED ISSUE #1183: [Bug]: TransformSafeAsync options overload accepts a null legacy factory
### Describe the bug 🐞

The compatibility overload wraps a nullable three-argument transform factory in a non-null cancellation-aware lambda without first validating the original factory. The delegated null check sees the wrapper, so invalid configuration becomes per-item NullReferenceException errors rather than an eager argument error.

### Step to reproduce

Pass a typed null Func<Person, Optional<Person>, string, Task<Person>> to TransformSafeAsync with an error handler and TransformAsyncOptions.Default. Do not subscribe. Repeat with empty and populated source caches.

Regression tests/checks:

```
OptionsOverload_NullFactory_ThrowsBeforeSubscription
```

### Reproduction repository

Regression tests and fixes:

### Expected behavior

The operator call must immediately throw ArgumentNullException naming transformFactory, before any subscription or item processing.

### Screenshots 🖼️

N/A.

### IDE

### Operating system

Windows.

### Version

### Device

N/A.

### DynamicData Version

### Additional information ℹ️

Both repository cases fail because no exception is thrown. Restoring the argument checks makes both pass.

dwcullop:
The legacy `TransformSafeAsync` overloads forwarded their arguments to the newer implementation without validating them first, so a null `transformFactory` or `errorHandler` surfaced later (or not at all) instead of failing at the call site.

This adds eager argument validation to those overloads so a null factory or error handler throws `ArgumentNullException` immediately, matching the behavior of the rest of the cache operator surface. Tests cover the null-argument cases for the affected overloads.

Focused validation for this change passed prior to submission.

Fixes #1183


src/DynamicData.Tests/Cache/TransformSafeAsyncFixture.ArgumentValidation.cs
```diff
@@ -0,0 +1,52 @@
...
+using Xunit;
+using Xunit.Abstractions;
+
+namespace DynamicData.Tests.Cache;
+
+public partial class TransformSafeAsyncFixture
+{
+    private const int ArgumentValidationSeed = 0x2409_1165;
+
+    private readonly Randomizer _argumentRandomizer = new(ArgumentValidationSeed);
```
JakenVeina: There is 0 need for randomization in this test.

src/DynamicData/Cache/ObservableCacheEx.TransformSafeAsync.cs
```diff
@@ -134,6 +134,10 @@ public static IObservable<IChangeSet<TDestination, TKey>> TransformSafeAsync<TDe
         where TSource : notnull
         where TKey : notnull
     {
+        source.ThrowArgumentNullExceptionIfNull(nameof(source));
```
JakenVeina: `source` and `errorHandler` null checks here are redundant, when the inner operator already provides them.

JakenVeina (changes requested):

PR #1186, opened 2026-09-21
Make filtering regression assertions reject stale entries

### LINKED ISSUE #1184: [Bug]: Filter and RemoveKey regression assertions accept stale extra entries
### Describe the bug 🐞

The added full-collection assertions use Assert.Equivalent with its default strict:false setting. That allows extra actual entries, so the tests can accept the precise stale-item failure they intend to reject. This is a test defect, not an xUnit library defect.

### Step to reproduce

Use the existing refresh/removal fixture and omit its production removal operation as a controlled mutation. Compare the same fixture with its old non-strict assertions and corrected strict assertions.

Regression tests/checks:

```
AutoRefreshRemoveKeyFilterUpdate_CollectionUpdated; AutoRefreshFilterRemoveKeyUpdate_CollectionUpdated
```

### Reproduction repository

Regression tests and fixes:

### Expected behavior

A test claiming the exact filtered collection must fail when an excluded or duplicate entry remains.

### Screenshots 🖼️

N/A.

### IDE

### Operating system

Windows.

### Version

### Device

N/A.

### DynamicData Version

### Additional information ℹ️

In an isolated baseline snapshot, removing the Replace true-to-false downstream removal makes the real integration fixture pass with strict:false. The identical mutation fails with strict:true: Assert.Equivalent reports an excluded item left over in the actual collection. The live source is not mutated, and corrected strict assertions pass with normal production code.

dwcullop:
## Summary

The static `Filter` regression tests for the cache asserted only on counts and on membership of the items that were expected to survive. That shape passes even when the operator leaves a stale entry behind: a result set containing an extra item that should have been filtered out satisfies "contains the expected items" just fine. The assertions were weak enough to hide exactly the class of bug they exist to catch.

This change corrects the assertions to be exact rather than permissive. The result collections are now compared against the full expected set with `Assert.Equivalent(..., strict: true)`, so any leftover entry fails the test. The setup was also tightened so the comparison is meaningful: the threshold is derived from the generated data rather than from an average, which guarantees a known starting membership, and the mutations step through a partial exclusion and then a total exclusion so the strict comparison is exercised against both a non-empty and an empty expectation.

## Evidence

The assertion correction is backed by controlled mutation: each strengthened test was run against a deliberately broken operator that leaves a stale entry in place. Under the original assertions those mutants pass; under the corrected assertions they fail, and they pass again once the mutation is reverted. That confirms the new assertions are doing real work and are not merely more verbose restatements of the old ones.

## Scope

Tests only, limited to `src/DynamicData.Tests/Cache/FilterFixture.Static.cs`. No production code is touched by this PR.

Fixes #1184


src/DynamicData.Tests/Cache/FilterFixture.Static.cs
```diff
@@ -4,19 +4,28 @@
...
 namespace DynamicData.Tests.Cache;
 
 public static partial class FilterFixture
 {
     public sealed class Static
         : Base
     {
+        private const int RemoveKeySeed = 0x1165;
+
+        public Static(ITestOutputHelper output)
```
JakenVeina, line 26: What purpose does logging the seed serve?

JakenVeina (changes requested):
These two tests shouldn't even BE here. They're using the `List`-land `.Filter()` not `Cache`-land. they need to be moved over there.

They also should be updated to use `FluentAssertions`, like the other tests in this fixture, and the one they'll be moving to.

They also really shouldn't be using operators not under test (I.E. `.RemoveKey()`).

PR #1188, opened 2026-09-21
Preserve Switch subscription ownership across reentrant handoffs

### LINKED ISSUE #1179: [Bug]: Switch can overlap exclusive subscriptions and discard a reentrantly selected source
### Describe the bug 🐞

The cache Switch operator assigns the result of SubscribeSafe to its SerialDisposable after subscription has already begun. A replacement can try to acquire a resource still held by the previous subscription. Synchronous initial delivery, reset delivery, or disposal can also select a newer source, after which the older selection starts or overwrites the current subscription. The latest source is disconnected and its updates are lost.

### Step to reproduce

Select a cold changeset source that owns an exclusive resource, then select it again. Separately, select a source with a synchronous initial Add and select another source from the downstream callback. Both are supported single-threaded scenarios.

Regression tests/checks:

```
ReplacementReleasesPreviousSubscriptionBeforeAcquiringResource; ReentrantSelectionDuringSynchronousAddKeepsLatestSubscription; ReentrantSelectionDuringResetDoesNotSubscribeSupersededReplacement; ReentrantSelectionDuringDisposalDoesNotSubscribeSupersededReplacement; DisposalDuringResetDoesNotSubscribeReplacement; OuterErrorDuringResetDoesNotSubscribeReplacement
```

### Reproduction repository

Regression tests and fixes:

### Expected behavior

The old subscription must be released before its replacement starts. A superseded or canceled selection must not subscribe, and a late returning subscription must not replace the newest selection.

### Screenshots 🖼️

N/A.

### IDE

### Operating system

Windows.

### Version

### Device

N/A.

### DynamicData Version

### Additional information ℹ️

Repository tests fail with an exclusive-resource InvalidOperationException, latestRemainedSubscribed=false, or a canceled replacement subscription count of 1 rather than 0.

dwcullop:
## Summary

`Switch` could lose track of which inner subscription owned the current source when the outer observable re-entered during a reset, disposal, or the initial delivery of a newly selected source. A reentrant selection could activate a source that had already been superseded, or dispose the subscription belonging to the newest source, leaving the operator wired to a stale stream or to nothing at all.

### Generation-scoped ownership

Each selection of an inner source now carries its own generation identity. Activation, population, and teardown are all validated against the generation that initiated them, so a handoff triggered while an earlier handoff is still unwinding can no longer publish results from, or tear down, a subscription it does not own. A superseded generation quietly retires instead of racing the current one.

### Exclusive-resource ordering

The subscription handoff is ordered so that the outgoing subscription is fully released before its replacement starts, rather than overlapping or being released afterwards by whichever path completed last. Outer completion is deferred until the selected inner source terminates, while errors and disposal cancel any activation still pending. The result is a single, unambiguous owner of the inner subscription at every point in the handoff.

## Validation

Focused validation for this change was previously run and passed, covering the reentrant reset, disposal, and initial-delivery handoff paths along with the existing `Switch` suite.

Fixes #1179


dwcullop:
**P2: Reentrant initial delivery still permits overlapping exclusive-resource ownership.**

Locations: [`Switch.cs`, line 108](https://github.com/reactivemarbles/DynamicData/blob/0eac5f26b75b8169fffd5010dafd9603a7bf9023/src/DynamicData/Cache/Internal/Switch.cs#L108), and replacement subscription at line 139.

This reproduces on both the PR and its base commit, so it is an unresolved gap, not a regression introduced here. It also limits the newly documented guarantee that the previous subscription is released before its replacement starts.

Reproduction: an inner `Observable.Create` acquires an exclusive resource and emits synchronously. The downstream `OnNext` selects a replacement source that needs the same resource. The first `Subscribe()` has not yet returned its disposable, so `previous.Dispose()` only marks an empty `SingleAssignmentDisposable` as disposed. The replacement starts immediately, finds the resource still held, and terminates the switched stream with an error. The first resource is released only when its original subscription call unwinds.

Suggested fix: serialize activation so a replacement selected during an in-progress `Subscribe()` waits until that call returns and its late disposable has actually been released. Preserve latest-selection semantics while activation is pending, and cancel pending activation on disposal or error. This needs explicit activation-state coordination rather than a different disposable assignment alone. Add a test combining exclusive-resource ownership with reentrant synchronous initial delivery; the current tests exercise those conditions separately.

src/DynamicData/Cache/ObservableCacheEx.Switch.cs
```diff
@@ -47,6 +47,14 @@ public static IObservable<IChangeSet<TObject, TKey>> Switch<TObject, TKey>(this
     /// <returns>A changeset stream reflecting the items from the most recently emitted inner source.</returns>
     /// <remarks>
     /// <para>On switch: <b>Remove</b> is emitted for all items from the previous source, then <b>Add</b> for all items from the new source.</para>
+    /// <para>
```
JakenVeina, line 50: Redundant remarks. All are describing basic RX contract behavior that does not need to be re-stated.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -24,101 +25,170 @@ public IObservable<IChangeSet<TObject, TKey>> Run() => Observable.Create<IChange
 
                 // What the current source has contributed, so that switching away can take it back out.
                 var current = new Cache<TObject, TKey>();
-                var subscription = new SerialDisposable();
+                var outer = new SingleAssignmentDisposable();
```
JakenVeina, line 28: No idea what "outer" is supposed to mean, here. This seems to just be a container for the subscription to `_sources`? Why does it need its own container? It looks like it's exclusively because you're putting teardown logic at the beginning of the operator, in `lifetime`.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -13,8 +13,9 @@ internal sealed class Switch<TObject, TKey>(IObservable<IObservable<IChangeSet<T
 {
     private readonly IObservable<IObservable<IChangeSet<TObject, TKey>>> _sources = sources ?? throw new ArgumentNullException(nameof(sources));
 
-    public IObservable<IChangeSet<TObject, TKey>> Run() => Observable.Create<IChangeSet<TObject, TKey>>(
-            observer =>
+    public IObservable<IChangeSet<TObject, TKey>> Run() => Observable.Using(
```
JakenVeina, line 16: Not really clear what `.Using()` accomplishes here.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -24,101 +25,170 @@ public IObservable<IChangeSet<TObject, TKey>> Run() => Observable.Create<IChange
...
-                                    scope.EnqueueNext(changes);
-                                }
-                            },
-                            error =>
-                            {
-                                using var scope = queue.AcquireLock();
+                        subscription = new SingleAssignmentDisposable();
+                        previous = activeSubscription;
+                        activeSubscription = subscription;
+                        isSourceRunning = true;
```
JakenVeina, line 104: This seems wrong. Or at least misleading. No source is actually running, yet.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -24,101 +25,170 @@ public IObservable<IChangeSet<TObject, TKey>> Run() => Observable.Create<IChange
...
+                    }
 
-                                if (sourceId != activeSourceId)
-                                {
-                                    return;
-                                }
+                    // Disposal is user code too: it can select another source or terminate the result.
```
JakenVeina, line 107: Point being that we don't want to lock around user code, because user code can loop back and deadlock, under the DQ paradigm. Comment should make that clearer.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -24,101 +25,170 @@ public IObservable<IChangeSet<TObject, TKey>> Run() => Observable.Create<IChange
...
-                                    scope.EnqueueCompleted();
-                                }
-                            });
-                    },
-                    queue.OnError,
-                    () =>
+                    using (var scope = queue.AcquireLock())
                     {
-                        using var scope = queue.AcquireLock();
+                        if (!IsCurrent(subscription))
```
JakenVeina, line 112: Probably worth a comment here, to clarify that since we opened the lock for a bit, we need to re-check that the current source is still the latest, before we begin activating it. Or better yet, since we're doing it over and over, just one comment to that effect, to cover the whole operator. Maybe on the first spot where this happens?

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -24,101 +25,170 @@ public IObservable<IChangeSet<TObject, TKey>> Run() => Observable.Create<IChange
...
+                            scope.EnqueueNext(new ChangeSet<TObject, TKey>(
+                                current.KeyValues.Select(static pair => new Change<TObject, TKey>(ChangeReason.Remove, pair.Key, pair.Value))));
 
-                        // The current source may still be running, and the result ends only once both have.
-                        if (!isSourceRunning)
+                            current.Clear();
+                        }
+                    }
+
+                    // Reset delivery is another reentrant boundary. Outer completion alone does not
```
JakenVeina, line 126: So, I eventually figured out what the you were talking about, but I found this comment EXTREMELY confusing and unhelpful. "Reset" delivery really threw me off, because we're not resetting, we're clearing. And as I noted above, it's very unclear what the "outer" means.

Basically, I think the key point here is just that "we just did a delivery outside of a lock", so we need to double-check again that we're still the current source".

Also, really, this check is superfluous. The subscription already checks whether the source is still current on every incoming notification. All we're really doing here is an optimization, to skip the subscription if it's unnecessary. Which is fair, but it should be commented that way, then.

On a separate note, the fact that disposing a `.AcquireLock()` triggers delivery of queued notifications is really unclear from context. I think we should either split out a separate method for delivery like `.Flush()` or `.TryFlush()` or something, or rename `.AcquireLock()` to better clarify that it also performs delivery (say, `.AcquireDeliveryLock()` or `.TryDelivery()`). But that's not really part of this PR.

src/DynamicData/Cache/Internal/Switch.cs
```diff
@@ -24,101 +25,170 @@ public IObservable<IChangeSet<TObject, TKey>> Run() => Observable.Create<IChange
...
                         }
+                    });
 
-                        // Subscribed outside the lock. The source may deliver synchronously, and that
-                        // delivery takes the lock for itself.
-                        subscription.Disposable = source.SubscribeSafe(
-                            changes =>
-                            {
-                                using var scope = queue.AcquireLock();
+                return Disposable.Empty;
```
JakenVeina, line 85: Okay, I'm guessing this is part of the whole `.Using()` lifetime strategy? That the `.Using()` operator captures subscription disposal, and forwards that to `lifetime`?

Even if this is correct, I find it EXTREMELY confusing to have an operator return an empty subscription. If we truly need to leverage this pattern to get fully correct teardown behavior, we need to roll that into its own operator, where all these quirks are clearly documented.

However, I'm not convinced the `.Using()` is necessary. See my comment for the whole review.

src/DynamicData.Tests/Cache/SwitchFixture.SubscriptionLifetime.cs
```diff
@@ -0,0 +1,395 @@
...
+        results.RecordedChangeSets.SelectMany(changes => changes)
+            .Select(change => (change.Reason, change.Key, change.Current))
+            .Should().Equal(
+                (ChangeReason.Add, people[0].Name, people[0]),
+                (ChangeReason.Remove, people[0].Name, people[0]),
+                (ChangeReason.Add, people[1].Name, people[1]));
+    }
+
+    [Fact]
+    public void ReentrantSelectionDuringResetDoesNotSubscribeSupersededReplacement()
```
JakenVeina: I think this test, and all the "DoesNotSubscribeReplacement" tests are invalid. Functionally, who cares whether we subscribe to a source or not? What's really important is two-fold:

1. If a source is replaced, we stop forwarding its notifications. Never subscribing is certainly one way to accomplish that, but it's an implementation detail.
2. Everything that we DO subscribe to needs to get unsubscribed when we're done. I.E. no subscription leaks.

This ties into what I said above about the check that skips subscription to a stale source: it's really an optimization, not a functional requirement.

src/DynamicData.Tests/Cache/SwitchFixture.SubscriptionLifetime.cs
```diff
@@ -0,0 +1,395 @@
...
+namespace DynamicData.Tests.Cache;
+
+public partial class SwitchFixture
+{
+    private readonly ITestOutputHelper _output;
+
+    public SwitchFixture(ITestOutputHelper output) => _output = output;
+
+    [Fact]
+    public void ReplacementReleasesPreviousSubscriptionBeforeAcquiringResource()
```
JakenVeina, line 24: I don't really think this is a valid test. Nor are many of the others below. How the operator deals with subscriptions internally isn't what's functionally important. What's functionally important is two-fold:

1. Notification correctness. When a new source emits, we should stop delivering notifications from older sources. Or, more specifically, we should not violate the sequence of notifications. Notifications from older sources should never be delivered once notifications from a new source have started (including the initial clear, when present). In scenarios with concurrent notifications from different sources we can't guarantee which notifications we'll get to and which ones will get skipped, but we should always guarantee the sequencing of the ones we do publish. 
2. Subscription disposal. Avoiding an unnecessary subscription entirely is certainly a good way to prevent a leak, and a decent optimization (as I mention within the `Switch` operator itself) but it's really an implementation detail. The real functional requirement is that everything we DO subscribe to gets unsubscribed when we're done with it. That's what we should be testing for, in various scenarios.

JakenVeina (changes requested):
So, given that I had so much trouble reading the operator code, and it was basically a full re-write, I ended up re-writing it myself, to wrap my head around it.

Functionally, what I ended up with is pretty much the same, but I think it ended up quite a bit simpler, and more readable, while still passing all the tests on the PR branch.

Notably, I did not need to use `Observable.Using()` at all, nor did I have to abandon `SerialDisposable`.

Feel free to use it entirely, or partially, or not at all.
```cs
// Copyright (c) 2011-2025 Roland Pheasant. All rights reserved.
// Roland Pheasant licenses this file to you under the MIT license.
// See the LICENSE file in the project root for full license information.

using System.Reactive.Disposables;
using System.Reactive.Linq;

namespace DynamicData.Cache.Internal;

internal static class Switch2<TObject, TKey>
    where TObject : notnull
    where TKey : notnull
{
    public static IObservable<IChangeSet<TObject, TKey>> Create(
            IObservable<IObservable<IChangeSet<TObject, TKey>>> sources)
        => Observable.Create<IChangeSet<TObject, TKey>>(downstreamObserver =>
        {
            // We need to keep track of all the current downstream items, so that when a new source comes in,
            // we can issue remove changes for all of them.
            var downstreamItems = new Cache<TObject, TKey>();

            var hasSourcesCompleted = false;

            object currentSourceKey;
            var currentSourceSubscription = new SerialDisposable();
            var hasCurrentSourceCompleted = false;

            var hasFinalized = false;

            var downstreamNotificationQueue = new DeliveryQueue<IChangeSet<TObject, TKey>>(downstreamObserver);

            IDisposable sourcesSubscription;
            try
            {
                sourcesSubscription = sources
                    .SubscribeSafe(
                        onNext: upstreamSource =>
                        {
                            SingleAssignmentDisposable upstreamSourceSubscription;
                            object upstreamSourceKey;

                            // Begin tracking this source as the current one before actually subscribing to it. This ensures
                            // that if a new source comes in before subscription is complete (I.E. via reentrancy), that
                            // one still supersedes and replaces this one.
                            using (downstreamNotificationQueue.AcquireReadLock())
                            {
                                upstreamSourceSubscription = new SingleAssignmentDisposable();
                                upstreamSourceKey = upstreamSourceSubscription;
                                currentSourceKey = upstreamSourceKey;
                                hasCurrentSourceCompleted = false;
                            }

                            // Don't swap the subscription container inside a lock, as swapping it will trigger disposal of any
                            // prior subscription, and that can trigger user code that re-enters, potentially leading to a
                            // deadlock.
                            currentSourceSubscription.Disposable = upstreamSourceSubscription;

                            // Before subscribing to the new source, publish removals for any existing items from the prior one.
                            using (var downstreamNotificationScope = downstreamNotificationQueue.AcquireLock())
                            {
                                // Since we released the lock for a bit, we need to double-check that we're still the "current"
                                // source. We'll be doing this at the beginning of basically every in-lock operation.
                                // Also, make sure there's actual items to be removed.
                                if (!ReferenceEquals(currentSourceKey, upstreamSourceKey))
                                    return;

                                if (downstreamItems.Count is not 0)
                                {
                                    downstreamNotificationScope.EnqueueNext(new ChangeSet<TObject, TKey>(downstreamItems.KeyValues
                                        .Select(static pair => new Change<TObject, TKey>(
                                            reason: ChangeReason.Remove,
                                            key: pair.Key,
                                            current: pair.Value))));

                                    downstreamItems.Clear();
                                }
                            }

                            // Subscription is costly, so double-check once more that it's still necessary.
                            // This is really just an optimization, as the subscription already has all the correctness
                            // checks it needs.
                            using (downstreamNotificationQueue.AcquireReadLock())
                            {
                                if (!ReferenceEquals(currentSourceKey, upstreamSourceKey) || hasFinalized)
                                    return;
                            }

                            // Same as above, don't perform subscription inside a lock, as it can trigger user code that
                            // re-enters, potentially leading to a deadlock.
                            upstreamSourceSubscription.Disposable = upstreamSource.SubscribeSafe(
                                onNext: upstreamChanges =>
                                {
                                    using var downstreamNotificationScope = downstreamNotificationQueue.AcquireLock();

                                    if (!ReferenceEquals(currentSourceKey, upstreamSourceKey) || (upstreamChanges.Count is 0))
                                        return;

                                    downstreamItems.Clone(upstreamChanges);
                                    downstreamNotificationScope.EnqueueNext(upstreamChanges);
                                },
                                onError: error =>
                                {
                                    using (downstreamNotificationQueue.AcquireReadLock())
                                    {
                                        if (!ReferenceEquals(currentSourceKey, upstreamSourceKey))
                                            return;
                                    }

                                    OnError(error);
                                },
                                onCompleted: () =>
                                {
                                    using (downstreamNotificationQueue.AcquireReadLock())
                                    {
                                        if (!ReferenceEquals(currentSourceKey, upstreamSourceKey))
                                            return;

                                        hasCurrentSourceCompleted = true;

                                        if (!hasSourcesCompleted)
                                            return;

                                        hasFinalized = true;
                                    }

                                    using var downstreamNotificationScope = downstreamNotificationQueue.AcquireLock();

                                    downstreamNotificationScope.EnqueueCompleted();
                                });
                        },
                        onError: OnError,
                        onCompleted: () =>
                        {
                            using (downstreamNotificationQueue.AcquireReadLock())
                            {
                                hasSourcesCompleted = true;

                                if (!hasCurrentSourceCompleted)
                                    return;

                                hasFinalized = true;
                            }

                            using var downstreamNotificationScope = downstreamNotificationQueue.AcquireLock();

                            downstreamNotificationScope.EnqueueCompleted();
                        });
            }
            catch
            {
                downstreamNotificationQueue.Dispose();
                currentSourceSubscription.Dispose();

                throw;
            }

            return Disposable.Create(
                state: (
                    downstreamNotificationQueue,
                    currentSourceSubscription,
                    sourcesSubscription),
                dispose: state =>
                {
                    hasFinalized = true;
                    state.downstreamNotificationQueue.Dispose();
                    state.currentSourceSubscription.Dispose();
                    state.sourcesSubscription.Dispose();
                });

            void OnError(Exception error)
            {
                // Because setting a flag that only ever sets is atomic, and Observable.Create() automatically
                // terminates subscriptions upon an error notification, we can safely skip any state checking and just
                // forward the error.
                hasFinalized = true;

                using var downstreamNotificationScope = downstreamNotificationQueue.AcquireLock();

                downstreamNotificationScope.EnqueueError(error);
            }
        });
}
```

PR #1189, opened 2026-09-21
Restore named timer BatchIf overload compatibility

### LINKED ISSUE #1175: [Bug]: BatchIf no longer accepts an existing named-timer call without initialPauseState
### Describe the bug 🐞

Making initialPauseState required on the timer overload breaks the previously unambiguous source.BatchIf(pause, timer: ticks) call shape, including the form with an explicit scheduler. The new overload set has no forwarding replacement for that shape.

### Step to reproduce

Compile changes.BatchIf(pause, timer: timer), and changes.BatchIf(pause, timer: timer, scheduler: scheduler), against current main.

Regression tests/checks:

```
NamedTimerWithoutInitialState_StartsUnpaused; NamedTimerWithoutInitialState_FlushesOnTimer
```

### Reproduction repository

Regression tests and fixes:

### Expected behavior

Both previously supported forms should compile, begin unpaused, and use the timer to flush explicitly paused batches.

### Screenshots 🖼️

N/A.

### IDE

### Operating system

Windows.

### Version

### Device

N/A.

### DynamicData Version

### Additional information ℹ️

Four repository call expressions fail compilation with CS7036 for the missing initialPauseState argument. After adding the forwarding overload, all four behavioral cases pass.

dwcullop:
## Summary

Restores source compatibility for `BatchIf` calls that pass a named `timer` argument (with an optional `scheduler`) but omit `initialPauseState`. A compatibility overload is added so those call sites bind again; the batch starts unpaused, matching the previous behavior.

Also documents the overload's initial pause state in the cache operator instructions.

## Validation

Prior validation already passed: the solution built across all seven target frameworks, and the focused `BatchIf` test suite passed. New API surface is captured in the approved API verification files for .NET 8 and .NET 9.

Fixes #1175


dwcullop:
**P2: The new timer overload breaks an existing positional call.**

Location: [`ObservableCacheEx.BatchIf.cs`, lines 47-49](https://github.com/reactivemarbles/DynamicData/blob/ab8183d798be466917c295e2daaa53dcfc2d1bba/src/DynamicData/Cache/ObservableCacheEx.BatchIf.cs#L47-L49).

`source.BatchIf(pause, null, scheduler)` compiles against the base commit, selecting the overload with a nullable timeout. Against this PR it fails with `CS0121`, because the new `IObservable<Unit>?` parameter and the existing `TimeSpan?` parameter both accept the positional `null`.

Suggested fix: change the added overload shape so named `timer` calls remain available without introducing another candidate for the existing positional nullable-timeout call. This requires an API-shape decision rather than a null check in the implementation, since resolution fails before any method runs. Include compile-time compatibility coverage for both named-timer calls and the existing `BatchIf(pause, null, scheduler)` form. Requiring downstream callers to add a cast would leave this as a source-breaking change rather than a compatibility restoration.

src/DynamicData.Tests/Cache/BatchIfFixture.Overloads.cs
```diff
@@ -0,0 +1,73 @@
+﻿// Copyright (c) 2011-2026 Roland Pheasant. All rights reserved.
```
JakenVeina, line 1: Based on the rest of the codebase, we generally don't test operator overloads that are just a no-logic alias of another. I don't see why we should start here.

JakenVeina (changes requested):

PR #1191, opened 2026-09-21
Correct sample standard deviation calculations

### LINKED ISSUE #1181: [Bug]: StdDev applies the sample divisor outside the square root and truncates integer variance
### Describe the bug 🐞

Restoring deleted standard-deviation coverage reveals a numeric defect. All five numeric implementations divide sqrt(sum of squared deviations) by n-1 rather than taking sqrt(variance/(n-1)). Two-item examples hide this because the divisor is one. Integer implementations additionally truncate the mean-square term before computing variance.

### Step to reproduce

Add three equally spaced generated values in one cache or list edit. Their sample standard deviation must equal the spacing. Exercise int, long, float, double and decimal selectors. Also use integer data with a fractional mean and compare with the sample-variance formula.

Regression tests/checks:

```
StdDevFixture.Cache_ThreeEquallySpacedValues_ReportsTheirSpacing; StdDevFixture.List_ThreeEquallySpacedValues_ReportsTheirSpacing; StdDevFixture.IntegerValues_FractionalMean_PreservesFractionalVariance
```

### Reproduction repository

Regression tests and fixes:

### Expected behavior

Return sqrt(sum((value - mean)^2)/(n-1)), preserving fractional means/variance for integer inputs. Continue using the configured fallback for fewer than two items.

### Screenshots 🖼️

N/A.

### IDE

### Operating system

Windows.

### Version

### Device

N/A.

### DynamicData Version

### Additional information ℹ️

Twelve repository cases fail on unchanged main: expected 3, actual about 2.1213203435596424; fractional-mean expected 5.033222956847166, actual 3.570714214271425. Seventeen cases pass after correcting the expressions, including fallback controls.

dwcullop:
Fixes #1181

## Problem

Every `StdDev` overload computed the result with a misplaced divisor. The corrected sum of squares was divided by `Count` **inside** `Math.Sqrt`, and the `1 / (Count - 1)` sample factor was then applied **outside** the square root:

```csharp
Math.Sqrt(SumOfSquares - ((SumOfItems * SumOfItems) / Count)) * (1.0d / (Count - 1))
```

That is not the sample standard deviation, nor the population standard deviation. It is the square root of the total corrected sum of squares scaled by `1 / (n - 1)`, so results were wrong for every input of size 2 or more.

The `int` and `long` overloads had a second defect: `(SumOfItems * SumOfItems) / Count` was evaluated in integer arithmetic, truncating the mean correction before it reached the floating-point path.

## Fix

The divisor now lives inside the square root, and the integer overloads widen to `double` before dividing:

```csharp
Math.Sqrt((SumOfSquares - (((double)SumOfItems * SumOfItems) / Count)) / (Count - 1))
```

The `decimal` overload gets the same restructuring using its `Sqrt` helper.

## ⚠️ Behavioral change: returned values intentionally change

**This changes the numbers returned by `StdDev` for all callers.** Any consumer comparing against previously observed output, or with recorded baselines/snapshots, will see different values. The new values are the correct sample standard deviation (Bessel-corrected, `n - 1` denominator); the old values were not a standard deviation under any convention, so the change is a correction rather than a redefinition. The public signatures are unchanged.

Fallback behavior is unchanged: when fewer than two items remain, the configured `fallbackValue` is returned.

## Coverage

`StdDev` previously had **no test coverage at all**. This adds `StdDevFixture` with tests over all five numeric selector types (`int`, `long`, `float`, `double`, `decimal`), validating computed values against independently calculated expectations, plus incremental add/remove/update behavior and the fewer-than-two-items fallback path.

## Documentation

Adds `StdDev` operator sections to the cache and list instruction references, documenting the sample-deviation formula, the fallback rule, and per-change-reason handling.


dwcullop:
**P2: Integer standard deviation loses small variances when the moments are converted to double.**

Locations: [`StdDevEx.cs`, line 146](https://github.com/reactivemarbles/DynamicData/blob/ce90221f23d11048cbbe5f976f98d34625290186/src/DynamicData/Aggregation/StdDevEx.cs#L146), and the corresponding `long` calculation at line 157.

Casting `SumOfItems` to `double` before calculating and subtracting the squared-mean term avoids integer division, but it also converts the subtraction to floating-point arithmetic. Large, nearly equal moments then round to the same value, losing the variance.

Reproduced with the explicit integer overload, `source.Connect().StdDev(value => value, 0)`, and inputs `[174542947, 174542949]`: the PR returns `0`; the base returns the correct sample standard deviation, `1.4142135623730951`. Neither the sum nor the squared moments overflow `long` for this input.

Suggested fix: calculate variance using a numerically stable centered representation, such as a running mean and centered sum of squares with matching add/remove updates, rather than subtracting two large floating-point moments. This requires coordinated accumulator changes, not a replacement of the final cast alone. Add translation-invariance coverage for both `int` and `long` selectors: widely offset inputs with identical spacing should produce the same deviation, including after removals and replacements.

src/DynamicData/Aggregation/StdDevEx.cs
```diff
@@ -7,8 +7,12 @@
 namespace DynamicData.Aggregation;
 
 /// <summary>
-/// Extensions for calculating standard deviation.
+/// Extensions for calculating sample standard deviation.
 /// </summary>
+/// <remarks>
```
JakenVeina: We do not need to be responsible for educating consumers on what a standard deviation is.

src/DynamicData/Aggregation/StdDevEx.cs
```diff
@@ -139,7 +143,7 @@ public static IObservable<double> StdDev<TObject, TKey>(this IObservable<IChange
...
-    public static IObservable<double> StdDev<T>(this IObservable<IAggregateChangeSet<T>> source, Func<T, int> valueSelector, int fallbackValue = 0) => source.StdDevCalc(t => (long)valueSelector(t), fallbackValue, (current, item) => new StdDev<long>(current.Count + 1, current.SumOfItems + item, current.SumOfSquares + (item * item)), (current, item) => new StdDev<long>(current.Count - 1, current.SumOfItems - item, current.SumOfSquares - (item * item)), values => Math.Sqrt(values.SumOfSquares - ((values.SumOfItems * values.SumOfItems) / values.Count)) * (1.0d / (values.Count - 1)));
+    public static IObservable<double> StdDev<T>(this IObservable<IAggregateChangeSet<T>> source, Func<T, int> valueSelector, int fallbackValue = 0) => source.StdDevCalc(t => (long)valueSelector(t), fallbackValue, (current, item) => new StdDev<long>(current.Count + 1, current.SumOfItems + item, current.SumOfSquares + (item * item)), (current, item) => new StdDev<long>(current.Count - 1, current.SumOfItems - item, current.SumOfSquares - (item * item)), values => Math.Sqrt((values.SumOfSquares - (((double)values.SumOfItems * values.SumOfItems) / values.Count)) / (values.Count - 1)));
```
JakenVeina: Can we get some formatting on these, while we're making changes? As it is, this is unreadable garbage.

JakenVeina (changes requested):

PR #1193, opened 2026-09-21
Release property handlers when initialization throws

### LINKED ISSUE #1176: [Bug]: Property observation leaks event handlers when initial subscription throws
### Describe the bug 🐞

Shallow and deep property-observation subscription constructors install event handlers and then perform initial reads/delivery. If a getter failure reaches the default throwing error handler, or an initial OnNext callback throws, construction never returns the disposable to Rx. The installed handlers retain the failed subscription.

### Step to reproduce

Use an INotifyPropertyChanged model with counted event add/remove accessors. Subscribe with an initial observer that throws, or with an initial getter that throws and the default error handler. Inspect handler counts after Subscribe throws. Repeat with a multi-level property chain and partially attached chain.

Regression tests/checks:

```
Shallow_InitialObserverThrows_DetachesHandler; DeepChain_InitialObserverThrows_DetachesEveryHandler; Shallow_InitialGetterThrows_DefaultErrorHandler_DetachesHandler; DeepChain_InitialGetterThrows_DefaultErrorHandler_DetachesEveryHandler
```

### Reproduction repository

Regression tests and fixes:

### Expected behavior

Every installed handler must be detached when initialization fails, while preserving the original exception and valid initial/property-change notifications.

### Screenshots 🖼️

N/A.

### IDE

### Operating system

Windows.

### Version

### Device

N/A.

### DynamicData Version

### Additional information ℹ️

Repository tests retain one shallow handler or deep handler counts of [1,1,1] / [1,1] instead of zero. Controls with an explicit nonthrowing error recorder pass.

dwcullop:
## Problem

`WhenPropertyChanged` / `WhenValueChanged` attached `PropertyChanged` handlers and emitted the initial value from the subscription's constructor. If anything threw during that work (a property getter, or a subscriber callback receiving the initial value), the exception escaped before `Observable.Create` could hand back a disposable. Rx had nothing to dispose, so every handler installed on the way in stayed attached and kept the observed objects alive.

## Fix: ownership before activation

The subscription object is now constructed first, scoped by a `RefCountDisposable`, and only then started:

```csharp
var subscription = new SinglePropertySubscription(observer, source, memberName, accessor);
using var lifetime = new RefCountDisposable(subscription);
subscription.Start(notifyInitial);
return lifetime.GetDisposable();
```

The scope owns initialization. If `Start` throws, the `using` releases the subscription and detaches every handler attached so far, at every level of a deep chain. Only successful activation hands a dependent lease to Rx. The original exception still propagates unchanged, and subscriber callback failures are never converted into `OnError`.

Both the shallow single-property path and the deep-chain path are covered; `DeepChainSubscription` gains the same `Start()` seam so its chain walk runs under the same ownership scope.

## Tests

`WhenPropertyChangedBehaviorFixture.Initialization.cs` adds regression coverage for a throwing initial observer, a throwing getter under both the default and a handled error path, shallow and deep chains, explicit disposal, and synchronous completion via `Take(1)`. Each asserts handler counts drop to zero while confirming handlers attach before the value is read. Prior focused validation of these tests passed.

Fixes #1176


src/DynamicData/Binding/NotifyPropertyChangedEx.cs
```diff
@@ -245,6 +245,12 @@ public static class NotifyPropertyChangedEx
     /// For an object like Parent.Child.Sibling, sibling is an object so if Child is null, the value null and obtainable and is returned as null.</param>
     /// <returns>A observable which also notifies when the property value changes.</returns>
     /// <exception cref="ArgumentNullException">propertyAccessor.</exception>
+    /// <remarks>
```
JakenVeina, line 248: "Not leaking" isn't really a behavior that needs to be documented, it should be assumed true everywhere. So the note about detaching event handlers doesn't really make sense.

Noting that exceptions during initialization propagate upstream instead of down is valid, except... didn't we just discuss that this operator always propagates upstream, in that PR a couple months ago where you wanted to change that behavior? Am I mis-remembering?

src/DynamicData/Binding/ObservablePropertyFactory.cs
```diff
@@ -22,8 +22,16 @@ public ObservablePropertyFactory(Func<TObject, TProperty> valueAccessor, Observa
...
         // chain is leaf-first (output of SplitIntoSteps). Reverse once to root-to-leaf order.
         var rootToLeaf = chain.AsEnumerable().Reverse().ToArray();
-        _factory = (source, notifyInitial) => Observable.Create<PropertyValue<TObject, TProperty>>(
-            observer => new DeepChainSubscription(observer, source, rootToLeaf, valueAccessor, notifyInitial));
+        _factory = (source, notifyInitial) => Observable.Create<PropertyValue<TObject, TProperty>>(observer =>
+        {
+            var subscription = new DeepChainSubscription(observer, source, rootToLeaf, valueAccessor, notifyInitial);
+
+            // The scope owns initialization; only successful activation hands a dependent lease to Rx.
+            using var lifetime = new RefCountDisposable(subscription);
```
JakenVeina, line 30: What a lovely little utility class that I didn't know about.

src/DynamicData.Tests/Binding/WhenPropertyChangedBehaviorFixture.cs
```diff
@@ -6,20 +6,35 @@
...
 
 /// <summary>
 /// Single-threaded contract tests for <see cref="NotifyPropertyChangedEx.WhenPropertyChanged{TObject, TProperty}"/>:
-/// handler attachment ordering, no-dedup semantics, deep-chain re-walks on swaps.
+/// handler attachment ordering, subscription cleanup, no-dedup semantics, and deep-chain swaps.
 /// </summary>
-public sealed class WhenPropertyChangedBehaviorFixture
+public sealed partial class WhenPropertyChangedBehaviorFixture
 {
+    private readonly Randomizer _randomizer;
```
JakenVeina, line 27: Personally, I don't think putting state on test fixtures is a good idea. Not for, like, general-purpose test fixtures. Maybe for reusable test fixtures. Any test that might not want to use that shared state still has to pay for it.

What I generally do when I have setup logic I want to share is put it on a static method, or a separate nested `TestContext` class, so individual tests can opt into it.

In this case, the only thing being shared is really a one-liner initialization of a `Randomizer`. The seed really doesn't matter.

It's testing, though, and my other general rule for testing code is "Code quality/style REALLY doesn't matter that much. Strategy and readability are what's important."

src/DynamicData.Tests/Binding/WhenPropertyChangedBehaviorFixture.Initialization.cs
```diff
@@ -0,0 +1,305 @@
+﻿// Copyright (c) 2011-2025 Roland Pheasant. All rights reserved.
```
JakenVeina, line 1: I don't want to go too overboard with splitting tests across separate fixtures. My general rule of thumb is "one fixture per UUT", and then sub-splitting for "Unit" tests versus "Integration" tests, where necessary, because "Integration" tests usually shouldn't be parallelizable.

But same note as above, it's testing code. Just my 2 cents.

src/DynamicData.Tests/Binding/WhenPropertyChangedBehaviorFixture.Initialization.cs
```diff
@@ -0,0 +1,305 @@
...
+        models.Select(model => model.HandlerCount).Should().OnlyContain(count => count == 0,
+            because: "disposing the returned subscription must release every retained handler");
+        results.RecordedValues.Should().Equal(new[] { amount }, because: "initialization must publish the observed value");
+        results.Error.Should().BeNull(because: "explicit disposal is not an observation failure");
+        results.HasCompleted.Should().BeFalse(because: "unsubscribing does not publish a completion notification");
+    }
+
+    /// <summary>Verifies that synchronous completion during initial delivery releases every property handler.</summary>
+    /// <param name="deepChain">Whether the observed property is reached through intermediate objects.</param>
+    [Theory]
```
JakenVeina, line 228: Test name should reflect the actual behavior being tested (I.E. synchronous cleanup), rather than the mechanism by which you're exercising it (I.E. `.Take(1)`).

JakenVeina (changes requested):

src/DynamicData/Binding/ObservablePropertyFactory.cs
dwcullop: `RefCountDisposable` is so useful.

PR #1194, opened 2026-09-21
Evaluate value-changing conversions in property chains

### LINKED ISSUE #1177: [Bug]: Numeric conversions in observed property chains are discarded
### Describe the bug 🐞

ExpressionBuilder treats every Convert step as identity. For an observable double Amount, WhenValueChanged(x => ((decimal)x.Amount).Scale) consequently passes a boxed double to a property declared on decimal, and reports TargetException. The final compiled accessor cannot repair the preceding subscription/navigation walk.

### Step to reproduce

Create an INotifyPropertyChanged model with a double Amount. Observe the decimal Scale property through an explicit numeric conversion, then change Amount. Exercise initial notifications both enabled and disabled.

Regression tests/checks:

```
NumericConversionBeforePropertyAccess_InitialValue_IsObserved; NumericConversionBeforePropertyAccess_PropertyChanges_AreObserved
```

### Reproduction repository

Regression tests and fixes:

### Expected behavior

The conversion must be evaluated before subsequent property access, and the resulting values and later notifications must be delivered. Reference and interface casts must continue to work.

### Screenshots 🖼️

N/A.

### IDE

### Operating system

Windows.

### Version

### Device

N/A.

### DynamicData Version

### Additional information ℹ️

All three repository regression cases record TargetException. Leaf numeric conversion, reference-cast and interface-cast controls pass. This is distinct from the runtime INPC discovery limitation in #1156.

dwcullop:
## Problem

Property path expressions passed to `WhenPropertyChanged` / `WhenValueChanged` may contain conversions. `ExpressionBuilder.CreateInvoker` treated every `ExpressionType.Convert` node as a no-op pass-through, returning the original chain target unchanged.

That is correct for reference casts and boxing/unboxing, where the target object is the same instance and the runtime resolves the cast anyway. It is wrong for value-changing conversions: numeric conversions (`(decimal)price.Amount`) and user-defined conversion operators produce a *different* value, and the next step in the property path must read from that converted value rather than from the original operand.

The result was that an expression such as `price => ((decimal)price.Amount).Scale` attempted to read `decimal.Scale` off the source `double`, surfacing an error instead of the observed value.

## Fix

`CreateInvoker` now distinguishes the two cases:

- Built-in reference casts and boxing/unboxing (no conversion method, and at least one side is a reference type) keep the existing pass-through.
- Numeric and user-defined conversions compile a small invoker that unboxes the operand, applies the original conversion node, and re-boxes the converted result, so subsequent property access targets the converted value.

Conversions are also supported as the final step of a path, and reference/interface casts continue to follow object replacement in nested chains.

## Tests

Adds `WhenPropertyChangedBehaviorFixture.Conversions.cs` covering:

- numeric conversion before property access — initial value
- numeric conversion before property access — subsequent changes, with and without `notifyOnInitialValue`
- numeric conversion as the leaf of the path
- reference cast before property access
- interface cast before property access, including a replaced intermediate child

The shared fixture gains a seeded `Randomizer` (deterministic, seed written to test output) and a small `ObservablePrice` model to support these tests. Prior focused validation of these conversion tests passed.

## Docs

Public XML documentation on `WhenPropertyChanged` / `WhenValueChanged` and the cache/list instruction files now state that property paths evaluate numeric conversions before subsequent property access.

## Scope

Conversion handling only. No changes to property-subscription initialization lifetime behavior.

Fixes #1177


dwcullop:
**P2: Converted property paths collide in the property-factory cache.**

Locations: [`ExpressionBuilder.GetMembers`, lines 15-21](https://github.com/reactivemarbles/DynamicData/blob/48f628e9eaf183ec4e91304468946ed80f757beb/src/DynamicData/Binding/ExpressionBuilder.cs#L15-L21), and [`ToCacheKey`, lines 134-147](https://github.com/reactivemarbles/DynamicData/blob/48f628e9eaf183ec4e91304468946ed80f757beb/src/DynamicData/Binding/ExpressionBuilder.cs#L134-L147).

`GetMembers()` stops when it reaches a conversion. Consequently, `((decimal)x.First).Scale` and `((decimal)x.Second).Scale` generate the same key: the root type followed by `Scale`, without the property preceding the conversion. `ObservablePropertyFactoryCache.GetFactory()` then reuses the first expression's accessor and notification subscriptions for the second expression.

Reproduced with both expressions on the same model: the expected initial scales were `2` and `3`, but both subscriptions reported `2`. Changing `Second` to a value with scale `4` produced no update on its subscription. This is an existing cache-key weakness exposed by the newly supported conversion paths.

Suggested fix: build factory identity from the complete expression path, including each member's identity, conversion source/destination types, any user-defined conversion method, and the result type. `SplitIntoSteps()` already traverses the supported conversion nodes in this file, so it provides an existing traversal to reuse. Merely skipping conversion nodes in `GetMembers()` would still collide for different conversions of the same member. Add coverage for two different converted properties on one model and for different conversions of one property.

src/DynamicData/Binding/NotifyPropertyChangedEx.cs
```diff
@@ -245,6 +245,11 @@ public static class NotifyPropertyChangedEx
     /// For an object like Parent.Child.Sibling, sibling is an object so if Child is null, the value null and obtainable and is returned as null.</param>
     /// <returns>A observable which also notifies when the property value changes.</returns>
     /// <exception cref="ArgumentNullException">propertyAccessor.</exception>
+    /// <remarks>
```
JakenVeina, line 248: Pointless documentation. "Value-changing conversions are evaluated before accessing subsequent properties in the path" is just describing how C# expressions work. Consumers should expect that when they supply an expression to a function, it is evaluated correctly.

src/DynamicData/Binding/NotifyPropertyChangedEx.cs
```diff
@@ -267,6 +272,10 @@ public static IObservable<PropertyValue<TObject, TProperty>> WhenPropertyChanged
     /// For example when observing Parent.Child.Age, if Child is null the value is unobtainable as Age is a struct and cannot be set to Null.
     /// For an object like Parent.Child.Sibling, sibling is an object so if Child is null, the value null and obtainable and is returned as null.</param>
     /// <returns>An observable which emits the results.</returns>
+    /// <remarks>
```
JakenVeina, line 275: Same as above.

JakenVeina (changes requested):

